PR #8, opened 2018-03-14
As a User I should be able to clone Activity

underscope:


lbebic:
Working as intended except for a minor bug which will be covered on issue https://github.com/ExtensionEngine/tailor/issues/14

PR #28, opened 2018-03-21
Added global repo version

### LINKED ISSUE #24: Introduce global repository version
When there is a change in repo content, a global repo version should indicate it.

underscope:
FIxes #24


server/shared/publishing/helpers.js
```diff
@@ -171,6 +172,8 @@ function saveFile(parent, key, data) {
 }
 
 function saveSpine(spine) {
+  if (spine.version) delete spine.version;
+  spine.version = hasha(JSON.stringify(spine), { algorithm: 'sha1' });
```
vladimyr: `JSON.stringify` does not do object keys sorting thus two identical spines with different properties insertion order would result in different hashes 🛑 
Instead of relaying on fact that spines are always generated in same way i.e. having same properties insertion order we could use safer alternative: [hash-obj](https://npm.im/hash-obj) 🚑

PR #67, opened 2018-03-30
Add Audio teaching element

mbareta:


client/components/editor/toolbar/AudioToolbar.vue
```diff
@@ -0,0 +1,78 @@
+<template>
```
mbareta: This file is 1:1 copy of VideoToolbar.vue.

client/components/editor/teaching-elements/Audio.vue
```diff
@@ -0,0 +1,145 @@
...
+  },
+  data() {
+    return {
+      hasError: false
+    };
+  },
+  computed: {
+    audioElement() {
+      return this.$refs.audio;
+    },
```
vladimyr: For sake of having less keys to hit please use `player` or just `audio`. :bowtie:

client/components/editor/teaching-elements/Audio.vue
```diff
@@ -0,0 +1,145 @@
...
+    source() {
+      return this.element.data.url;
+    },
+    showPlaceholder() {
+      return !this.source;
+    }
+  },
+  methods: {
+    playAudio() {
+      this.audioElement.play();
```
vladimyr: This is actually returning `Promise` with request result. If error occurs you'll get it in form of rejection so please handle it: https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play

client/components/editor/teaching-elements/Audio.vue
```diff
@@ -0,0 +1,145 @@
...
+        return this.playAudio();
+      }
+      return this.pauseAudio();
+    },
+    source() {
+      this.checkError();
+    }
+  },
+  mounted() {
+    this.checkError();
```
vladimyr: Instead of manually checking for error, at particular points in time; use event listener 😉

client/components/editor/teaching-elements/Audio.vue
```diff
@@ -0,0 +1,145 @@
...
+      const { error } = this.audioElement;
+      this.hasError = !!error;
+    }
+  },
+  watch: {
+    isFocused() {
+      if (this.isFocused) {
+        return this.playAudio();
+      }
+      return this.pauseAudio();
```
vladimyr: _Play on focus_ is one possible approach but it is not really aligned with existing video component. Please add some UI layer to enable manual playback control.
Pick your poison: https://github.com/vuejs/awesome-vue#audio--video 😉

mbareta: Are the standard HTML audio controls enough (play/pause, seek, volume)? Because that is what gets rendered to the user. This just triggers play when you focus the element.

vladimyr: If they look same across different browsers 🤔

mbareta:
screenshot from 2018-03-30 14-01-06

vladimyr:
> Vue Aplayer sets audio preload to none by default so this will not waste any resources and we don't need to handle that ourselves.

What we have here is probably good enough (:+1: :100: ) cause I don't expect someone to put there live audio stream but for sake of explaining myself take a look on following example: https://output.jsbin.com/gemazitada/1

Even though it uses `preload="none"` trick as soon as you hit <kbd>Play</kbd> stream will get loaded and `<audio>` won't unload it after you hit <kbd>Pause</kbd>.

mbareta:
@vladimyr if we remove the src of the audio track, it throws an error "Failed to init decoder". So in this case we would need to unsubscribe or suppress the error message and handle it only on play event. What do you think?

PR #70, opened 2018-03-30
Prefer `const` if not reassigned

vladimyr:
Applying ESLint's [`prefer-const`](https://eslint.org/docs/4.0.0/rules/prefer-const) through whole codebase.

I understand drive behind using `let` for objects that get changed later on but that is added semantics and I'm not sure we should put that burden on developer's back. It doesn't provide that much value and might be confusing due to fact that [`const` isn't really about immutability](https://mathiasbynens.be/notes/es6-const) of a value but binding instead? 🤔

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,6 +8,8 @@ module.exports = {
   // required to lint *.vue files
   plugins: ['html'],
   rules: {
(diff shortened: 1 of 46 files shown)


vladimyr:
Taking step back and closing this until it gets discussed offline... 🔙

PR #89, opened 2018-04-23
Fix case sensitive emails

vladimyr:
Making sure all emails are stored lowercase. Processing submission data for public forms (login, forgot & reset password) so we end up with _lowercased_ database queries.

:construction: This probably needs migration script to lowercase emails of existing users. :warning:

### DIFF
--- a/server/router.js
+++ b/server/router.js
@@ -7,6 +7,7 @@ const teRouter = require('./teaching-element').router;
 const userRouter = require('./user').router;
 
 const router = express.Router();
+router.use(processBody);
 router.use('/', courseRouter);
 router.use('/', activityRouter);
 router.use('/', commentRouter);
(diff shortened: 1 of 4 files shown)


kjuej:
@vladimyr An error is displayed in the console ("The specified email address is already in use.") when inviting the same user twice with different upper/lower casing. Emails consisting of any upper case letters will display this error (if there already is such email address in the database), and emails consisting of any lower case letters will not show this error.

vladimyr:
@kjuej It should now work as expected. Please try following scenarios:
1. Adding admin using `npm run add:admin <email> <password>` where email contains mixed casing and verifying login functionality working correctly (case insensitive)
2. Re-inviting users to course with different casing
3. Requesting password reset link using mixed casing email

Additionally, please make sure that all emails are stored lowercase in the database.

kjuej:
@vladimyr Everything is working as expected, these three scenarios cause no issues. Emails are all stored as lowercase in the database no matter what.

PR #111, opened 2018-08-21
Text selection assessment tool

BKova:
Adds a assessment featuring text selection. Supports content editing using quill and multiple correct answers selection.


client/components/editor/teaching-elements/Assessment/TextSelection/Selector.vue
```diff
@@ -0,0 +1,107 @@
+<template>
+  <div
+    @mouseup="onMouseup">
+    <div
+      v-html="structure"
+      :class="{ editing: this.isEditing }"
+      class="selector ql-editor"/>
```
vladimyr: Hmm, there are few issues with this template:
1. `<div>`'s can't be self closed
2. What's the point of having additional wrapper? Why haven't you just attach this `mouseup` listener onto `.selector.ql-editor` directly?

client/components/editor/teaching-elements/Assessment/TextSelection/Selector.vue
```diff
@@ -0,0 +1,107 @@
...
+    const processNode = selections.processNode.bind(selections);
+    const mapper = curry(nodeMapper)(processNode);
+    return {
+      selections,
+      structure: generateStructure(this.content, mapper)
+    };
+  },
+  methods: {
+    onMouseup(event) {
+      if (!this.isEditing) return;
```
vladimyr: Move this into template to prevent this handler being called at all.

client/components/editor/teaching-elements/Assessment/TextSelection/Selector.vue
```diff
@@ -0,0 +1,107 @@
...
+    color: $selectionColor;
+  }
+
+  .separator {
+    white-space: pre;
+  }
+
+  .targeted {
+    background-color: #42b983;
+  }
```
vladimyr: Where do you use those `.separator` & `.targeted` classes? 🤔

client/components/editor/teaching-elements/Assessment/TextSelection/Selector.vue
```diff
@@ -0,0 +1,107 @@
...
+    save() {
+      this.$emit('save', this.selections.ranges);
+    }
+  }
+};
+</script>
+
+<style lang="scss" scoped>
+$selectionBackground: rgb(181, 197, 136);
+$selectionColor: rgb(13, 60, 85);
```
vladimyr: Please use hex format for defining colors :bowtie:

client/components/editor/teaching-elements/Assessment/TextSelection/Selector.vue
```diff
@@ -0,0 +1,107 @@
...
+    white-space: pre;
+  }
+
+  .targeted {
+    background-color: #42b983;
+  }
+}
+
+.ql-editor{
+  height: 0%;
```
vladimyr: Hmm, you are missing space before opening curly and don't put units next to zero values ☝️

client/components/editor/teaching-elements/Assessment/TextSelection/Selector.vue
```diff
@@ -0,0 +1,107 @@
...
+    isEditing: { type: Boolean, required: true },
+    content: { type: String, required: true }
+  },
+  data() {
+    const selections = new Selections(this.correct);
+    const processNode = selections.processNode.bind(selections);
+    const mapper = curry(nodeMapper)(processNode);
+    return {
+      selections,
+      structure: generateStructure(this.content, mapper)
```
vladimyr: Why this isn't implemented as computed property because at the end of day it is _computed_ from content 😉

vladimyr: Whenever you can avoid currying by using lambda please do it so:
```js
generateStructure(this.content, (node, i) => nodeMapper(processNode, node, i));
```

client/components/editor/teaching-elements/Assessment/TextSelection/helpers.js
```diff
@@ -0,0 +1,101 @@
+import Blast from 'blast-vanilla';
+import first from 'lodash/first';
+import generateRange from 'lodash/range';
+import isEmpty from 'lodash/isEmpty';
+import last from 'lodash/last';
+
+const blastClass = 'blast';
+const wordClass = 'word';
+const blastWordSelector = `span.${blastClass}.${wordClass}`;
```
vladimyr: This is used only once; so inline it :bowtie:

client/components/editor/teaching-elements/Assessment/TextSelection/helpers.js
```diff
@@ -0,0 +1,101 @@
...
+const idPrefix = 'text-';
+
+const isText = node => node.nodeType === Node.TEXT_NODE;
+
+export const isSelected = node => node.hasAttribute('range');
+
+export function extractIndex(node) {
+  const index = node.id.substr(idPrefix.length);
+  return parseInt(index, 10);
```
vladimyr: Prefix is actually irrelevant:
```js
const [_, index] = node.id.split('-');
```

client/components/editor/teaching-elements/Assessment/TextSelection/helpers.js
```diff
@@ -0,0 +1,101 @@
...
+export function generateStructure(content, mapper) {
+  if (!content) return '';
+  const container = document.createElement('div');
+  container.innerHTML = content;
+  // eslint-disable-next-line no-new
+  new Blast(container);
+  const iterator = document.createNodeIterator(container, NodeFilter.SHOW_TEXT);
+  const nodes = [];
+  while (iterator.nextNode()) { nodes.push(iterator.referenceNode); }
+  nodes.forEach(mapper);
```
vladimyr: Calling `map`er in `forEach` ❓

client/components/editor/teaching-elements/Assessment/TextSelection/helpers.js
```diff
@@ -0,0 +1,101 @@
...
+
+export function generateStructure(content, mapper) {
+  if (!content) return '';
+  const container = document.createElement('div');
+  container.innerHTML = content;
+  // eslint-disable-next-line no-new
+  new Blast(container);
+  const iterator = document.createNodeIterator(container, NodeFilter.SHOW_TEXT);
+  const nodes = [];
+  while (iterator.nextNode()) { nodes.push(iterator.referenceNode); }
```
vladimyr: Previous three lines are completely generic and can be extracted to `filterNodes(nodeFilter)` function :bowtie:

client/components/editor/teaching-elements/Assessment/TextSelection/index.vue
```diff
@@ -0,0 +1,158 @@
...
+        <span class="mdi mdi-stop-circle-outline"/>
+          Edit content
+      </button>
+      <button
+        v-else
+        @click="select"
+        class="btn btn-primary btn-material">
+        <span class="mdi mdi-marker"/>
+          Select solutions
+      </button>
```
vladimyr: Well, few issues:
1. This is toggle button which means it should be **one** button
2. `<span>` can't be selfclosed
3. Class `.icon` is missing here
4. Button labels are wrongly indented
5. Attribute `type="button"` is missing on `<button>` element

client/components/editor/teaching-elements/Assessment/TextSelection/index.vue
```diff
@@ -0,0 +1,158 @@
...
+          content: this.text,
+          width: this.assessment.width
+        }
+      };
+    },
+    isFocused() {
+      if (!get(this.focusedElement, 'type')) return false;
+      const isEmbedded = this.focusedElement.embedded;
+      const isSameId = this.focusedElement.id === this.id;
+      return isEmbedded && isSameId;
```
vladimyr: How about making it simpler:
```js
const { focusedElement = {} } = this;
return focusedElement.embedded && focusedElement.id === this.id;
```

client/components/editor/teaching-elements/Assessment/TextSelection/index.vue
```diff
@@ -0,0 +1,158 @@
+<template>
+  <div :class="{ disabled: !isEditing }">
+    <h5 class="title">Text and selection</h5>
+    <div
+      @click="focus"
+      class="container">
+      <te-html
+        v-if="!isSelecting && isEditing"
+        :element="textElement"
+        :is-focused="isFocused"
```
vladimyr: Keep consistent formatting:
```html
:isFocused="isFocused"
```

client/components/editor/teaching-elements/Assessment/TextSelection/index.vue
```diff
@@ -0,0 +1,158 @@
...
+      this.update();
+    },
+    saveSelection(selection) {
+      this.correct = selection;
+      this.update();
+    },
+    select() {
+      if (this.isFocused) {
+        this.focusElement();
+      }
```
vladimyr: You don't need block here :bowtie:

client/components/editor/teaching-elements/Assessment/TextSelection/helpers.js
```diff
@@ -0,0 +1,101 @@
...
+  if (isEmpty(selection)) return [];
+  return [extractIndex(first(selection)), extractIndex(last(selection)) + 1];
+}
+
+export function isBlast(node) {
+  if (!(node instanceof HTMLSpanElement)) return false;
+  return node.classList.contains(blastClass);
+}
+
+export function modifySelectionNodes(element, range, options) {
```
vladimyr: Decompose this to:
- `words(el: HTMLElement): Array.<HTMLSpanElement>`
- `attr(el: HTMLElement, config: Object)`
- `classname(el: HTMLElement, config: Object)`

where `config` is hash with attribute/class-names as keys and boolean values indicating add/remove operation.

client/components/editor/teaching-elements/Assessment/TextSelection/Selector.vue
```diff
@@ -0,0 +1,107 @@
...
+    };
+  },
+  methods: {
+    onMouseup(event) {
+      if (!this.isEditing) return;
+      const node = event.target;
+      const domSelection = document.getSelection();
+      const selection = getSelectedWords();
+      if (!isEmpty(selection) && !domSelection.isCollapsed) {
```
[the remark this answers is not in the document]
BKova: `getSelectionWords` actually returns selection range without regards to `.isCollapsed` state.
e.g. if user does a click on a word, then `.isCollapsed` will be `true`, but `getSelectionWords` will return that word. In that case, `.isCollapsed` servers to differentiate between click and highlighted selection.
As I previously mentioned `getSelectionWords` validates `rangeCount` and not `isCollapsed`:
https://github.com/ExtensionEngine/tailor/blob/66d9981d65077277200201a5feb544966841f949/client/components/editor/teaching-elements/Assessment/TextSelection/helpers.js#L36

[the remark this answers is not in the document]
BKova: Then `getSelectedWords` won't return `[start, end]`, but something like `{ range, isCollapsed }`?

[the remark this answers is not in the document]
vladimyr: Who said you can't put property on array 😜

client/components/editor/teaching-elements/Assessment/TextSelection/Selector.vue
BKova: This isn't a computed property because it should be computed only once. It has a dependency on `this.selections` and that would mean regenerating and re-rendering  whole component each time a selection is made or removed. It's much better to manipulate few specific DOM nodes and add/remove classes on them.

vladimyr: > It has a dependency on this.selections and that would mean regenerating and re-rendering whole component each time a selection is made or removed.

Through bound `processNode` though, I'm not 100% sure that would force Vue into tracking `selections` and recalculate it on change? 🤔

client/components/editor/teaching-elements/Assessment/TextSelection/Selector.vue
```diff
@@ -29,44 +26,42 @@ export default {
   },
   data() {
     const selections = new Selections(this.correct);
-    const processNode = selections.processNode.bind(selections);
-    const mapper = curry(nodeMapper)(processNode);
     return {
       selections,
-      structure: generateStructure(this.content, mapper)
+      structure: generateStructure(this.content, (node, index) => {
+        const processNode = selections.processNode.bind(selections);
```
vladimyr: You are aware that calling `.bind` every time creates new function aka you're gonna create `content.length` number of exactly the same functions 🙃

client/components/editor/teaching-elements/Assessment/TextSelection/Selector.vue
```diff
@@ -29,44 +26,42 @@ export default {
...
-      const options = {
-        addClass: 'selected',
-        setAttribute: { name: 'range', value: range }
-      };
-      modifySelectionNodes(this.$el, range, options);
+      const classConfig = { selected: true };
+      const attributeConfig = { range };
+      const text = getText(this.$el, range);
+      modifyClass(text, classConfig);
+      modifyAttribute(text, attributeConfig);
```
vladimyr: How about:
```js
const text = getText(this.$el, range);
modifyClass(text, { selected: true });
modifyAttribute(text, { range }); 
```
Btw. it is actually `toggleClass` & `toggleAttribute`...

client/components/editor/teaching-elements/Assessment/TextSelection/Selector.vue
```diff
@@ -29,44 +26,42 @@ export default {
...
-      const options = {
-        removeClass: 'selected',
-        removeAttribute: 'range'
-      };
-      modifySelectionNodes(this.$el, range, options);
+      const classConfig = { selected: false };
+      const attributeConfig = { range: null };
+      const text = getText(this.$el, range);
+      modifyClass(text, classConfig);
+      modifyAttribute(text, attributeConfig);
```
vladimyr: Same as above ⬆️

client/components/editor/teaching-elements/Assessment/TextSelection/helpers.js
```diff
@@ -43,34 +41,46 @@ export function getSelectedWords() {
...
   const endWord = range.endContainer.parentElement;
   if (wordIsTarget(selection, startWord)) selection = [startWord, ...selection];
   if (wordIsTarget(selection, endWord)) selection = [...selection, endWord];
-  if (isEmpty(selection)) return [];
-  return [extractIndex(first(selection)), extractIndex(last(selection)) + 1];
+  if (isEmpty(selection)) return null;
+  return Object.assign(
+    [extractIndex(first(selection)), extractIndex(last(selection)) + 1],
+    { isCollapsed: domSelection.isCollapsed }
+  );
```
vladimyr: TBH this looks messy 🙃

client/components/editor/teaching-elements/Assessment/TextSelection/helpers.js
```diff
@@ -43,34 +41,46 @@ export function getSelectedWords() {
...
+  return indexes.map(index => element.querySelector(query(index)));
+}
+
+export function modifyAttribute(elements, config) {
+  const insert = pickBy(config);
+  const remove = pickBy(config, isNull);
+  elements.forEach(element => {
+    for (let key in insert) { element.setAttribute(key, insert[key]); }
+    for (let key in remove) { element.removeAttribute(key); }
+  });
```
vladimyr: Isn't this simpler:
```js
elements.forEach(el => {
  forEach(config, (name, value) => {
    const action = value !== null ? 'setAttribute' : 'removeAttribute';
    el[action](name, value);
  });
});
```

client/components/editor/teaching-elements/Assessment/TextSelection/helpers.js
```diff
@@ -43,34 +41,46 @@ export function getSelectedWords() {
...
+}
+
+export function modifyClass(elements, config) {
+  const insert = pickBy(config);
+  const remove = pickBy(config, conf => !conf);
+  elements.forEach(element => {
+    for (let key in insert) { element.classList.add(key); }
+    for (let key in remove) { element.classList.remove(key); }
```
vladimyr: Using same logic as above you can simplify this one too ⬆️

client/components/editor/teaching-elements/Assessment/TextSelection/index.vue
```diff
@@ -19,18 +19,13 @@
...
-        @click="select"
-        class="btn btn-primary btn-material">
-        <span class="mdi mdi-marker"/>
-          Select solutions
+        @click="isSelecting ? edit() : select()"
+        class="btn btn-primary btn-material"
+        type="button">
+        <span
+          :class="isSelecting ? 'mdi-stop-circle-outline' : 'mdi-marker'"
+          class="mdi"/>
```
vladimyr: Those are **only** html elements that can be self closed:
https://html.spec.whatwg.org/multipage/syntax.html#void-elements
No `<span>` here ❕

client/utils/assessment.js
```diff
@@ -109,6 +110,10 @@ export const schemas = {
       response: yup.string().trim().min(1).max(200).required()
     })
   }),
+  TS: yup.object().shape({
```
vladimyr: Is there any specific reason why you don't build upon `baseSchema` here? 🤔

client/utils/assessment.js
```diff
@@ -109,6 +110,10 @@ export const schemas = {
...
+    text: yup.string().trim().min(1).required(),
+    correct: yup.array().of(yup.mixed().required()).min(1)
```
vladimyr: Isn't this array of ranges/pairs? [`yup#mixed`](https://github.com/jquense/yup#mixed) is effectively _any_ type?

client/components/editor/teaching-elements/Assessment/TextSelection/index.vue
```diff
@@ -0,0 +1,148 @@
...
+  text-align: left;
+}
+
+.container {
+  margin-bottom: 20px;
+  width: 100%;
+}
+
+.ql-container.ql-snow {
+  border-style: none;
```
vladimyr: Prefer `border: none;` because it communicates intent better/stronger :bowtie:

client/components/editor/teaching-elements/Assessment/TextSelection/index.vue
```diff
@@ -0,0 +1,148 @@
+<template>
+  <div :class="{ disabled: !isEditing }">
+    <h5 class="title">Text and selection</h5>
+    <div
+      @click="focus"
+      class="container">
```
vladimyr: This fits under 80 columns? :bowtie:

client/components/editor/teaching-elements/Assessment/TextSelection/index.vue
```diff
@@ -0,0 +1,148 @@
...
+  methods: {
+    ...mapMutations(['focusElement'], 'editor'),
+    ...mapActions({ updateElement: 'update' }, 'tes'),
+    saveContent({ content }) {
+      this.text = content;
+      this.update();
+    },
+    saveSelection(selection) {
+      this.correct = selection;
+      this.update();
```
vladimyr: `saveSelection` & `saveContent` are almost identical; unify them? :bowtie:

client/components/editor/teaching-elements/Assessment/TextSelection/index.vue
```diff
@@ -0,0 +1,148 @@
...
+    focus(e) {
+      this.focusElement(this.textElement);
+      e.component = { name: 'selection-assesment', data: this.textElement };
+    },
+    update() {
+      const data = {
+        text: this.text,
+        correct: this.correct
+      };
+      this.$emit('update', data);
```
vladimyr:
```js
const { text, correct } = this;
this.$emit('update', { text, correct });

// or if you prefer using lodash

const data = pick(this, ['text', 'correct']);
this.$emit('update', data);
```

client/components/editor/teaching-elements/Assessment/TextSelection/index.vue
```diff
@@ -0,0 +1,148 @@
...
+    }
+  },
+  watch: {
+    errors() {
+      const notSelected = this.errors.includes('correct');
```
vladimyr: `notSelected` what? `hasCorrectAnswers`? 😉

client/components/editor/teaching-elements/Assessment/TextSelection/Selector.vue
```diff
@@ -0,0 +1,87 @@
...
+    content: { type: String, required: true }
+  },
+  data() {
+    const selections = new Selections(this.correct);
+    const processNode = selections.processNode.bind(selections);
+    return {
+      selections,
+      structure: generateStructure(this.content, (node, index) => {
+        return nodeMapper(processNode, node, index);
+      })
```
vladimyr: Move this above to separate variable:
```js
const structure = generateStructure(...);
```
so you can do:
```js
return { selections, structure };
```
Also this is in fact an html string, `structure` implies more complex type?

client/components/editor/teaching-elements/Assessment/TextSelection/Selector.vue
```diff
@@ -0,0 +1,87 @@
...
+  },
+  methods: {
+    onMouseup(event) {
+      const node = event.target;
+      const selection = getSelectionRange();
+      if (selection && !selection.isCollapsed) {
+        this.addSelection(selection);
+      } else if (Selections.isSelected(node) && isBlast(node)) {
+        const range = Selections.extractRange(node);
+        this.removeSelection(range);
```
vladimyr: Huh, all is great here from control flow standpoint but in the same time:
- `getSelectionRange` returns `selection` (which is in fact range)
- `this.addSelection` accepts `selection`
- `Selections.extractRange` returns `range` (ok)
- `this.removeSelection` accepts `range` (should be `removeRange` then?)

Naming is inconsistent and instead of lowering cognitive load it actually increases it 😢

client/components/editor/teaching-elements/Assessment/TextSelection/Selector.vue
```diff
@@ -0,0 +1,87 @@
...
+};
+</script>
+
+<style lang="scss" scoped>
+$selectionBackground: rgb(181, 197, 136);
+$selectionColor: rgb(13, 60, 85);
+
+.selector {
+  &.editing ::selection {
+      background: lighten($selectionBackground, 20%);
```
vladimyr: Indent is off here ☝️

client/components/editor/teaching-elements/Assessment/TextSelection/helpers.js
```diff
@@ -0,0 +1,110 @@
...
+import forEach from 'lodash/forEach';
+
+const blastClass = 'blast';
+const wordClass = 'word';
+const idPrefix = 'text-';
+
+const isText = node => node.nodeType === Node.TEXT_NODE;
+
+export function extractIndex(node) {
+  const [, index] = node.id.split('-');
```
vladimyr:
```js
const [_, index] = node.id.split('-');
```
triggered lint error or there is another reason you decided omit placeholder param listing? 🤔

client/components/editor/teaching-elements/Assessment/TextSelection/Selections.js
```diff
@@ -0,0 +1,62 @@
...
+    this.remove(range);
+    this.ranges.push(range);
+    return range;
+  }
+
+  remove(range) {
+    const check = curryRight(isOverlaped)(range);
+    this.ranges = this.ranges.filter(range => (
+      !check(first(range)) &&
+      !check(last(range) - 1)
```
vladimyr: Hmm, correct me if I'm wrong but this can be simplified by:
1. applying DeMorgan's law:
```js
range => !(check(first(range)) || check(last(range) - 1))
```
2. after which you can get rid of negation by _flipping_ operation itself
```js
reject(this.ranges, it => check(first(it)) || check(last(it) - 1);
```

client/components/editor/teaching-elements/Assessment/TextSelection/Selections.js
```diff
@@ -0,0 +1,62 @@
...
+
+  static extractRange(node) {
+    const range = node.getAttribute('range');
+    return range.split(',').map(Number);
+  }
+
+  static isSelected(node) {
+    return node.hasAttribute('range');
+  }
+}
```
vladimyr: You are actually violating Single Responsibility Principle here 🙃 
1. `Selections` class works with virtual data structure, while
2. `processNode`, `extractRange` and `isSelected` touch specific (word) DOM nodes (plus imported `extractRange` helper)

Divide that by responsibilities :bowtie:

client/components/editor/teaching-elements/Assessment/TextSelection/helpers.js
BKova: `[_, index]` syntax is triggering lint errors because `_` is unused.

vladimyr: Bad boy ESLint, bad boy 😁

client/components/editor/teaching-elements/Assessment/TextSelection/Selections.js
```diff
@@ -0,0 +1,43 @@
...
+    ));
+  }
+
+  getByIndex(index) {
+    const check = curry(isOverlaped)(index);
+    return this.ranges.find(check);
+  }
+}
+
+function isOverlaped(index, selection) {
```
vladimyr: You mean "overla**pp**ed"? 😉

client/components/editor/teaching-elements/Assessment/TextSelection/Selections.js
```diff
@@ -0,0 +1,43 @@
...
+import curry from 'lodash/curry';
+import curryRight from 'lodash/curryRight';
+import first from 'lodash/first';
+import inRange from 'lodash/inRange';
+import last from 'lodash/last';
+import reject from 'lodash/reject';
+
+export default class Selections {
+  constructor(selection = []) {
+    this.ranges = cloneDeep(selection);
```
vladimyr: Hmm, if selection is array of ranges and every range is actually a pair of start and end index then it is not required to use deep clone here? 🤔 
It is sufficient to simply copy by slicing? 🤔

client/components/editor/teaching-elements/Assessment/TextSelection/Selections.js
```diff
@@ -0,0 +1,43 @@
...
+    this.ranges.push(range);
+    return range;
+  }
+
+  remove(range) {
+    const check = curryRight(isOverlaped)(range);
+    this.ranges = reject(this.ranges, range => (
+      check(first(range)) || check(last(range) - 1)
+    ));
+  }
```
vladimyr: We need to establish some sort of terminology because it seems that `merge(selection)` and `remove(range)` use exactly same input types but list them under different names 🙃

client/components/editor/teaching-elements/Assessment/TextSelection/Selector.vue
```diff
@@ -0,0 +1,88 @@
...
+    );
+    return { selections, html };
+  },
+  methods: {
+    onMouseup(event) {
+      const node = event.target;
+      let range = getSelectionRange();
+      if (range && !range.isCollapsed) {
+        this.addRange(range);
+      } else if (isSelected(node) && isBlast(node)) {
```
vladimyr: Simplify this to independent `if`-s using early return :bowtie:

client/components/editor/teaching-elements/Assessment/TextSelection/helpers.js
```diff
@@ -0,0 +1,128 @@
+import Blast from 'blast-vanilla';
+import first from 'lodash/first';
+import generateRange from 'lodash/range';
```
vladimyr: Avoid this because later on it is almost impossible to guess from code that you are actually using lodash's utility.

client/components/editor/teaching-elements/Assessment/TextSelection/Range.js
```diff
@@ -0,0 +1,52 @@
+import { extractIndex } from './helpers';
+import filter from 'lodash/filter';
+import first from 'lodash/first';
+import inRange from 'lodash/inRange';
+import last from 'lodash/last';
+import sortBy from 'lodash/sortBy';
+
+export default class Range {
+  constructor(start, end) {
+    [start, end] = sortBy([start, end]);
```
vladimyr: How is this different than:
```js
[start, end] = [start, end].sort();
```

client/components/editor/teaching-elements/Assessment/TextSelection/Range.js
```diff
@@ -0,0 +1,52 @@
...
+    return this[1];
+  }
+
+  set end(val) {
+    this[1] = val;
+  }
+
+  static getFromDomSelection(selection = document.getSelection()) {
+    if (!selection.rangeCount) return null;
+    const selRange = selection.getRangeAt(0);
```
vladimyr: How about `domRange` or simply `range`? :bowtie: 
Avoid names that contain shortened segments.

client/components/editor/teaching-elements/Assessment/TextSelection/Range.js
```diff
@@ -0,0 +1,52 @@
...
+    const { parentElement } = selRange.commonAncestorContainer.parentElement;
```
vladimyr: This seems very fragile and also there is no in-code description of target's element kind. I cannot guess its tagName nor className without explicitly knowing html structure 🙃

client/components/editor/teaching-elements/Assessment/TextSelection/Range.js
```diff
@@ -0,0 +1,52 @@
...
+    let words = parentElement.querySelectorAll('span.blast.word');
+    words = filter(words, it => selRange.intersectsNode(it));
+    if (!words.length) return null;
+    return new Range(extractIndex(first(words)), extractIndex(last(words)) + 1);
+  }
+
+  static extract(node) {
+    const range = node.getAttribute('range');
+    return new Range(...range.split(',').map(Number).sort().reverse());
```
vladimyr: `.sort().reverse()` what? why? 🤔

client/components/editor/teaching-elements/Assessment/TextSelection/Selections.js
```diff
@@ -0,0 +1,32 @@
...
+import Range from './Range';
+import reject from 'lodash/reject';
+
+export default class Selections {
+  constructor(selections = []) {
+    this.ranges = selections.map(it => new Range(...it));
+  }
+
+  merge(range) {
+    const overlapses = this.ranges.filter(it => it.intersectsWith(range));
```
vladimyr: I'll just say: https://www.merriam-webster.com/dictionary/overlapses 😉

client/components/editor/teaching-elements/Assessment/TextSelection/Selections.js
```diff
@@ -0,0 +1,32 @@
...
+    range.start = min([...overlapses.map(it => it.start), range.start]);
+    range.end = max([...overlapses.map(it => it.end), range.end]);
```
vladimyr: I'm not sure if you aware of it but `Math.max` & `Math.min` are actually capable of accepting `...args` 😉

client/components/editor/teaching-elements/Assessment/TextSelection/Selector.vue
```diff
@@ -0,0 +1,88 @@
...
+      const node = event.target;
+      let range = Range.getFromDomSelection(selection);
+      if (range && !selection.isCollapsed) {
+        this.addRange(range);
+      } else if (isSelected(node) && isBlast(node)) {
+        range = Range.extract(node);
+        this.removeRange(range);
+      }
+    },
+    addRange(selection) {
```
vladimyr: `addRange` accepts `selection` 🙅‍♂️

client/components/editor/teaching-elements/Assessment/TextSelection/helpers.js
```diff
@@ -0,0 +1,81 @@
...
+    element.classList.toggle(name, value);
+  });
+}
+
+export function processNode(node, selections) {
+  const index = extractIndex(node);
+  const selection = selections.getByIndex(index);
+  if (!selection) return node;
+  node.classList.add('selected');
+  node.setAttribute('range', [...selection]);
```
vladimyr: If this is why you opted to implementing iterator on `Range` you could simply done:
```js
node.setAttribute('range', [selection.start, selection.end]);
```
And btw. this is not **selection** but **range** 😉

client/components/editor/teaching-elements/Assessment/TextSelection/helpers.js
```diff
@@ -0,0 +1,81 @@
...
+  return node;
+}
+
+export function isSelected(node) {
+  return node.hasAttribute('range');
+}
```
vladimyr:
```js
export const isSelected = node => node.hasAttribute('range');
```

client/components/editor/teaching-elements/Assessment/TextSelection/index.vue
```diff
@@ -0,0 +1,141 @@
...
+      e.component = { name: 'selection-assesment', data: this.textElement };
+    },
+    update() {
+      const { content, correct } = this;
+      this.$emit('update', { content, correct });
+    }
+  },
+  watch: {
+    errors() {
+      const correctEmpty = this.errors.includes('correct');
```
vladimyr: What does the 🦊 say 🕺

client/components/editor/teaching-elements/Assessment/TextSelection/Range.js
BKova: Quote from MDN for Array.sort: `The default sort order is according to string Unicode code points.` I need sort order according to values. e.g. `[19,9].sort() // => [19,9]`

client/components/editor/teaching-elements/Assessment/TextSelection/index.vue
BKova: Do you have any naming recommendations? 😞

client/components/editor/teaching-elements/Assessment/TextSelection/helpers.js
```diff
@@ -23,7 +25,7 @@ export function isBlast(node) {
 
 export function getText(element, range) {
   const query = index => `span#${idPrefix}${index}.${blastClass}`;
-  const indexes = generateRange(...range);
+  const indexes = generateRange(range.start, range.end);
   return indexes.map(index => element.querySelector(query(index)));
```
vladimyr: Actually you can get all elements using single DOM query:
```js
export function getText(container, range) {
  const length = range.end - range.start;
  const selectors = Array.from({ length }, (_, offset) => {
    const index = range.start + offset;
    return `span.${blastClass}[id="${index}"]`;
  });
  return Array.from(container.querySelectorAll(selectors.join());
}
```

client/components/editor/teaching-elements/Assessment/TextSelection/Range.js
```diff
@@ -0,0 +1,60 @@
...
+import last from 'lodash/last';
+
+const isText = node => node.nodeType === Node.TEXT_NODE;
+const sort = (...values) => values.sort((a, b) => a - b);
+
+export default class Range {
+  constructor(start, end) {
+    [start, end] = sort(start, end);
+    this[0] = start;
+    this[1] = end;
```
vladimyr: Well, because you don't use it as array convert it to `this._start` & `this._end` :bowtie:

PR #115, opened 2018-09-20
Add text highlight assessment

tkarabatic:
Adds a text highlight assessment. The component allows inputting a text, formatting it and highlighting parts of it that represent correct answers to the question.


underscope:

PR #117, opened 2018-10-02
Add embed element support on questions

ibebic:


kjuej:
@ibebic I can't enter a half width embed in the knowledge check(only full width). Also, I can no longer enter half width for both text and Images (I'm able to do this on the develop branch).

PR #128, opened 2018-11-07
Limited activity siblings fetching to only search from the same level in the schema

abasic:


server/activity/activity.model.js
```diff
@@ -192,12 +192,18 @@ class Activity extends Model {
...
-      return this.siblings().then(siblings => {
+      const filter = this.getReorderFilter();
+      return this.siblings(filter).then(siblings => {
         this.position = calculatePosition(this.id, index, siblings);
         return this.save({ transaction: t });
       });
     });
   }
+
+  getReorderFilter() {
```
underscope: Not sure if we need this as a separate method (since it won't be reused and is relatively simple).

config/shared/activities.js
```diff
@@ -61,6 +62,15 @@ function getLevel(type) {
   return schemaId && find(getOutlineLevels(schemaId), { type });
 }
 
+function getFromSameLevel(type) {
```
underscope: Consider renaming this to getSiblingLevels in order to keep the consistent naming

PR #131, opened 2018-11-28
Add content containers min and max count

kronicker:
- `min` and `max` numerical config fields are added to content containers config
- `min` number of containers are auto-created upon the opening of activity editor
-  If `min` is reached further deletion is disabled
-  If `max` is reached further addition is disabled

### DIFF
--- a/.activities-rc.json.example
+++ b/.activities-rc.json.example
@@ -32,7 +32,9 @@
     "contentContainers": [{
       "type": "PAGE",
       "label": "Page",
-      "multiple": false
+      "multiple": false,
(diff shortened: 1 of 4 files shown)


kronicker:
@franosincic @vladimyr @underscope @abasic This was intended as a feature introducing:
1. Automatic container creation which is already taken care of, and
2. More control over count limits of containers then current `multiple` flag.
If we don't want/need the latter, should we close this?

kronicker:
After an offline discussion and since part of it is already implemented, we decided to close until the need for this appear.

PR #135, opened 2018-11-28
Pin vee-validate package to 2.0.4

vkaracic:
New version of vee-validate (2.1.3) breaks the password reset form validation.

screenshot from 2018-11-28 14-38-13


kronicker:
This was fixed via #132.

PR #136, opened 2018-11-28
Change user invitation to run synchronously

vkaracic:
Asynchronously sending the invitation email does not work for
the invite:admin command since it exists the process before
the email is sent.

The `invite:admin` was killing the process before an email was sent to the user.
https://github.com/ExtensionEngine/tailor/blob/a4328bcaa4c81e4ff396a8506d96ce0b806aac01/server/script/inviteAdmin.js#L14-L18

This change will include the `mail.invite` step into the Promise chain, making sure (?) that it resolves before killing the process.

### DIFF
--- a/server/user/user.model.js
+++ b/server/user/user.model.js
@@ -96,12 +96,16 @@ class User extends Model {
     };
   }
 
-  static invite(user) {
-    return this.create(user).then(user => {
-      user.token = user.createToken({ expiresIn: '5 days' });
-      mail.invite(user);
-      return user.save();
-    });
+  static invite(userAttrs) {
+    let user;
+    return this
+      .create(userAttrs)
+      .then(u => {
+        user = u;
+        user.token = user.createToken({ expiresIn: '5 days' });
+      })
+      .then(() => mail.invite(user))
+      .then(() => user.save());
   }
 
   isAdmin() {


server/user/user.model.js
```diff
@@ -96,12 +96,16 @@ class User extends Model {
...
+  static invite(userAttrs) {
+    let user;
+    return this
+      .create(userAttrs)
+      .then(u => {
+        user = u;
+        user.token = user.createToken({ expiresIn: '5 days' });
+      })
+      .then(() => mail.invite(user))
+      .then(() => user.save());
```
vladimyr: This is how you can achieve same effect with minimal code changes:
```js
static invite(user) {
  return this.create(user).then(user => {
    user.token = user.createToken({ expiresIn: '5 days' });
    return mail.invite(user).then(() => user.save());
  });
}
```

vladimyr:
> This change will include the `mail.invite` step into the Promise chain, making sure (?) that it resolves before killing the process.

However this is not the only usage of `User.invite`. In scenario where you get http request from client side (e.g. from create user form) you don't want to delay response by waiting for email to get sent. What we need here is ability to hook into execution without altering its behavior which is implemented in #133 

That being said, thank you for your effort 👍 🏆 but in this particular case I'm closing it and accepting other one. Remember there is still large amount of work waiting to be done 😉

vkaracic:
Thanks for the review! Next time I'll check opened PR's first before opening duplicates :man_facepalming:

Issue #140, opened 2018-12-07
Reordering upper level elements is no longer possible


kjuej:
## I'm submitting a...

- [x] Bug report  
- [ ] Regression (a behavior that used to work and stopped working in a new release)
- [ ] Feature request
- [ ] Documentation issue or request

## Context
Can no longer drag and drop the upper level elements to change their position. Able to do that for lower level elements, but not for the most upper level one.
![screen shot 2018-12-07 at 15 44 34](https://user-images.githubusercontent.com/34127147/49654043-08a55b80-fa37-11e8-951d-49fdd841fac7.png)

PR #141, opened 2018-12-07
Re-enable top level reordering on outline

ibebic:


ibebic:
This resolves #140.

kjuej:
@ibebic Seems to be working correctly now.

client/utils/reorder.js
```diff
@@ -0,0 +1,13 @@
+import { mapActions } from 'vuex-module';
```
underscope: Please move it within outline folder and rename it to reorderMixin (unless you had more ideas where you could include it).

client/components/course/Outline/Activity.vue
```diff
@@ -86,7 +88,6 @@ export default {
   },
   methods: {
     ...mapMutations(['focusActivity', 'toggleActivity'], 'course'),
-    ...mapActions({ updatePosition: 'reorder' }, 'activities'),
```
underscope: Have you accidentally kept reorder function? (Bellow)

ibebic: My bad 🙈

underscope:
Thanks!

PR #148, opened 2018-12-14
Throw error when there are pending migrations

mbareta:


server/index.js
```diff
@@ -7,6 +7,7 @@ const capitalize = require('to-case').capital;
 const pkg = require('../package.json');
 const { promisify } = require('util');
 const sequelize = require('sequelize');
+const umzug = require('./shared/database/umzug');
```
vladimyr: ⚠️ Loading this loads `./shared/database/index.js` too which in fact loads all our models before `sequelize.Promise` gets configured.

server/index.js
```diff
@@ -20,6 +21,13 @@ const logger = require('./shared/logger');
 const runApp = promisify(app.listen.bind(app));
 
 database.initialize()
+  .then(() => umzug.pending())
+  .then(migrations => {
+    if (migrations.length) {
+      return bluebird.reject(new Error('There are pending migrations.'));
+    }
+    return bluebird.resolve();
```
vladimyr: You are inside promise chain anyway, anything you return (_or do not return_) will get wrapped inside promise 😉

server/shared/database/umzug.js
```diff
@@ -0,0 +1,19 @@
...
+  // The logging function.
+  // A function that gets executed everytime migrations start and have ended.
+  logging: false,
+  migrations: {
+    params: [db.sequelize.getQueryInterface(), db.sequelize.constructor, db],
+    path: './server/shared/database/migrations'
+  }
+});
+
+module.exports = umzug;
```
vladimyr: Personally I think this is good candidate for creating factory that'll receive `db` module as param 🤔

mbareta: 👌

mbareta:
@vladimyr please another pass.

server/shared/database/umzug.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+const Umzug = require('umzug');
+
+const umzug = db => new Umzug({
```
vladimyr:
```suggestion
module.exports = db => new Umzug({
```

underscope:
@kjuej Please test it

kjuej:
@mbareta @underscope Seems to be working as expected.

PR #152, opened 2018-12-19
Tweak activity deletion confirmation modal

### LINKED ISSUE #149: General activity deletion should be consistent to deletion inside the tree view
## I'm submitting a...

- [ ] Bug report  
- [x] Feature request

## Context
When deleting an activity inside the tree view, the activity ID is displayed.
screen shot 2018-12-14 at 14 10 33

However, the deletion of the elements in the outline does not display the ID.
screen shot 2018-12-14 at 14 10 52
Should the activity ID be added to the outline view or maybe have the activity name displayed when deleting the activity in the outline?

Note: After trying to delete the activity inside the tree view, then returning to the outline and deleting it there, the activity id gets displayed. This no longer occurs upon page refresh.
screen shot 2018-12-14 at 14 10 43

ibebic:
This resolves #149 .

When opening an activity deletion modal from the Outline, it will display the activity name only, but when opening the modal from the Tree view, it will display both the activity id and name (e.g. `Are you sure you want to delete activity 1051: Random Goal?`

### DIFF
--- a/client/components/common/ConfirmationModal.vue
+++ b/client/components/common/ConfirmationModal.vue
@@ -5,7 +5,7 @@
     </div>
     <div slot="body">
       Are you sure you want to delete
-      <span v-if="info">{{ context.type }} <b>{{ info }}</b></span>
+      <span v-if="context.info">{{ context.type }} <b>{{ context.info }}</b></span>
       <span v-else>this {{ context.type }}</span>?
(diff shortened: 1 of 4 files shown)


kjuej:
@ibebic This seems to be working correctly. The element deletion modal is more consistent.

PR #157, opened 2018-12-20
Expand upper activity on outline only if new activity is its child

### LINKED ISSUE #97: Minor issue when adding new elements
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request

## Context
After adding a new element below an element which contains sublevels (Make sure the added element is not a sublevel of the element above it, and that the element above it does not have its sublevels visible) all of the sublevels of the element above the added element become visible.

## Steps to Reproduce
1. Create a course that has multiple levels
2. Add an element and a couple of sublevels to that element
3. Hide the sublevels of that element
4. Add another element below that element (make sure it's not a sublevel of that element) - Notice after adding this element, all of the sublevels of the element above it become visible

ibebic:
This resolves #97 .

When adding a new activity on the outline below another activity, that upper activity will now expand only if the newly added activity is its child.

### DIFF
--- a/client/components/course/Outline/InsertActivity/index.vue
+++ b/client/components/course/Outline/InsertActivity/index.vue
@@ -88,7 +88,7 @@ export default {
       activity.parentId = this.resolveParent(activity);
       activity.position = this.calculatePosition(activity);
       this[this.action](activity);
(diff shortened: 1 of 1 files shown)


underscope:
🎖 Thanks

PR #161, opened 2019-01-07
Calculate and include dimensions when saving an image teaching element

2Pacalypse-:


client/components/editor/teaching-elements/Image.vue
```diff
@@ -43,6 +43,15 @@ function toDataUrl(imageUrl) {
   return imgSrcToDataURL(imageUrl, 'image/png', 'Anonymous');
 }
 
+function getImageDimensions(url) {
+  return new Promise((resolve, reject) => {
+    const img = new Image();
+    img.onload = () => resolve({ width: img.width, height: img.height });
+    img.onerror = (event) => reject(event);
```
vladimyr: You can omit parentheses here; also keep in mind that `event` is not an `Error` instance and rejecting with it is considered antipattern.

2Pacalypse-: Generally I would agree with you, but I did it like this because `event` here is of the type `error` and transforming it into an `Error` instance would make it lose some pertinent information.

If you think that's not a big deal, I can still do something like:
```js
new Error('Error loading the image')
```

2Pacalypse-: Hmm, but putting custom properties on an `Error` instance still means that the `.catch` block needs to be aware of what it gets. However, I do agree that it's perhaps more intuitive to send custom `Error` instances than events of the type `error`, so I'll do that.

underscope:
@vladimyr / @2Pacalypse- 

Please merge into latest develop. New root for content elements: `https://github.com/ExtensionEngine/tailor/tree/develop/client/components/content-elements`

2Pacalypse-:
@underscope done

underscope:
Thanks!

PR #181, opened 2019-02-25
Updated "like" filter

kjuej:


server/course/course.controller.js
```diff
@@ -14,7 +15,7 @@ const sample = require('lodash/sample');
 const DEFAULT_COLORS = ['#689F38', '#FF5722', '#2196F3'];
 
 function index({ query, user, opts }, res) {
-  if (query.search) opts.where.name = { $iLike: `%${query.search}%` };
+  if (query.search) opts.where.name = { [Op.like]: `%${query.search}%` };
```
vladimyr: You mean `[Op.iLike]`? `i` stands for `ignore case` - case insensitive `LIKE`.

PR #190, opened 2019-03-05
Build system improvements

vladimyr:
This PR introduces improvements to `poi` powered build system in terms of:
- fixing bugs
- ensuring correct dist directory layout
- ensuring that 3rd party libs don't recreate their original `node_module` paths inside dist directory


kjuej:
@vladimyr @underscope Seems to be working as expected.

PR #193, opened 2019-03-08
Sort elements

franosincic:
Registry initialization doesn't guarantee same order of elements every time, so on adding element/assessment elements will be sorted by position in configuration file.

### DIFF
--- a/client/ElementRegistry.js
+++ b/client/ElementRegistry.js
@@ -18,27 +18,30 @@ export default class ElementRegistry {
   }
 
   async initialize() {
-    await Promise.map(elementList, location => this.load(location));
-    const extensions = await this.loadExtensionList();
-    await Promise.map(extensions, location => this.load(location, true));
+    const elements = elementList.map(location => ({ location }));
+    const extensionList = await this.loadExtensionList();
+    const extensions = extensionList.map(location => ({ location, isExtension: true }));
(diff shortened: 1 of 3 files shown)


vladimyr:
@underscope Is this ok to get merged?

underscope:
I communicated with @franosincic desired order.

client/components/common/tce-core/AddElement/SelectAssessment.vue
```diff
@@ -30,7 +31,7 @@ export default {
   },
   computed: {
     rows() {
-      return chunk(this.elements, this.rowSize);
+      return chunk(sortBy(this.elements, 'position'), this.rowSize);
```
underscope: Might make sense to do the sorting on registry side upon import.

client/ElementRegistry.js
```diff
@@ -18,12 +18,16 @@ export default class ElementRegistry {
...
+    });
     const extensions = await this.loadExtensionList();
-    await Promise.map(extensions, location => this.load(location, true));
+    await Promise.map(extensions, (location, position) => {
+      return this.load(location, elementList.length + position, true);
+    });
   }
 
-  async load(location, isExtension) {
+  async load(location, position, isExtension) {
```
underscope: Move isExtension and position to options object. Set position to last as a default.

PR #194, opened 2019-03-13
Reflections - non gradable assessments

franosincic:
Introduce new element type Reflection that allows Single & Multiple Choice, True - False, Text Response and Fill the Blank to be non gradable assessment subtypes.


client/components/content-elements/tce-fill-blank/edit/index.vue
```diff
@@ -65,7 +66,7 @@ export default {
       return this.assessment.correct;
     },
     hasAnswers() {
-      return this.answerGroups.length > 0;
+      return get(this.answerGroups, 'length');
```
underscope: Convert to boolean

client/components/common/tce-core/AddElement/SelectElement.vue
```diff
@@ -59,17 +60,25 @@ export default {
...
+          type: 'REFLECTION',
+          ui: { icon: 'mdi-comment-question-outline' }
         });
       }
       if (!this.include) return result;
       return filter(result, it => includes(this.include, it.type));
     },
     assessments() {
-      return filter(this.registry, { type: 'ASSESSMENT' });
+      const types = ['QUESTION'];
```
underscope: If the goal was to avoid 2 computed for list of questions depending on which type was selected then computed name should be questions.

client/components/common/tce-core/Assessment/Feedback/FeedbackItem.vue
```diff
@@ -46,10 +46,16 @@ export default {
   props: {
     answer: { type: String, default: '' },
     feedback: { type: String, default: '' },
+    isGraded: { type: Boolean, default: false },
     isEditing: { type: Boolean, required: true },
     index: { type: Number, required: true },
     quillOptions: { type: Object, default: quillOptions }
   },
+  computed: {
+    title() {
```
underscope: label

client/components/common/tce-core/utils.js
```diff
@@ -1,13 +1,21 @@
...
 export function getComponentName(type) {
-  return `tce-${toCase.slug(type)}`;
+  return `tce-${toCase.slug(getElementType(type))}`;
 }
 
 export function processAssessmentType(type) {
   return `assessment-${toCase.slug(type)}`;
 }
 
+export function isAssessment(type) {
```
underscope: isQuestion ?

client/components/common/tce-core/Assessment/index.vue
```diff
@@ -113,7 +122,10 @@ export default {
       if (!this.isEditing) return;
       this.errors = [];
       this.validate().then(() => {
-        this.$emit('save', cloneDeep(this.editedElement.data));
+        const elementData = this.isGraded
+          ? cloneDeep(this.editedElement.data)
+          : cloneDeep(omit(this.editedElement.data, ['correct']));
```
underscope: might be nicer if you just clone it and then omit if graded:
```javascript
const elementData = cloneDeep(this.editedElement.data);
if (this.isGraded) omit(elementData....

```

franosincic:
@underscope , @vladimyr
I tried to figure out reasonable naming convention to incorporate new element types (Question & Reflection), but not sure if we could have it without renaming at least ```Assessment``` and ```SelectAssessment``` components. There is always some naming discrepancy between components and properties.

client/components/common/tce-core/Assessment/index.vue
```diff
@@ -113,7 +122,9 @@ export default {
       if (!this.isEditing) return;
       this.errors = [];
       this.validate().then(() => {
-        this.$emit('save', cloneDeep(this.editedElement.data));
+        const elementData = cloneDeep(this.editedElement.data);
+        if (!this.isGraded) delete elementData.correct;
```
underscope: Can we hit this scenario if we init data based on type and we disable correct input?

client/components/common/tce-core/utils.js
```diff
@@ -1,13 +1,21 @@
...
 
 export function processAssessmentType(type) {
   return `assessment-${toCase.slug(type)}`;
 }
 
+export function isQuestion(type) {
+  return ['QUESTION', 'REFLECTION', 'ASSESSMENT'].includes(type);
+}
+
+export function getElementType(type) {
```
underscope: Rename to processElementType / resolveElementType and document the purpose (Might have common wrapper for multiple element types). We also might consider renaming root element to something other since Assessment is a sub-class.

client/components/common/tce-core/AddElement/SelectElement.vue
```diff
@@ -20,8 +20,8 @@
       </div>
     </div>
     <select-assessment
```
underscope: Change component name

client/components/common/tce-core/AddElement/SelectElement.vue
```diff
@@ -20,8 +20,8 @@
...
-      v-if="showAssessments"
-      :assessments="assessments"
+      v-if="showQuestions"
+      :assessments="questions"
```
underscope: change prop name

client/components/common/tce-core/AddElement/SelectElement.vue
```diff
@@ -56,20 +57,29 @@ export default {
...
+          ui: { icon: 'mdi-comment-question-outline' }
         });
       }
       if (!this.include) return result;
       return filter(result, it => includes(this.include, it.type));
     },
-    assessments() {
-      return filter(this.registry, { type: 'ASSESSMENT' });
+    questions() {
+      const types = ['QUESTION'];
```
underscope: No need to compute it if type is not set. Might be nicer as:
```javascript
if (!isQuestion(this.type)) return [];
const types = ['QUESTION', this.type];
return filter(this.registry, it => types.includes(it.type));
```

client/components/common/tce-core/AddElement/SelectElement.vue
```diff
@@ -97,10 +107,6 @@ export default {
       this.type = null;
     }
   },
-  created() {
```
underscope: We should have a way to only show question types if container was configured to only accept `ASSESSMENT` or `REFLECTION`.

client/components/common/tce-core/AddElement/SelectElement.vue
```diff
@@ -56,20 +57,29 @@ export default {
       return Math.min(this.elements.length, this.rowSize);
     },
     elements() {
-      const result = filter(this.registry, it => it.type !== 'ASSESSMENT');
-      if (this.assessments.length) {
+      const result = filter(this.registry, it => !isQuestion(it.type));
+      if (this.questions.length) {
         result.push({
```
underscope: Should we check individual subtypes (`ASSESSMENT`, `REFLECTION`) before we insert them?

PR #199, opened 2019-03-19
Toolbar's element state fix

### LINKED ISSUE #196: Unable to save teaching element changes in a specific scenario
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request

## Context
Save button will be disabled if the element URL is changed to its previous value.

## Steps to Reproduce
1. Add a PDF element with a valid URL to any course in tailor and refresh the page
2. Change that PDF's URL to any URL and save changes
3. Change the PDF's URL again to the URL it was on step 1 and notice the save button is disabled

mcapeta:
This PR fixes #196 and the bug regarding the toolbar's element (data) resolution.
Now, the toolbar always has "fresh" (updated) instance of element (and optionally its parent element), thus solving the bug which was caused by accidentally keeping the element's old state.


mcapeta:
Could we replace the `TABLE-CELL` type with the `HTML` even when adding embeds in a table?
Does that affect users of the content published with the `TABLE-CELL` type, and do we actually want to abandon the `TABLE-CELL` or it could/should become the superset of the `HTML` type in the future?

PR #202, opened 2019-03-22
Preview activity

rkusan00:
Added route to get perspective content and display it by sending data to boutique and redirecting to its preview component.


client/api/preview.js
```diff
@@ -0,0 +1,10 @@
+import request from './request';
+
+function createPreview(courseId, activityId) {
+  return request.get(`courses/${courseId}/activities/${activityId}/preview`)
+  .then(res => window.open(res.data));
```
vladimyr: Wrong indent :bowtie:

server/activity/activity.controller.js
```diff
@@ -4,7 +4,9 @@ const { Activity } = require('../shared/database');
 const { getOutlineLevels } = require('../../config/shared/activities');
 const find = require('lodash/find');
 const get = require('lodash/get');
+const got = require('got');
```
vladimyr: Rename `got` to `request`.

server/activity/activity.controller.js
```diff
@@ -59,6 +61,17 @@ function clone({ activity, body }, res) {
   });
 }
 
+function getPreview({ activity }, res) {
+  return publishingService.fetchActivity(activity)
+  .then(data => got.post(previewUrl, { json: true, body: data }))
+  .then(data => res.json(resolve(previewUrl, data.body.data)));
```
vladimyr: Wrong indent :bowtie:

server/activity/activity.controller.js
```diff
@@ -59,6 +61,17 @@ function clone({ activity, body }, res) {
...
+}
+
+function resolve(baseUrl, urlPath) {
+  const myURL = new URL(baseUrl);
```
vladimyr: See 6th example: https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#Examples

server/shared/publishing/publishing.service.js
```diff
@@ -12,6 +13,10 @@ class PublishingService {
     this.queue = new PromiseQueue(1, Infinity);
   }
 
+  fetchActivity(activity) {
+    return this.queue.add(() => fetchContainer(activity));
+  }
```
vladimyr: This has nothing to do with publishing. I understand that it was convenient to expose it through publishing service so you could reference it using publishing service singleton later on but it is definitely wrong move.

.env.example
```diff
@@ -53,7 +53,7 @@ ENABLE_DEFAULT_SCHEMA=1
 
 # String template that will be interpolated on the client using two route params,
 # repositoryId and activityId.
-PREVIEW_URL=https://myserver.com/repository/{repositoryId}/activity/{activityId}
+PREVIEW_URL=https://myserver.com/api/v1/preview/
```
vladimyr: How about: `http://boutique.localhost/api/v1/preview` so reader knows it points to boutique-esque LMS, additionally comment above does not longer apply :bowtie:

server/activity/activity.controller.js
```diff
@@ -59,6 +61,17 @@ function clone({ activity, body }, res) {
   });
 }
 
+function getPreview({ activity }, res) {
+  return publishingService.fetchActivity(activity)
```
vladimyr: Reference publishing helper directly instead of going through publishing service or at least implement this as static method so it does not get queued.

client/api/preview.js
```diff
@@ -0,0 +1,10 @@
+import request from './request';
+
+function createPreview(courseId, activityId) {
+  return request.get(`courses/${courseId}/activities/${activityId}/preview`)
+    .then(res => window.open(res.data));
```
vladimyr: Opening new window shouldn't be part of api method ⚠️ 
Move it to callsite instead :bowtie:

server/activity/activity.controller.js
```diff
@@ -59,6 +62,16 @@ function clone({ activity, body }, res) {
   });
 }
 
+function getPreview({ activity }, res) {
+  return fetchContainer(activity)
+    .then(data => request.post(previewUrl, { json: true, body: data }))
+    .then(data => res.json(resolve(previewUrl, data.body.data)));
```
vladimyr:
```suggestion
    .then(data => res.json(`${new URL(data.body.data, previewUrl)}`);
```
However don't return plain string, make it part of an object:
```js
{ location: ... }
```

client/api/preview.js
```diff
@@ -0,0 +1,10 @@
+import request from './request';
+
+function createPreview(courseId, activityId) {
+  return request.get(`courses/${courseId}/activities/${activityId}/preview`)
+    .then(res => res.data.location);
+}
+
+export default {
+  createPreview
+};
```
vladimyr: Inline this:
```js
export default { createPreview };
```

server/activity/activity.controller.js
```diff
@@ -59,6 +62,14 @@ function clone({ activity, body }, res) {
   });
 }
 
+function getPreview({ activity }, res) {
+  return fetchContainer(activity)
+    .then(data => request.post(previewUrl, { json: true, body: data }))
+    .then(data => res.json({
+      location: `${new URL(data.body.data.url, previewUrl)}`
+    }));
```
vladimyr: This is not `data` but `resp`onse:
```js
    .then(resp => res.json({
      location: `${new URL(resp.body.data.url, previewUrl)}`
    }));
```

vladimyr:
@rkusan00 After resolving PR comments please squash those commits to single one :bowtie:

PR #213, opened 2019-03-27
Add relationships filtering by activity type

ilijaveselica:
Added `allowedTypes` flag to configuration file which allows filtering `relationships` by activity type.

### DIFF
--- a/client/components/course/Sidebar/Relationship.vue
+++ b/client/components/course/Sidebar/Relationship.vue
@@ -85,8 +85,11 @@ export default {
       return get(activity, `refs.${this.type}`, []);
     },
     onRelationshipChanged(associations) {
+      const associationIds = associations.id
+        ? [associations.id]
+        : map(associations, 'id');
       let activity = cloneDeep(this.activity) || {};
-      set(activity, `refs.${this.type}`, map(associations, 'id'));
+      set(activity, `refs.${this.type}`, associationIds);
       this.update(activity);
     }
   },
--- a/client/components/course/Sidebar/Relationship.vue
+++ b/client/components/course/Sidebar/Relationship.vue
@@ -22,7 +22,9 @@
 <script>
 import { getLevel, getSchemaId } from 'shared/activities';
 import { mapActions, mapGetters } from 'vuex-module';
+import castArray from 'lodash/castArray';
 import cloneDeep from 'lodash/cloneDeep';
+import compact from 'lodash/compact';
 import every from 'lodash/every';
 import filter from 'lodash/filter';
 import get from 'lodash/get';
@@ -89,9 +91,7 @@ export default {
       return get(activity, `refs.${this.type}`, []);
     },
     onRelationshipChanged(value) {
-      const associations = Array.isArray(value)
-        ? value
-        : value ? [value] : [];
+      const associations = compact(castArray(value));
       let activity = cloneDeep(this.activity) || {};
       set(activity, `refs.${this.type}`, map(associations, 'id'));
       this.update(activity);
--- a/client/components/course/Sidebar/Relationship.vue
+++ b/client/components/course/Sidebar/Relationship.vue
@@ -3,7 +3,7 @@
     <label :for="type">{{ label }}</label>
     <multiselect
       :value="multiple ? associations : associations[0]"
-      :options="optionsGrouped"
+      :options="optionGroups"
       :searchable="searchable"
       :multiple="multiple"
       :allowEmpty="allowEmpty"
@@ -67,7 +67,7 @@ export default {
       }
       return filter(activities, it => every(conds, cond => cond(it)));
     },
-    optionsGrouped() {
+    optionGroups() {
       return map(groupBy(this.options, 'type'), (it, type) => ({
         type: parseType(type).typeName,
         activities: it
--- a/client/components/course/Sidebar/Relationship.vue
+++ b/client/components/course/Sidebar/Relationship.vue
@@ -18,7 +18,7 @@
 </template>
 
 <script>
-import { getLevel } from 'shared/activities';
+import { getLevel, getTypeName } from 'shared/activities';
 import { mapActions, mapGetters } from 'vuex-module';
 import cloneDeep from 'lodash/cloneDeep';
 import every from 'lodash/every';
@@ -41,14 +41,20 @@ export default {
     allowEmpty: { type: Boolean, default: true },
     placeholder: { type: String, default: 'Click to select' },
     allowCircularLinks: { type: Boolean, default: false },
-    allowInsideLineage: { type: Boolean, default: false }
+    allowInsideLineage: { type: Boolean, default: false },
+    allowedTypes: { type: Array, default: () => ([]) }
   },
   computed: {
     ...mapGetters(['activity', 'activities'], 'course'),
     ...mapGetters(['getLineage'], 'activities'),
     options() {
       const { allowInsideLineage, allowCircularLinks, activity: { id } } = this;
-      const activities = without(this.activities, this.activity);
+      let activities = without(this.activities, this.activity);
+      if (this.allowedTypes.length) {
+        activities = filter(activities, activity => {
+          return includes(this.allowedTypes, getTypeName(activity.type));
+        });
+      }
       const conds = [it => getLevel(it.type)];
       if (!allowCircularLinks) conds.push(it => !includes(this.getAssociationIds(it), id));
       if (!allowInsideLineage) {
@@ -63,12 +69,17 @@ export default {
     associations() {
       const ids = this.getAssociationIds(this.activity);
       return filter(this.options, it => includes(ids, it.id));
+    },
+    displayType() {
+      return this.allowedTypes.length > 1;
     }
   },
   methods: {
     ...mapActions(['update'], 'activities'),
     getCustomLabel(activity) {
-      return get(activity, 'data.name', '');
+      const name = get(activity, 'data.name', '');
+      const type = get(getLevel(activity.type), 'label');
+      return !this.displayType ? name : `${name} (${type})`;
     },
     getAssociationIds(activity) {
       return get(activity, `refs.${this.type}`, []);
--- a/client/components/course/Sidebar/Relationship.vue
+++ b/client/components/course/Sidebar/Relationship.vue
@@ -20,7 +20,7 @@
 </template>
 
 <script>
-import { getLevel, parseType } from 'shared/activities';
+import { getLevel, getSchemaId } from 'shared/activities';
 import { mapActions, mapGetters } from 'vuex-module';
 import cloneDeep from 'lodash/cloneDeep';
 import every from 'lodash/every';
@@ -48,7 +48,7 @@ export default {
     allowedTypes: { type: Array, default: null }
   },
   computed: {
-    ...mapGetters(['activity', 'activities'], 'course'),
+    ...mapGetters(['activity', 'activities', 'getConfig'], 'course'),
     ...mapGetters(['getLineage'], 'activities'),
     options() {
       const { allowInsideLineage, allowCircularLinks, activity: { id } } = this;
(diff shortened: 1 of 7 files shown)


client/components/course/Sidebar/Relationship.vue
```diff
@@ -63,19 +69,27 @@ export default {
...
+      const type = get(getLevel(activity.type), 'label');
+      return !this.displayType ? name : `${name} (${type})`;
     },
     getAssociationIds(activity) {
       return get(activity, `refs.${this.type}`, []);
     },
     onRelationshipChanged(associations) {
+      const associationIds = associations.id
+        ? [associations.id]
+        : map(associations, 'id');
```
kronicker: I think this would be more robust:
```js
const associationIds = Array.isArray(associations)
	? map(associations, 'id')
	: get(associations, 'id');
```

client/components/course/Sidebar/Relationship.vue
```diff
@@ -63,19 +69,27 @@ export default {
...
       let activity = cloneDeep(this.activity) || {};
-      set(activity, `refs.${this.type}`, map(associations, 'id'));
+      set(activity, `refs.${this.type}`, associationIds);
```
kronicker: Maybe `ref` instead of `associationIds` (since it can be one or more ids)

config/shared/activities.js
```diff
@@ -54,6 +55,10 @@ function getSchemaId(type) {
   return type.includes('/') && first(type.split('/'));
 }
 
+function getTypeName(type) {
+  return type.includes('/') && type.split('/')[1];
+}
```
kronicker: It's very similar to `getSchemaId`.
How about `parseType` that will return `{ schema, typeName }` and will be re-used in both?

client/components/course/Sidebar/Relationship.vue
```diff
@@ -63,19 +69,27 @@ export default {
     associations() {
       const ids = this.getAssociationIds(this.activity);
       return filter(this.options, it => includes(ids, it.id));
+    },
+    displayType() {
```
kronicker: Always display type, but let's change multiselect options so they are either grouped by type or have a type as a smaller info/subtitle type of label.

client/components/course/Sidebar/Relationship.vue
```diff
@@ -41,14 +41,20 @@ export default {
...
     ...mapGetters(['getLineage'], 'activities'),
     options() {
       const { allowInsideLineage, allowCircularLinks, activity: { id } } = this;
-      const activities = without(this.activities, this.activity);
+      let activities = without(this.activities, this.activity);
+      if (this.allowedTypes.length) {
+        activities = filter(activities, activity => {
+          return includes(this.allowedTypes, getTypeName(activity.type));
+        });
+      }
```
kronicker: Make this one of filter conditions like it's done below.

client/components/course/Sidebar/Relationship.vue
```diff
@@ -41,14 +41,20 @@ export default {
     allowEmpty: { type: Boolean, default: true },
     placeholder: { type: String, default: 'Click to select' },
     allowCircularLinks: { type: Boolean, default: false },
-    allowInsideLineage: { type: Boolean, default: false }
+    allowInsideLineage: { type: Boolean, default: false },
+    allowedTypes: { type: Array, default: () => ([]) }
```
kronicker: Semantically, an empty array would mean no types are allowed and `null` would mean rule not defined. That being said no types allowed would make no sense here so we could let it slide for this one.
Actually, let me tag @abasic for this to weigh in.

client/components/course/Sidebar/Relationship.vue
```diff
@@ -41,28 +44,45 @@ export default {
...
+    allowInsideLineage: { type: Boolean, default: false },
+    allowedTypes: { type: Array, default: null }
   },
   computed: {
     ...mapGetters(['activity', 'activities'], 'course'),
     ...mapGetters(['getLineage'], 'activities'),
     options() {
       const { allowInsideLineage, allowCircularLinks, activity: { id } } = this;
-      const activities = without(this.activities, this.activity);
+      let activities = without(this.activities, this.activity);
```
vladimyr: But you don't reassign it? 🤔

client/components/course/Sidebar/Relationship.vue
```diff
@@ -41,28 +44,45 @@ export default {
...
+      }));
+    },
     selectPlaceholder() {
       return isEmpty(this.options) ? 'No activities' : this.placeholder;
     },
     associations() {
       const ids = this.getAssociationIds(this.activity);
-      return filter(this.options, it => includes(ids, it.id));
+      const associations = filter(this.options, it => includes(ids, it.id));
+      return this.multiple ? associations : associations[0];
```
vladimyr: So return type of this becomes union of `Array(id)` and `id`. Make it always return array (with single member if needed) and pick first inside template depending on `multiple` prop state.

client/components/course/Sidebar/Relationship.vue
```diff
@@ -74,8 +94,11 @@ export default {
       return get(activity, `refs.${this.type}`, []);
     },
     onRelationshipChanged(associations) {
+      const ref = Array.isArray(associations)
+        ? map(associations, 'id')
+        : [get(associations, 'id')];
```
vladimyr: This becomes simple mapper once you ensure that associations is always an array.

client/components/course/Sidebar/Relationship.vue
```diff
@@ -74,8 +94,11 @@ export default {
...
       let activity = cloneDeep(this.activity) || {};
-      set(activity, `refs.${this.type}`, map(associations, 'id'));
+      set(activity, `refs.${this.type}`, ref);
       this.update(activity);
```
vladimyr: If you have time investigate could we get rid of `deepClone` here 💡

config/shared/activities.js
```diff
@@ -111,3 +108,9 @@ function getRepositoryRelationships(schemaId) {
   return flatMap(structure, it => it.relationships)
     .reduce((acc, { type }) => union(acc, [type]), []);
 }
+
+function parseType(type) {
+  const schemaId = type.includes('/') && first(type.split('/'));
+  const typeName = type.includes('/') && type.split('/')[1];
+  return { schemaId, typeName };
```
vladimyr: Umm, if it does not include slash return an empty object; oherwise just split and destructure array items forming the result.

client/components/course/Sidebar/Relationship.vue
ilijaveselica: `associations` argument is emitted from multiselect. It will be an `Object` if `multiple=false`

config/shared/activities.js
kronicker: Something like this:
```js
function parseType(type) {
  const [schemaId, typeName] = type.split('/');
  return typeName ? { schemaId, typeName } : {};
}
```

client/components/course/Sidebar/Relationship.vue
ilijaveselica: Sorry, I misunderstood your comment. It's refactored now

client/components/course/Sidebar/Relationship.vue
ilijaveselica: Why?

vladimyr: Because it pulls in quite complex lodash implementation (both heavy and slow) and in most cases we know data layout in advance plus don't have deeply nested tree.

config/shared/activities.js
vladimyr: Be aware that if typeName contains slash you'll get only first part of it.

client/components/course/Sidebar/Relationship.vue
```diff
@@ -73,9 +92,10 @@ export default {
     getAssociationIds(activity) {
       return get(activity, `refs.${this.type}`, []);
     },
-    onRelationshipChanged(associations) {
+    onRelationshipChanged(value) {
+      const associations = Array.isArray(value) ? value : [value];
       let activity = cloneDeep(this.activity) || {};
-      set(activity, `refs.${this.type}`, map(associations, 'id'));
+      set(activity, `refs.${this.type}`, filter(map(associations, 'id')));
```
vladimyr: What are you filtering here? 🙄

ilijaveselica: When deselecting item, `null` is returned. We don't want to store `[null]` to activity

client/components/course/Sidebar/Relationship.vue
```diff
@@ -55,14 +59,29 @@ export default {
...
       }
+      if (this.allowedTypes.length) {
+        conds.push(({ type }) => {
+          const { typeName } = parseType(type);
+          return includes(this.allowedTypes, typeName);
+        });
+      }
       return filter(activities, it => every(conds, cond => cond(it)));
     },
+    optionsGrouped() {
```
vladimyr: Calculated value is actually an array of option groups so let's call it `optionGroups` 💡

client/components/course/Sidebar/Relationship.vue
```diff
@@ -73,9 +88,14 @@ export default {
     getAssociationIds(activity) {
       return get(activity, `refs.${this.type}`, []);
     },
-    onRelationshipChanged(associations) {
+    onRelationshipChanged(value) {
+      const associations = Array.isArray(value) ? value : [value];
       let activity = cloneDeep(this.activity) || {};
-      set(activity, `refs.${this.type}`, map(associations, 'id'));
+      const ids = associations.reduce((acc, it) => {
+        if (it && it.id) acc.push(it.id);
```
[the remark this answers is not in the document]
vladimyr: Thing is if you deselect in multiple mode it will give you array of nulls. And instead of mapping and filtering we are doing that here in single pass.

client/components/course/Sidebar/Relationship.vue
```diff
@@ -73,9 +88,14 @@ export default {
     getAssociationIds(activity) {
       return get(activity, `refs.${this.type}`, []);
     },
-    onRelationshipChanged(associations) {
+    onRelationshipChanged(value) {
+      const associations = Array.isArray(value) ? value : [value];
```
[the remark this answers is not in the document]
vladimyr: Value may contain nulls, see my comment bellow.

client/components/course/Sidebar/Relationship.vue
```diff
@@ -55,8 +59,19 @@ export default {
         const lineage = this.getLineage(this.activity);
         conds.push(it => !includes(lineage, it));
       }
+      if (this.allowedTypes && this.allowedTypes.length) {
```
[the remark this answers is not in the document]
vladimyr: I'm all for using empty array as default but this implementation actually matches one of `SelectElement` and others throughout the system where defaults for whitelist arrays are nulls...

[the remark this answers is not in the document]
vladimyr: Regarding `isEmpty` point I would go into opposite direction by eliminating it completely because it is used for checking of `this.options` length which is guaranteed to be an array. So either inline that length check there or implement _one line_ `isEmpty` helper instead of pulling complex lodash implementation sounds more practical to me.

client/components/course/Sidebar/Relationship.vue
```diff
@@ -55,8 +59,19 @@ export default {
...
+        const schemaId = getSchemaId(this.activity.type);
+        const allowedTypes = this.allowedTypes.map(type => `${schemaId}/${type}`);
+        conds.push(({ type }) => includes(allowedTypes, type));
+      }
       return filter(activities, it => every(conds, cond => cond(it)));
     },
+    optionGroups() {
+      return map(groupBy(this.options, 'type'), (it, type) => ({
+        type: getLevel(type).label,
```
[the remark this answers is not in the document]
vladimyr: Then it would it make sense to rename `activities` to `options` too? 🤔

PR #216, opened 2019-04-02
Fix file upload validation

kronicker:
Introduces support for both of `v-validate`'s validation rules config styles (w/ or w/o dedicated `validate.rules` prop) and preserves backward compatibility for old schema configs.


client/components/common/FileUpload.vue
```diff
@@ -45,13 +45,20 @@ export default {
...
     sm: { type: Boolean, default: false }
   },
   data() {
     return { uploading: false };
   },
+  computed: {
+    fileFilter() {
+      const { validate: { ext, rules } } = this;
+      if (Array.isArray(ext)) return ext.join(',');
+      if (rules && Array.isArray(rules.ext)) return rules.ext.join(',');
```
vladimyr: Actually we can avoid re-implementing `vee-validate`'s internals by leveraging `Field` properties: https://baianat.github.io/vee-validate/api/field.html#api
```js
fileFilter() {
  const field = this.$validator.vFields({ name: this.id });
  if (!field || !Array.isArray(field.rules.ext)) return '';
  return field.rules.ext.join(',');
}
```

vladimyr:
@kronicker Please take a look 👀

client/directives/file-filter.js
```diff
@@ -0,0 +1,40 @@
+const isFileInput = el => el.tagName === 'INPUT' || el.type === 'file';
+const isProduction = process.env.NODE_ENV === 'production';
+
+export const install = Vue => {
+  const warn = (msg, vm) => Vue.util.warn(`[v-filefilter]: ${msg}`, vm);
+  Vue.directive('filefilter', {
```
kronicker: Directives should be named using `kebab-case`.

client/directives/file-filter.js
```diff
@@ -0,0 +1,40 @@
+const isFileInput = el => el.tagName === 'INPUT' || el.type === 'file';
+const isProduction = process.env.NODE_ENV === 'production';
+
+export const install = Vue => {
+  const warn = (msg, vm) => Vue.util.warn(`[v-filefilter]: ${msg}`, vm);
```
kronicker: Since you're always using warn in conjunction with `isProduction` check, you could've make check a part of `warn` method.

client/directives/file-filter.js
```diff
@@ -0,0 +1,40 @@
...
+      if (!value) return !isProduction && warn('Missing required argument.', vm);
+      if (value === 'auto') {
+        if (!vm.$validator) return;
+        const specifiers = readSpecifiers(vm.$validator, { name: el.name });
+        el.accept = specifiers.join(',');
+        return;
+      }
+      if (!Array.isArray(value)) {
+        return !isProduction &&
+          warn('Invalid argument type received; expected an Array or a String.', vm);
```
kronicker: You're actually expecting Array or "auto" which is a String and makes this type 'error' technically correct but IMO might be misleading in a way that consumer might assume he can send any String like `'jpg,png'` which you didn't anticipate here.

client/directives/file-filter.js
vladimyr: True that but I'm soooo _satisfied_ with looks of `v-filefilter` 😏

client/directives/file-filter.js
vladimyr: I might but this looks more _standardish_ from directive authoring perspective 🙃

client/directives/file-filter.js
vladimyr: Indeed, so here I'm refactoring it 👷

vladimyr: This should be fixed now 'cause `v-filefilter` now allows all valid `accept` configuration inputs.

PR #224, opened 2019-04-15
User profile settings

lukasaric:
- Implemented user profile feature with basic information and password modification
- Allows various editable information fields and profile image setup


client/components/user/Settings/UserGeneral.vue
```diff
@@ -0,0 +1,194 @@
...
+      }];
+    }
+  },
+  methods: {
+    ...mapActions(['updateInfo']),
+    updateKey(key, value) {
+      const data = cloneDeep(this.user);
+      this.updateInfo(set(data, key, value));
+    },
+    saveCroppedImage(file) {
```
lukasaric: Profile photo handling actions still in progress. I read croppa docs, still little confused on how to set image permanently (after refresh). I managed to send image data to server, then the multer does his job and returns the object with the key, url, and publicUrl. Furthermore, I still do not know how to extract an img element from croppa. Because of that I can not set url from the returned response on image src. Any suggestions?

PR #244, opened 2019-04-24
Filesystem storage provider

lukasaric:


vladimyr:
@lukasaric You pushed too many commits here so I'm closing this in favor of #245 
As a reminder: please use git-flow conventions when naming your branches.

PR #250, opened 2019-04-26
Fix postgres version check

vladimyr:
This PR fixes postegres version checking by adding version constraint to `package.json` `engines` map and adding missing `semver` dependency. :bug:

### DIFF
--- a/server/shared/database/index.js
+++ b/server/shared/database/index.js
@@ -130,15 +130,13 @@ function getConfig(sequelize) {
 }
 
 function checkPostgreVersion(sequelize) {
-  const type = sequelize.QueryTypes.VERSION;
-  return sequelize.query('SHOW server_version', { type })
-    .then(version => {
-      logger.info({ version }, 'PostgreSQL version:', version);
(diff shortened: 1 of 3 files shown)


server/shared/database/index.js
```diff
@@ -130,15 +130,13 @@ function getConfig(sequelize) {
...
-    .then(version => {
-      logger.info({ version }, 'PostgreSQL version:', version);
-      const range = pkg.engines && pkg.engines.postgres;
-      if (!range) return;
-      if (semver.satisfies(semver.coerce(version), range)) return;
-      const err = new Error(`"${pkg.name}" requires PostgreSQL ${range}`);
-      logger.error({ version, required: range }, err.message);
-      return Promise.reject(err);
-    });
+  return sequelize.queryInterface.databaseVersion().then(version => {
```
vladimyr: Using _hidden_ gems: https://github.com/sequelize/sequelize/blob/v5.6.1/lib/query-interface.js#L124-L138

PR #253, opened 2019-04-29
Last position fix

kjuej:


client/components/editor/structure/TesList.vue
```diff
@@ -72,8 +71,7 @@ export default {
       });
     },
     lastPosition() {
-      const lastItem = last(this.list);
-      return lastItem ? lastItem.position + 1 : 1;
+      return this.list.length + 1;
```
underscope: Given 2 elements, if you sort them couple of times (move 1 after 2 and then repeat the action) lastPosition won't be the right one because last one is last.position + 1. Not length + 1.

client/components/editor/structure/TesList.vue
```diff
@@ -83,22 +77,18 @@ export default {
     },
     onHiddenElementDrawer() {
       this.isElementDrawerVisible = false;
-      this.insertPosition = this.lastPosition;
+      this.insertPosition = null;
```
vladimyr: Well, this means `isElementDrawerVisible` has esentially become computed property?
```js
computed: {
  isElementDrawerVisible: ({ insertPosition }) => insertPosition !== null
}
```

client/components/editor/structure/TesList.vue
```diff
@@ -83,22 +77,18 @@ export default {
     },
     onHiddenElementDrawer() {
```
vladimyr: Can we rename this to `onCloseElementDrawer` to maintain parity with `showElementDrawer`? :bowtie:
Also, it _requires_ changing event from `hidden` to `close`:
```vue
@close="onCloseElementDrawer"
```

client/components/editor/structure/TesList.vue
```diff
@@ -83,22 +77,18 @@ export default {
...
-      this.$emit(type, element);
-    }
-  },
-  watch: {
-    lastPosition: {
-      handler(val) { this.insertPosition = val; },
-      immediate: true
+    insertElement(element) {
+      let { insertPosition: position, list } = this;
+      if (!Number.isInteger(position)) position = list.length - 1;
```
vladimyr: Ummm use `const` :grin:
How we can end up here with `insertPosition` being unset? :thinking:
If that's possible I'm more inclined towards preventing it than ensuring it here...

underscope:
Don't forget to handle composites ;)

client/components/editor/structure/TesList.vue
```diff
@@ -44,60 +42,52 @@
...
-        handle: '.drag-handle',
-        scrollSpeed: 15,
-        scrollSensitivity: 125
-      });
-    },
-    lastPosition() {
-      const lastItem = last(this.list);
-      return lastItem ? lastItem.position + 1 : 1;
-    }
+    options: ({ dragOptions }) => ({ ...dragOptions, ...draggableConfig }),
```
zcuric: Can we check if the `dragOptions` is ever used? As @abasic mention in #350, if prop isn't use, there is no point of declaring it. Cause we use the the `draggableConfig` anyway.

client/components/common/tce-core/ElementList.vue
```diff
@@ -42,21 +41,13 @@ export default {
...
-        handle: '.drag-handle',
-        scrollSpeed: 15,
-        scrollSensitivity: 125
-      };
-    },
-    nextPosition() {
-      const lastItem = last(this.elements);
-      return lastItem ? lastItem.position + 1 : 1;
-    }
+    options: () => ({
```
zcuric: Do need this computed here? Maybe just move the object in the template `:option` prop.

PR #254, opened 2019-04-30
Add progress status for publish course element

kjuej:


client/components/common/mixins/publish.vue
```diff
@@ -26,8 +27,10 @@ export default {
       this.isPublishing = true;
       return Promise.each(activities, activity => {
         this.publishStatus = `Publishing ${activity.data.name}`;
-        return this.publishActivity(activity);
+        return this.publishActivity(activity)
+          .then(() => this.activityNum++);
```
vladimyr: Couldn't this be written as? :thinking:
```js
return Promise.each(activities, (activity, i) => {
  this.progress = i; // assuming that `activityNum` gets renamed to `progress` ofc ;)
  this.publishStatus = `Publishing ${activity.data.name}`;
  return this.publishActivity(activity)
})
```

client/components/common/mixins/publish.vue
```diff
@@ -26,8 +27,10 @@ export default {
...
       }).then(() => {
+        this.activityNum = 0;
```
vladimyr: Ok, but please put this (and following lines too) in `finally` block instead to make sure this gets executed on failure too. :bowtie:

client/components/common/mixins/publish.vue
```diff
@@ -9,6 +9,7 @@ const prefix = message => `Are you sure you want to publish ${message}`;
 export default {
   data() {
     return {
+      activityNum: 0,
       isPublishing: false,
```
vladimyr: `isPublishing` is merely an computed value calculated from `publishStatus`, isn't it?
Personally it makes more sense to me to use something like:
```js
data() {
  return {
    publishStatus: {
      progress: 0,
      message: ''
    }
  }
},
computed() {
  isPublishing: ({ publishStatus }) => publishStatus.progress > 0;
}
```

client/components/common/mixins/publish.vue
vladimyr: If you go with `publishStatus: Object` option this becomes:
```js
return Promise.each(activities, (activity, progress) => {
  const message = `Publishing ${activity.data.name}`;
  Object.asssign(this.publishStatus, { progress, message });
  return this.publishActivity(activity)
})
```

client/components/common/ProgressDialog.vue
```diff
@@ -8,7 +8,7 @@
         <v-card-text>
           Please wait...
           <v-progress-linear
-            indeterminate
+            v-model="status"
```
underscope: If progress is controlled via prop then it might make sense to bind it to `:value` instead of using v-model

client/components/common/mixins/publish.vue
```diff
@@ -23,14 +26,13 @@ export default {
...
+        const progress = (i + 1) / activities.length;
+        const message = `Publishing ${activity.data.name}`;
+        this.publishStatus = { progress, message };
         return this.publishActivity(activity);
-      }).then(() => {
-        this.isPublishing = false;
-        this.publishStatus = '';
-      });
+      })
+      .finally(() => (this.publishStatus = { progress: 0, message: '' }));
```
underscope: Might make sense to extract default state since it is reused.

client/components/course/Settings/index.vue
```diff
@@ -57,7 +57,10 @@ export default {
       showCloneModal: false
     };
   },
-  computed: mapGetters(['course', 'outlineActivities'], 'course'),
+  computed: {
+    ...mapGetters(['course', 'outlineActivities'], 'course'),
+    publishPercentage: ({ publishStatus }) => publishStatus.progress * 100
```
underscope: Why not convert it to percentage upfront?

vladimyr: That's actually my fault. Because `[0..1]` is _agnostic_ range that allows you to easily transpose to other potential ranges required by particular component. Yeah in our case we deal with `v-loader` only so assuming `[0-100]` is safe bet but still 😁

PR #256, opened 2019-05-06
Upgrade stylelint to v10

vladimyr:
This PR upgrades [`stylelint`](https://npm.im/stylelint) to latest version together with
some presets packages.


vladimyr:
Only major upgrade here is [`stylelint-order`](https://npm.im/stylelint-order) from `v2` to `v3`: https://github.com/hudochenkov/stylelint-order/releases/tag/3.0.0 but that shouldn't introduce any changes inside ordering presets 🤔

PR #257, opened 2019-05-06
Split PM2 run & deploy configuration

vladimyr:
This PR introduces common `ecosystem.config.js` that gets constructed in runtime from:
1. populated `pm2.deploy.toml` (in case of pm2 driven deployment)
2. `$PM2_CONFIG_var` environment variables (defining pm2 runtime params like `exec_mode` or `instances` count)
3. defaults baked into `ecosystem.config.js`

In order to inspect generated config simply run _builder_ as node program:
`$ node ecosystem.config.js`
and you'll get resulting JSON on stdout. 🎉 

This allows us to run project locally and on the server using pm2 without doing anything besides setting few env vars inside `.env`. Plus it makes sure that our `.env` values are injected into pm2 running environment.
In case of deployment additional `pm2.deployment.toml` gets in play which is:
- git ignored by default
- if no `ref` is specified it uses default branch you are on 🎉


vladimyr:
/cc @marinko-peso
Please review this and let me know does it work in your case 🤞

PR #260, opened 2019-05-10
Express async errors

lukasaric:


server/app.js
```diff
@@ -7,6 +7,7 @@ const helmet = require('helmet');
 const origin = require('./shared/origin');
 const passport = require('passport');
 const path = require('path');
+require('express-async-errors');
```
abasic: @vladimyr On line 14. there is a require that is assigned to an unused variable. Is there a reason why we would want to do this in different ways at different times or should we consolidate it?

vladimyr: I'm all for removing it and just requiring stuff at appropriate place without that eslint pragma above.

vladimyr:
@lukasaric Please resolve reported conflicts :wink:

server/app.js
lukasaric: > @lukasaric Please resolve reported conflicts 😉

@vladimyr unused auth variable problem with unnecessary eslint pragmas fixed :point_down: .
Question, should we refactor auth/passport setup like in boutique, or should we leaving it as it is?

vladimyr: If that means new pr go for it 😁

lukasaric: :+1:

vladimyr:
Recreated PR due to snyk test failure 🤦‍♂ 
@abasic Leaving to you to hit the green button 😉

PR #261, opened 2019-05-13
Modified middleware routes

lukasaric:
- Refactored routes with unnecessary wildcards


server/comment/index.js
```diff
@@ -30,19 +30,19 @@ router
   .patch(canEdit, ctrl.patch)
   .delete(canEdit, ctrl.remove);
 
-function getComment(req, res) {
+function getComment(req, _res, next) {
```
vladimyr: This should also become `param`.

PR #263, opened 2019-05-20
Added support for m4v video format

abasic:


client/components/content-elements/tce-video/edit/index.vue
```diff
@@ -48,7 +48,8 @@ const MediaError = window.MediaError;
 
 const NOT_NATIVE = /youtu\.?be|vimeo/;
 const CUSTOM_SUBTYPE_MAPPING = {
-  ogv: 'ogg'
+  ogv: 'ogg',
+  m4v: 'mp4'
```
vladimyr: Hmm, this is bit _controversial_ in sense that m4v technically is a mp4 container but Apple marks these with `video/x-m4v` while rest of world sticks with standardized `video/mp4`: https://stackoverflow.com/a/15279480
It would be nice to add comment note here to prevent confusion 💡

abasic: I saw the same discussion but neglected to consider the need for an explanation. I will add a comment with the link to the StackOverflow discussion.

PR #265, opened 2019-05-24
UX improvements

underscope:
- Tweaked default color schema
- Outline creation improvements
- Editor UI improvements


client/App.vue
```diff
@@ -40,12 +40,11 @@ html {
 
 #app {
   color: rgba(0,0,0,0.87);
-  font-family: Roboto, Helvetica, Arial, sans-serif;
+  font-family: Poppins, Helvetica, Arial, sans-serif;
```
zcuric: `font-family: Poppins, Helvetica, Arial, sans-serif;`  is used in several places. Maybe it can be defined as variable `$fontFamily` in https://github.com/ExtensionEngine/tailor/blob/develop/client/assets/stylesheets/common/_variables.scss ?

underscope: WIP F&R 😉

vladimyr: It shouldn't be scattered through codebase as variable and/or literal but inherited instead.

underscope: Removed remaining ones 💄

PR #266, opened 2019-05-24
Refactor auth

lukasaric:
This PR refactors authentication using different passport approach which among everything includes local strategy


server/user/user.controller.js
```diff
@@ -29,19 +30,12 @@ function resetPassword({ body, params }, res) {
...
-    .then(user => {
-      const token = user.createToken({ expiresIn: '5 days' });
-      res.json({ data: { token, user: user.profile } });
-    });
+function login({ user }, res) {
+  const token = user.createToken({
+    audience: Audience.Scope.Access,
+    expiresIn: '5 days'
+  });
+  res.json({ data: { token, user: user.profile } });
```
franosincic: return? 😀

server/shared/auth/audience.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+const Scope = {
+  Access: 'scope:access',
+  Setup: 'scope:setup'
+};
+
+const Audience = { Scope };
```
franosincic: Can we get one empty line between ```module.exports``` and everything else to be consistent? 😀

server/user/index.js
```diff
@@ -1,17 +1,18 @@
...
 const ctrl = require('./user.controller');
 const model = require('./user.model');
 const router = require('express').Router();
 
 router
   // Public routes:
-  .post('/users/login', ctrl.login)
+  .post('/users/login', authenticate('local'), ctrl.login)
   .post('/users/forgotPassword', ctrl.forgotPassword)
   .post('/users/resetPassword', ctrl.resetPassword)
```
kronicker: This route should validate the token. 
We're just checking whether the token exists in the database, but we don't validate token at all which makes audience, expiration and other JWT goodies we set on the token unused.
Create a middleware that verifies the JWT.
Also, after using the token it should be removed from the database to avoid the possibility to use it more than once.

lukasaric: 👍

kjuej:
@vladimyr @lukasaric Seems there's an issue loading the courses.
Screen Shot 2019-06-17 at 11 24 09

server/shared/auth/index.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
-const { ExtractJwt, Strategy } = require('passport-jwt');
 const Audience = require('./audience');
-const config = require('../../../config/server/auth');
+const config = require('../../../config/server').auth;
+const { ExtractJwt, Strategy } = require('passport-jwt');
```
lukasaric: I think multiple destructuring has higher priority import order and in this case it should be on top of everything. @vladimyr why did you move it?

kjuej:
Works on node version 8.10.0 and higher.

server/shared/auth/index.js
vladimyr: Didn't anticipate that we are following ESLint's import sorting convention :upside_down_face: Please move it up then.

vladimyr:
> Works on node version 8.10.0 and higher.

And what prevents it from working on earlier ones? :thinking:

lukasaric:
> > Works on node version 8.10.0 and higher.
> 
> And what prevents it from working on earlier ones? 🤔

Replacing `bcryptjs` with `bcrypt`packet in my early commit `Changed encryption packet`.

server/user/user.model.js
```diff
@@ -1,15 +1,18 @@
 'use strict';
 
+const Audience = require('../shared/auth/audience');
+const bcrypt = require('bcrypt');
+const { callbackify } = require('util');
```
vladimyr: @ExtensionEngine/tailor Are we ok with this? It sets minimal required version of node to `v8.2.0`: https://nodejs.org/api/util.html#util_util_callbackify_original

/cc @kjuej

server/shared/database/migrations/20190613113029-remove-user-token-field.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+const { Sequelize } = require('sequelize');
+
+const TABLE = 'user';
+const COLUMN = 'token';
+const options = { type: Sequelize.STRING, unique: true };
+
+module.exports = {
```
underscope: Rather then removing it should we use it upon reset for invalidation?

.env.example
```diff
@@ -21,17 +22,21 @@ DATABASE_ADAPTER=postgres
...
 AUTH_SALT_ROUNDS=10
-
-# Session secret string.
+AUTH_JWT_SCHEME=JWT
 AUTH_JWT_SECRET=
+AUTH_JWT_ISSUER=tailor
 
 # Email
-EMAIL_ADDRESS=tailor@extensionengine.com
+EMAIL_SENDER_NAME=Tailor
```
underscope: Why did we do email service refactor within auth refactor? Did we manage backwards compatibility and set deprecation warnings? This feels like a separate PR 😕 🤔

.env.example
```diff
@@ -2,6 +2,7 @@
 HOSTNAME=127.0.0.1
 PORT=3000
 PROTOCOL=http
+API_PATH=/api/v1/
```
underscope: Whats the reasoning behind this configurability? Tailor API in future should not be available for integrations. 🤔

client/api/request.js
```diff
@@ -1,13 +1,19 @@
...
+  baseURL: process.env.API_PATH,
   withCredentials: true,
   headers: { 'Content-Type': 'application/json' }
+};
+
+// Instance of axios to be used for all API requests.
+const client = axios.create(config);
+
+Object.defineProperty(client, 'base', {
+  get() {
```
underscope: Why using this approach to get singleton ?

client/api/request.js
```diff
@@ -21,7 +27,7 @@ client.interceptors.request.use(config => {
 });
 
 client.interceptors.response.use(res => res, err => {
-  if (err.response.status === 401) {
+  if (err.response && err.response.status === 401) {
```
underscope: 🎖 🤡

underscope: Btw, anybody for enabling optional chaining (https://github.com/tc39/proposal-optional-chaining)?

server/shared/database/migrations/20190613113029-remove-user-token-field.js
vladimyr: I don't get you; what do you mean by: _upon reset for invalidation_ 🤔 This system relies on _auto_-invalidation of reset tokens because their secret is generated using previous password hash.

.env.example
vladimyr: Huh, strictly speaking you are right, it should be separate PR. But if I understand it correctly only breaking change is this `EMAIL_ADDRESS` to `EMAIL_SENDER_*` transition. If so, could we just throw in some deprecation warnings here and proxy old stuff to avoid hassle of factoring those changes out? 🤡

.env.example
vladimyr: Here we just followed what we already have in boutique and primary reason is moving hardcoded value out of code: https://github.com/ExtensionEngine/tailor/blob/1abb956/client/api/request.js#L4

client/api/request.js
vladimyr: Why it is lazily created or? 😕

client/api/request.js
vladimyr: Here is what `ag` thinks of it :neckbeard: 
```
$ ag '(\w+)\s&&\s\1\.' -C1

server/router.js
32-  const { body } = req;
33:  if (body && body.email) body.email = body.email.toLowerCase();
34-  next();

client/components/content-elements/tce-table/edit/index.vue
48-    anchor.position,
49:    (adjacent && adjacent.position) || limit(anchor, direction)
50-  );

client/components/course/TreeView/index.vue
100-  const desc = find(structure, { type });
101:  return desc && desc.color;
102-}

client/components/editor/Toolbar.vue
11-    <element-toolbar
12:      v-if="element && element.parent"
13-      :key="`${element.parent._cid}-${element.id}`"

client/components/editor/structure/Assessments.vue
67-      const question = assessment.data.question;
68:      const hasQuestion = question && question.length > 0;
69-      if (this.isSelected(assessment) && !hasQuestion) {

client/components/editor/structure/Exams/AssessmentGroup.vue
103-      const question = assessment.data.question;
104:      const hasQuestion = question && question.length > 0;
105-      if (this.isSelected(assessment) && !hasQuestion) {

client/store/helpers/collection.js
49-        const previous = this.context.state.items[model._cid];
50:        if (previous && previous._version === model._version) model._synced
 = true;
51-        this.commit('save', model);
```

underscope: Because we use get 😉

.env.example
underscope: 👍🏻

.env.example
underscope: Config still feels like adding it to the wrong place since it doesn't make sense for it to be configured. Or does it 🤔?

server/shared/database/migrations/20190613113029-remove-user-token-field.js
underscope: Missed that change 👍🏻

underscope: Make sure INTEGRATION user is handled since this will obviously invalidate tokens for preview. Set proper warnings. We should have drafts for breaking changes / release notes so we don't forget this.

client/api/request.js
vladimyr: You lodash people 👾

server/shared/database/migrations/20190613113029-remove-user-token-field.js
vladimyr: Tbh I totally forgot about integration users but we should be able to handle them differently by checking their audience maybe?

.env.example
vladimyr: Semantically speaking you are right which leads us to declaring it as global inside poi config?

client/api/request.js
kronicker: > Btw, anybody for enabling optional chaining ([tc39/proposal-optional-chaining](https://github.com/tc39/proposal-optional-chaining))?

Why not? I'm up for it. It's in stage 2 which means it's arriving in the spec soon.

PR #270, opened 2019-05-31
Styles refactor according to stylelint

lukasaric:
This PR refactors vue components on client according to `stylelint` rules.

### DIFF
--- a/client/assets/stylesheets/common/_variables.scss
+++ b/client/assets/stylesheets/common/_variables.scss
@@ -7,3 +7,5 @@ $font-weight-bold: 700;
 
 // Font family
 $font-family-icons: "Material Design Icons";
+$font-family-primary: Poppins, Helvetica, Arial, sans-serif;
+$font-family-secondary: Roboto, Helvetica, Arial, sans-serif;
--- a/client/assets/stylesheets/common/_variables.scss
+++ b/client/assets/stylesheets/common/_variables.scss
@@ -4,3 +4,6 @@ $font-weight-regular: 400;
 $font-weight-medium: 500;
 $font-weight-semibold: 600;
 $font-weight-bold: 700;
+
+// Font family
+$font-family-icons: "Material Design Icons";
--- a/client/components/common/Meta/Switch.vue
+++ b/client/components/common/Meta/Switch.vue
@@ -109,9 +109,9 @@ input[type=checkbox] {
   &::after {
     background-color: $unchecked;
     box-shadow:
-      0 3px 1px -2px rgba(0,0,0,0.2),
-      0 2px 2px 0 rgba(0,0,0,0.14),
-      0 1px 5px 0 rgba(0,0,0,0.12);
+      0 3px 1px -2px rgba(0, 0, 0, 0.2),
+      0 2px 2px 0 rgba(0, 0, 0, 0.14),
+      0 1px 5px 0 rgba(0, 0, 0, 0.12);
   }
 }
 
@@ -133,7 +133,7 @@ label {
 
   .lever:active::before {
     transform: scale(2.4);
-    background-color: rgba(0,0,0,0.08);
+    background-color: rgba(0, 0, 0, 0.08);
 
     .checked & {
       transform: scale(2.4);
--- a/client/components/common/Meta/Switch.vue
+++ b/client/components/common/Meta/Switch.vue
@@ -75,23 +75,6 @@ input[type=checkbox] {
   opacity: 0;
 }
 
-label {
-  position: relative;
-  cursor: pointer;
-
-  &.checked .lever {
-    background: $lever-checked;
-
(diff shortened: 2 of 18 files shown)


client/components/course/Revisions/EntitySidebar.vue
```diff
@@ -56,6 +56,11 @@ export default {
 <style lang="scss" scoped>
 $revision-padding: 32px;
 
+.selected-revision {
+  color: #fff !important;
```
abasic: Can this be done with the `!import` rule?

lukasaric: You mean without `!important` rule? I tried, but main selector `.revision` has color on it's own so when I extend this class on `.selected.revision:hover`, it overrides color applied on this selector.

abasic: The `!important` rule can be avoided in two ways:
1) We use a `mixin` instead of an extended `class`.
2) we continue to use `@extend` on the `.selected-revision` `class` but we move the definition of the `class` below the definition of the `.revision` `class`.

The reason why both would work is the cascading rule of `CSS`. Currently, it does not work because the definition of the `.selected-revision` `class` is above the `.revision`. The way `class` extension works is like this:
 - it takes the names of the `classes` that are extending a `class` and add them to the definition of the rules for the extended `class`. Example:
```
 .a {
   margin: 0;
 }
 
 .b {
   margin: 10px;
 }
 
 .c {
   @extend .a;
   
   color: white;
 }
```
 
 becomes:
 
```
 .a, .c {
   margin: 0;
 }
 
 .b {
   margin: 10px;
 }
 
 .c {
   color: white;
 }
```
 
So, even though in the `SCSS` the definition of the `.c` `class` is below the `.b` `class`. in the final version of the `CSS` the margin rule for the `.c` is above the one for the `.b`. In this case, if there were a tag that had both of these `classes`, the margin rule of the `.b` `class` would be used. The reason why a `mixin` would work is that the `mixins` code is inserted into the `class` definitions or moving the `class` definition below would work is because the extension happens after the `.revision` `class` definition.

tl;dr Use a `mixin` since you are not using the `.selected-revision` `class` in the `HTML`.

client/assets/stylesheets/common/_variables.scss
```diff
@@ -4,3 +4,6 @@ $font-weight-regular: 400;
 $font-weight-medium: 500;
 $font-weight-semibold: 600;
 $font-weight-bold: 700;
+
+// Font family
+$font-family-icons: "Material Design Icons";
```
kronicker: Technically we could end up using multiple icon fonts so I would rename it to `font-family-icons`

kronicker: Since you extracted icons font to a variable you can do the same with others.
As far as I see we have two options:
1. `Poppins, Helvetica, Arial, sans-serif` as the primary font and
2. `Roboto, Helvetica, Arial, sans-serif` as secondary.

client/components/course/Revisions/EntitySidebar.vue
```diff
@@ -56,6 +56,11 @@ export default {
 <style lang="scss" scoped>
 $revision-padding: 32px;
 
+@mixin selected-revision() {
```
kronicker: Parens are redundant when you have no args. Also on `@include` statements ⬇️  as well.

poi.config.js
```diff
@@ -64,6 +64,12 @@ module.exports = {
         files: ['.brandrc', '.brandrc.js'],
         imagesPath
       }
+    },
+    {
```
kronicker: Move opening bracket to the line above. Check previous occurrences and be consistent 💅

client/components/content-elements/tce-modal/edit/Modal.vue
```diff
@@ -97,15 +98,15 @@ function toggleFocusTrap(focusTrap, condition) {
   max-width: 640px;
   margin: 0 auto;
   padding: 4px;
-  font: 14px Roboto, Helvetica, Arial;
+  font: 14px $font-family-secondary;
   text-align: left;
   border: none;
   border-radius: 2px;
   box-shadow:
     0 11px 15px -7px rgba(0,0,0,0.2),
```
abasic: Missing spaces in all of the `rgba` rules.

client/components/common/Meta/Switch.vue
```diff
@@ -132,13 +115,30 @@ label {
...
-  .checked & {
+    &::before, &::after {
+      left: 18px;
+    }
+  }
+
+  .lever:active::before {
     transform: scale(2.4);
-    background-color: transparentize($checked, 0.85);
+    background-color: rgba(0,0,0,0.08);
```
abasic: Missing spaces in `rgba` rule.

PR #278, opened 2019-06-04
Meta Select options image preview

franosincic:
This PR enables preview of activity meta Select's option image (if has one).


underscope:
@kjuej ☝️

kjuej:
@franosincic @underscope @mcapeta Seems to be working as expected. 👍

PR #280, opened 2019-06-05
Snackbar upgrade

lukasaric:
- added `right` position as default value for snackbar appearance
- added `primary` snackbar prop with global primary color


client/plugins/vuetify-snackbar/index.js
```diff
@@ -10,6 +10,7 @@ export const install = (Vue, { parent = document.body } = {}) => {
   const $snackbar = {
     show: (msg, options) => queue.add(() => vm.show(msg, options)),
     close: () => vm.close,
+    primary: (msg, options) => $snackbar.show(msg, { ...options, color: 'primary' }),
```
underscope: success, info  and warning have semantics while primary doesn't.  Maybe we should config info to primary color or figure out better name for 'primary'. @vladimyr ?

PR #281, opened 2019-06-05
Add the script which imports the tailor extensions

2Pacalypse-:
After the extension is imported, the Tailor server (if online) is
notified and is expected to actually install/register that extension.


underscope:
@vladimyr @2Pacalypse ETA ?

2Pacalypse-:
If I remember correctly (and it's been a while, so I forgot a lot of things), this PR is mostly done. It just needs to replace a couple of hardcoded values and read them from `package.json`, mainly:
- name of the build script
- path to the output directory of the build script

I would have to spend a bit more time going over this to make sure I'm not forgetting anything.

underscope:
I’m closing this issue because it has been inactive for > year

PR #282, opened 2019-06-06
Extend InputAsset component to accept additional prop

hrvojevu:
`InputAssets` component is now extended with `allowFileUpload` prop to determine if `file-upload` component should be rendered. Additionally, placeholder text for URL input field is changed depending on the prop. `allowFileUpload` defaults to `true` to preserve current behaviour.

### DIFF
--- a/client/components/common/InputAsset.vue
+++ b/client/components/common/InputAsset.vue
@@ -11,6 +11,7 @@
       <v-icon>mdi-open-in-new</v-icon>
     </v-btn>
     <file-upload
+      v-if="allowFileUpload"
       v-show="!file && isEditing"
       :uploading.sync="uploading"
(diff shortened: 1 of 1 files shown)


client/components/common/InputAsset.vue
```diff
@@ -73,7 +74,8 @@ export default {
     url: { type: String, default: null },
     publicUrl: { type: String, default: null },
     extensions: { type: Array, required: true },
-    uploadLabel: { type: String, default: 'Select file' }
+    uploadLabel: { type: String, default: 'Select file' },
+    allowFileUpload: { type: Boolean, default: true }
```
franosincic: ```uploadLabel``` is used only for file upload so ```allowFileUpload``` should be placed above it (props should be ordered by importance)

PR #283, opened 2019-06-10
Activity & TE restore

lukasaric:
This PR provides restore option for each activity & teaching element in `revision history` tab.


server/revision/revision.controller.js
```diff
@@ -15,14 +17,42 @@ function index({ course, query }, res) {
...
     return res.json(revision);
   });
 }
 
+function restoreActivity(id, opts, include, res) {
+  const revisionOpts = { limit: 1, include, order: [[ 'createdAt', 'DESC' ]] };
+  return Activity.findByPk(id, { paranoid: false })
+      .then(activity => activity || createError(NOT_FOUND))
+      .then(activity => activity.removeOrRestore(opts))
+      .then(() => Revision.findAll(revisionOpts))
```
franosincic: Use ```findOne``` instead ```findAll``` with limit 1, than you can change the next line also

server/revision/revision.controller.js
```diff
@@ -15,14 +17,42 @@ function index({ course, query }, res) {
...
+      .then(data => res.json({ data: data[0] }));
+}
+
+function restoreTe(id, opts, include, res) {
+  const revisionOpts = { limit: 1, include, order: [[ 'createdAt', 'DESC' ]] };
+  return TeachingElement.findByPk(id, { paranoid: false })
+    .then(teachingElement => teachingElement || createError(NOT_FOUND))
+    .then(teachingElement => teachingElement.restore(opts))
+    .then(() => Revision.findAll(revisionOpts))
```
franosincic: Same as above comment

server/activity/activity.model.js
```diff
@@ -166,8 +166,9 @@ class Activity extends Model {
       });
   }
 
-  remove(options = {}) {
-    if (!options.recursive) return this.destroy(options);
+  removeOrRestore(opts = {}) {
+    const { recursive, soft, isRestore } = opts;
+    if (!recursive) return isRestore ? this.restore(opts) : this.destroy(opts);
```
franosincic: Instead having ```isRestore``` flag, could you just check the ```deletedAt``` or am I missing something here? On both ```Activity``` & ```TeachingElement``` models? 🤔

server/revision/revision.controller.js
```diff
@@ -15,14 +17,42 @@ function index({ course, query }, res) {
   return Revision.findAll(opts).then(data => res.json({ data }));
 }
 
+function restore({ body: { revision }, user }, res) {
+  const { id, state } = revision;
+  const opts = { recursive: true, isRestore: true, context: { userId: user.id } };
+  const include = [{ model: User, attributes: ['id', 'email'] }];
+  Revision.update({ restored: true }, { where: { id } });
+  if (revision.entity === 'ACTIVITY') return restoreActivity(state.id, opts, include, res);
```
franosincic: Why didn't you destructure ```entity``` from revision like you did ```id & state```? Why ```revision``` is property of body, why didn't you just send revision properties that you need in body?

client/store/modules/revisions.js
```diff
@@ -40,6 +42,14 @@ action(function fetch() {
   });
 });
 
+action(function restore(revision) {
+  const { id, courseId } = revision.state;
+  const url = `/courses/${courseId}/revisions/${id}/restore`;
+  this.commit('update', { ...revision, restored: true });
+  return request.post(url, { revision })
+    .then(({ data: { data } }) => this.commit('add', data));
```
franosincic: Instead double destructuring of data there can be nice little helper😀
```js
const extractData = res => res.data.data;
```

client/components/course/Outline/index.vue
```diff
@@ -64,7 +64,13 @@ export default {
         .sort((x, y) => x.position - y.position);
     }
   },
-  methods: mapMutations(['toggleActivities'], 'course'),
+  methods: {
+    ...mapActions({ getActivities: 'fetch' }, 'activities'),
+    ...mapMutations(['toggleActivities'], 'course')
+  },
+  created() {
+    this.getActivities();
```
franosincic: If you did fetch again here because you wanted updated state after you restored something on history page then it makes more sense to do it [there where you restored it](https://github.com/ExtensionEngine/tailor/blob/7a8de47ed8cc060b0def150bfa05bef5d7b4e927/client/components/course/Revisions/RevisionItem.vue#L72-L76). Also use ```reset``` instead of ```fetch```.

client/components/course/Revisions/RevisionItem.vue
```diff
@@ -26,44 +29,37 @@ import {
...
 
 export default {
   name: 'revision-item',
   props: {
     revision: { type: Object, required: true }
   },
-  data() {
-    return { expanded: false };
-  },
+  data: () => ({
```
franosincic: This can be a one liner 😀

server/activity/activity.model.js
lukasaric: I removed `restored` field from database. But I also removed on client `isRestored` flag, so restore icon will render for all removed revision items. Furthermore, I left activity model as it was because now when using `deletedAt` instead of `isRestore` flag in `activity.model` if I click removed revision item for the second time, item will be removed in database. @abasic approved this approach for now, let me know if you reconsider.

client/store/modules/revisions.js
lukasaric: I left it as it was to be more consistent like in other store modules. Maybe refactor this in every module in new PR? :bowtie:

client/components/course/Revisions/RevisionItem.vue
```diff
@@ -26,44 +29,36 @@ import {
...
     revision: { type: Object, required: true }
   },
-  data() {
-    return { expanded: false };
-  },
+  data: () => ({ expanded: false }),
   computed: {
     ...mapGetters(['structure'], 'course'),
     ...mapGetters(['getParent'], 'activities'),
-    activity() {
```
kronicker: I don't see the benefit of going with the arrow function in this case.

client/store/modules/revisions.js
```diff
@@ -24,6 +26,7 @@ getter(function hasMoreResults() {
 });
 
 action(function fetch() {
+  if (!this.getters) return;
```
kronicker: This feels wrong... Getters are not defined, but the action is already used? Can you explain the use case?

server/activity/activity.controller.js
```diff
@@ -33,12 +33,17 @@ function list({ course, query, opts }, res) {
 }
 
 function remove({ course, activity, user }, res) {
-  const options = { recursive: true, soft: true, context: { userId: user.id } };
+  const options = {
+    recursive: true,
+    soft: true,
+    isRestore: false,
```
kronicker: IMO `restore` is enough

server/activity/activity.model.js
```diff
@@ -203,7 +205,8 @@ class Activity extends Model {
   }
 }
 
-function removeAll(Model, where = {}, soft = false) {
+function removeOrRestoreAll(Model, where = {}, soft = false, isRestore) {
+  if (isRestore) return Model.update({ detached: false }, { where });
   if (!soft) return Model.destroy({ where });
   return Model.update({ detached: true }, { where });
```
kronicker:
```js
if (!restore && !soft) return Model.destroy({ where });
return Model.update({ detached: !restore }, { where });
````

File: server/revision/revision.controller.js [hunk 12]
```diff
@@ -15,14 +17,40 @@ function index({ course, query }, res) {
   return Revision.findAll(opts).then(data => res.json({ data }));
 }
 
+function restore({ body: { state, entity }, user }, res) {
+  const opts = { recursive: true, isRestore: true, context: { userId: user.id } };
+  const include = [{ model: User, attributes: ['id', 'email'] }];
+  if (entity === 'ACTIVITY') return restoreActivity(state.id, opts, include, res);
+  return restoreTe(state.id, opts, include, res);
+}
```
kronicker: `restoreActivity` and `restoreTe` are helper methods so it makes more sense to have responses in controller method i.e. this one. This would remove this weird response argument.

```js
const action = entity === 'ACTIVITY' ? restoreActivity : restoreTe;
return action(state.id, opts, include)
  .then(data => res.json({ data });
```

client/components/course/Revisions/RevisionItem.vue
lukasaric: I used it for consistency, because other computed methods are arrow functions. Should I change it as it was before or?

kronicker: Actually not sure. I figured our convention is to use arrows where we can reduce the function to a one-liner and _normal_ otherwise.
@ExtensionEngine/tailor , thoughts?

client/store/modules/revisions.js
lukasaric: This `if` is here to fix render bug in history tab. I think that `getters` are not loaded yet, but still action is triggered. @franosincic suggested this, after this approach, error disappeared.

server/revision/revision.controller.js
lukasaric: Hmm, nice 👍

client/store/modules/revisions.js
vladimyr: This sounds like it is `vuex-module` related :thinking: However, please put comment above to investigate it further once we migrate to _raw_ `vuex`.

client/store/modules/revisions.js
vladimyr: Let's leave it for next PR :hourglass:

server/activity/activity.model.js
```diff
@@ -166,8 +166,9 @@ class Activity extends Model {
       });
   }
 
-  remove(options = {}) {
-    if (!options.recursive) return this.destroy(options);
+  removeOrRestore(opts = {}) {
+    const { recursive, soft } = opts;
+    if (!recursive) return this.deletedAt ? this.restore(opts) : this.destroy(opts);
```
zcuric: You have this in two places, right? Maybe extract it to variable:
```js
// name is 💩 , choose something better, maybe `removeOrDestroy`
const action = (opts) => this.deletedAt ? this.restore(opts) : this.destroy(opts); 
if (!recursive) return action(opts);
```
And on line https://github.com/ExtensionEngine/tailor/pull/283/files#diff-f710cd3ec887fbb8b2e0c95f34f9ba8dR191
```js
...
 .then(() => action(opts)
...
```

lukasaric: Hmm, not bad. Actually `action` name is fine with me. Suggestion? @zcuric

zcuric: Whatever, I've seen used `action` before in a similar case. Keep the consistency.

vladimyr: Isn't this second occurrence actually good candidate for something like:
```js
.then(() => this.removeOrRestore({ ...opts, recursive: false }));
```
thus avoiding duplication?

server/shared/database/migrations/20190606145833-add-revision-enum-value.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+const replaceEnum = require('sequelize-replace-enum-postgres').default;
```
vladimyr: I knew I had it somewhere:
https://github.com/ExtensionEngine/boutique/blob/9ae920/server/common/database/helpers.js#L12-L19

server/activity/activity.model.js
```diff
@@ -166,28 +166,17 @@ class Activity extends Model {
...
-            .then(() => descendants);
-        })
-        .then(descendants => {
-          const activities = map(descendants.nodes, 'id');
-          const where = { parentId: [...activities, this.id] };
-          return removeAll(Activity, where, options.soft);
-        })
-        .then(() => this.destroy(options))
-        .then(() => this);
+  removeOrRestore(options = {}) {
```
underscope, line 169: @lukasaric 
What if I restore activity where I have afterward deleted a parent as well 🤔?

client/components/course/Revisions/RevisionItem.vue
```diff
@@ -71,6 +66,15 @@ export default {
     },
     toggle() {
       if (this.isTeachingElement) this.expanded = !this.expanded;
+    },
+    restoreItem() {
+      const { name } = this.revision.state.data;
+      return this.restore(this.revision)
```
underscope, line 72: Confirmation screen?

server/activity/activity.model.js
underscope: @lukasaric 
What about publishing state of restored activity?

underscope:
Closing this one for now since it has some bigger considerations (as mentioned above),

PR #284, opened 2019-06-10
Logger modification

lukasaric:
This PR provides developer to set [bunyan](https://www.npmjs.com/package/bunyan#levels) logger `level`. Suggestion is to set level type `'debug'`. Reason for that is, for example: when developer is querying database, it will display query logs in terminal which can be very helpful.

### DIFF
--- a/server/shared/logger.js
+++ b/server/shared/logger.js
@@ -2,10 +2,15 @@
 
 const bunyan = require('bunyan');
 
+const logLevel = (() => {
+  if (process.env.LOG_LEVEL) return process.env.LOG_LEVEL;
+  return process.env.NODE_ENV === 'production' ? 'info' : 'debug';
+})();
+
 const logger = bunyan.createLogger({
   name: 'tailor-server',
   serializers: bunyan.stdSerializers,
-  level: process.env.LOG_LEVEL
+  level: logLevel
 });
 
 module.exports = logger;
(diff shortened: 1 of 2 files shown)


server/shared/logger.js
```diff
@@ -4,7 +4,8 @@ const bunyan = require('bunyan');
 
 const logger = bunyan.createLogger({
   name: 'tailor-server',
-  serializers: bunyan.stdSerializers
+  serializers: bunyan.stdSerializers,
+  level: process.env.LOG_LEVEL
```
abasic: It should default here to `info` if If `LOG_LEVEL` is not set in the `.env` file.

kronicker: @abasic How about dynamic default dependant on `NODE_ENV`?
```js
const logLevel = (() => {
  if(process.env.LOG_LEVEL) return process.env.LOG_LEVEL;
  return process.env.NODE_ENV === 'production' ? 'info' : 'debug';
})();

// ...
level: logLevel
```

abasic: I have nothing against having a different logging level based on the environment. The reason why I said it should default to `info` was that that is how it works now so I wanted to maintain parity. I assumed there was a reason why it is currently always `info`. We can go this way if nobody has a problem with it.

lukasaric: @kronicker Upgraded changing logger level based on `.env`👇 . If you reconsider this approach, let me know here, it will be done :bowtie: .

PR #287, opened 2019-06-13
Fixed tce-core imports

abasic:
This PR address the issue where some components inside the `tce-core` folder were importing parts of Tailor's common components and code. Code inside `tce-core` should only import other code from `tce-core` or `NPM` packages.

Currently `utils` are exported together but I am not sure if that is the right approach. Any feedback is appreciated.


client/components/common/ActivityBrowser/SelectActivity.vue
```diff
@@ -32,7 +32,7 @@
 
 <script>
 import activityApi from 'client/api/activity';
-import CircularProgress from 'components/common/CircularProgress';
+import { CircularProgress } from 'tce-core';
```
underscope: Can we remove it entirely and just use vuetify one which is registrated globally ❓

PR #292, opened 2019-06-21
Replace vuex-module with vuex 

parmezan38:
This PR swaps [`vuex-module`](https://npm.im/vuex-module) with native Vuex module management.


client/store/modules/course/mutations.js
```diff
@@ -0,0 +1,37 @@
+import compact from 'lodash/compact';
+import transform from 'lodash/transform';
+import Vue from 'vue';
+
+export const upsertUser = (state, user) => {
+  Vue.set(state.users, user.id, user);
+};
```
lukasaric: Also this can be one liner. He can change it according to this type of cases in other files. What do you think @vladimyr ?

client/store/modules/auth/actions.js
```diff
@@ -0,0 +1,23 @@
+import api from '../../../api/auth';
```
lukasaric: Same as above. Create another use alias `@/`.

client/store/modules/activities/actions.js
```diff
@@ -0,0 +1,49 @@
+import {
+  fetch,
+  get,
+  reset,
+  save,
+  setBaseUrl,
+  update
+} from '../../helpers/actions';
```
lukasaric: Maybe convert this multiple import into one line import to be more consistent because exports are one liner. And, also you can use alias `@/` from `poi.confing.js` and it should look like this: ` ... from '@/store/helpers/actions; `. Maybe he can create specific alias for `store` in poi ? @vladimyr

ilijaveselica:
Remove `vuex-module` from package.json since it's not used anymore

ilijaveselica:
"Toggle all" button is not working in course outline page

client/components/course/index.vue
```diff
@@ -50,14 +50,14 @@ export default {
...
-    ...mapActions({ getActivities: 'fetch' }, 'activities'),
-    ...mapMutations({ resetActivityFocus: 'focusActivity' }, 'course'),
-    ...mapMutations({ setupActivityApi: 'setBaseUrl' }, 'activities'),
-    ...mapMutations({ setupCommentsApi: 'setBaseUrl' }, 'comments'),
-    ...mapMutations({ setupRevisionApi: 'setBaseUrl' }, 'revisions'),
-    ...mapMutations({ setupTesApi: 'setBaseUrl' }, 'tes')
+    ...mapActions('course', ['getUsers']),
+    ...mapActions('courses', { getCourse: 'get' }),
+    ...mapActions('activities', { getActivities: 'fetch' }),
+    ...mapMutations('course', { resetActivityFocus: 'focusActivity' }),
```
ilijaveselica: Reorder

kjuej:
@parmezan38 Issue when trying to insert an element in-between two elements.
Screen Shot 2019-07-04 at 13 02 25

PR #294, opened 2019-06-25
JSend implementation

lukasaric:
This PR implements [jsend](https://www.npmjs.com/package/jsend) instead of standard `res.json()` usage.


kronicker:
@ExtensionEngine/tailor 
Can I be a bad guy and ask why do we want JSend in the first place? IMO, we don't get much out of it while adding a level of complexity?
I don't see a benefit over using standard HTTP codes more precisely (not just 200, 400 and 500) with appropriate payloads.
I know this forces a convention, but it actually really doesn't since maintainers still need to know JSend conventions which is another overhead IMO.

Another _bad guy_ topic I'd like to open is where do we stand on external packages? I'd prefer having packages in ee scope (both npm and github) if we made them and need to maintain them.

vladimyr:
> I don't see a benefit over using standard HTTP codes more precisely (not just 200, 400 and 500) with appropriate payloads.
> I know this forces a convention, but it actually really doesn't since maintainers still need to know JSend conventions which is another overhead IMO.

I understand your criticism so how about extending already known concepts instead: [sorry it is external 😿 ](https://www.npmjs.com/package/http-status-cats)

PR #300, opened 2019-07-02
Implement background & color format reset on html content element

vladimyr:
This PR alters color pickers on html content element toolbar to include new _format reset_ option:

image image

/cc @kronicker De nada 😉


markolovric:
Passed QA test!

PR #299, opened 2019-07-01
Handle the extensions import on the server and actually install it

2Pacalypse-:

PR #302, opened 2019-07-04
Registry creation without dynamic imports

abasic:
This version of the registry creation functions without dynamic imports. The reason for this proposed change is to allow custom teaching elements to use packages that do not exist in Tailor without requiring them to be built beforehand.

### DIFF
--- a/client/ElementRegistry.js
+++ b/client/ElementRegistry.js
@@ -6,20 +6,17 @@ import {
 } from 'tce-core/utils';
 import cloneDeep from 'lodash/cloneDeep';
 import elementList from './components/content-elements';
-import extensionsList from '../extensions/content-elements';
+import extensionList from '../extensions/content-elements';
 import find from 'lodash/find';
 import pick from 'lodash/pick';
-import Promise from 'bluebird';
 
 export default class ElementRegistry {
   constructor(Vue) {
     this._registry = [];
     this.Vue = Vue;
-  }
 
-  initialize() {
-    const list = [...elementList, ...extensionsList];
-    return Promise.map(list, (item, index) => this.load(item, index));
+    const list = [...elementList, ...extensionList];
+    list.map((item, index) => this.load(item, index));
   }
 
   load(element, position = this._registry.length) {
--- a/client/ElementRegistry.js
+++ b/client/ElementRegistry.js
@@ -6,11 +6,10 @@ import {
 } from 'tce-core/utils';
 import cloneDeep from 'lodash/cloneDeep';
 import elementList from './components/content-elements';
+import extensionsList from '../extensions/content-elements';
 import find from 'lodash/find';
 import pick from 'lodash/pick';
 import Promise from 'bluebird';
-
-const EXTENSIONS_LIST = 'index';
 
 export default class ElementRegistry {
   constructor(Vue) {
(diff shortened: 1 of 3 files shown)


2Pacalypse-:
Work on this has already been started in these two PRs: #281 and #299
The first PR implements a script which users can run to import an extension (content element) inside Tailor. The second PR is meant to implement the logic to actually make that extension available to the client. First PR is almost done (only needs to replace some hardcoded values by reading them from `package.json` of the extension, eg. name of build script, path to the output folder/files). Second PR was just an experimentation phase, so most of it will have to be rewritten.

Here are the rough steps in how this whole process should work:
1. User runs the import script, eg. `npm run import https://github.com/vladimyr/tce-example`
1.1. The repository at that link is cloned into Tailor's `extensions` folder
1.2. Dependencies of that extension are installed and its build script is run
1.3. The script makes a network request to Tailor server, letting it know about the new extension
2. After the new extension was successfully imported into Tailor, there are two ways in how it can be installed, depending on whether the server was online or offline
2.1. Tailor server was online when extension was imported
2.1.1. From the step `1.3.`, Tailor will receive a request to install the extension which will be handled in its controller
2.1.2. Controller will call an `install` method in the element registry service
2.2. Tailor server was offline when extension was imported
2.2.1. In this case, `1.3.` step will fail, but the extension was still imported into Tailor
2.2.2. When server starts again, an `initialize` method in the element registry service will be called, which will make sure all imported extensions are properly installed

Regardless of the method in which the extension is being installed (whether through the `initialize` or `install` method), the element registry service will be the central piece which will contain this logic and will be added in that second PR. The biggest change in how this feature currently works is that the element registry logic (what this PR is trying to refactor) will be moved from client to the server, so the server will be the source of truth on installed extensions.

Some of the things that the element registry service on the server will have to do are:
- Keep the list of all installed extensions
- Differentiate, in some way, the inbuilt content elements with the installed ones
- Generate an `extensions.js` and `extensions.css` files which will contain all the (compiled) javascript and css code. The client will load these files through the HTML and utilize the `__TAILOR_CONTENT_ELEMENTS__` global object to access the content elements (this is pretty much everything that the client will need to do; everything else will be server's job).
- Send an SSE event to the client to refresh the page so it gets the newest versions of above files (those files will also have to have caching turned off)
- Implement the `initialize` method which will run when the server starts. It will go through the extensions folder and generate the required files as explained above (no need for refresh event during initialization).
- Implement the `install` method, which will add the given extension to the internal list of extensions, regenerate the required files and send the refresh SSE event as explained above.
- Implement the `update` method, which will replace the existing extension with the new one, regenerate the required files and send the refresh SSE event as explained above (this can be implemented at a later date).
- Implement the `remove` method, which will find the given extension, remove it from the internal list of extensions, regenerate the required files without that extension and send the refresh SSE event as explained above (this can be implemented at a later date).

Another thing that was not talked about here is the development process of the extension, ie. content element, itself. Fortunately, we already have a working example on how an extension should look like, which users can fork and edit to fit their needs:
https://github.com/vladimyr/tce-example

So far the build process of the extension is pretty specific (requires a custom made `rollup` plugin), but it can be made more flexible in future.

Now I know that this (draft) PR was made more for the reason of getting something working before all of this can be done properly, but I think that there really isn't that much work left here that it requires an in-between solution. Unfortunately, I can't finish those two PRs in the foreseeable future, but I can supervise and offer feedback to whomever decides to tackle it. Afaik, @vladimyr already put it on his TODO list 😛

vladimyr:
Waiting for #308 to be merged ⌛️

underscope:
@vladimyr You got unblocked mister 👾 🚪

underscope:
Closing this one with 🤞 that #281 and #299 will be completed within ~ 2 weeks.

PR #307, opened 2019-07-08
Group all Quill modifications under custom theme

vladimyr:
This PR groups all [Quill](https://npm.im/quill) modifications under custom "tailor" theme. Quill constructor is dynamically injected thus removing `vue-quill-editor` dependency. 🦊 

This is stacked onto #300


vladimyr:
I've added material design icons and enabled more toolbar buttons and formatting colors 🎉 

### Before
image

<hr>

### After
image

client/components/content-elements/tce-html/edit/ui/image-embed-tooltip.js
```diff
@@ -0,0 +1,60 @@
+import './image-embed-tooltip.scss';
+import createTooltip from './tooltip';
+
+export default Quill => class ImageEmbedTooltip extends createTooltip(Quill) {
```
vladimyr: This used to be inlined inside image embed module as `EmbedTooltip`.

PR #275, opened 2019-06-04
Linking activities

zcuric:
### This PR
- [x] Adds ability to link an activity 
- [x] Adds ability to clone linked activity
- [x]  Adds ability to publish linked activity

### Several notes for developers and QA
- When one activity is being linked, two new activities are created. Both of those activities are links, one will replace the original activity, another will be linked in desired place. The original activity becomes an `origin` and it is not represent in the UI. Each link contains `originId`. 
- Activity can be linked any where the same repository/course if it satisfies the conditions which are the same as cloning. 
- Activity can be linked or created inside another linked parent. In that case, all siblings of the linked parent will receive link and if link is delete for one linked parent it will deleted for all linked parents. 
- When updating a link, updating will forward to origin and origin will be updated, but all links will receive the update. 
- When removing linked activity, linked activity will be deleted, not detached and if the number of links is 1, the origin activity will be restored to place of last remaining link. 
- Only origins are published. Published activities in the collection receive `children` array with all of children `ids` and children can have  array of `parentIds`. So children can have multiple parents and for LMS it's to represent links. 
- Cloning is also enabled. It works basically the same as normal cloning. If whole repository is cloned, links will be cloned as well. If individual nodes are cloned, links will be cloned. If there is only one link in the node, it will be resolved to origin. 
- Cloning into link is also posible.

Closes #47 #46

PR #312, opened 2019-07-11
Removing cuid and using only uuid for store

zcuric:
While working on #275, I encountered several issues related to `cuid` (`_cid`) when updating activities which are linked or adding activity into a linked tree. 

When updating a linked activity, origin gets updated, but from backend all linked activities are return, thus meaning that response is `array`. Frontend expects one updated activity with `cuid` added on frontend beforehand and attached when return from backend. In updating links this scenario doesn't work without hacks. 

Things get messy when you want to add a new activity into a linked subtree (in current state of specification for linking this is possible). You need to add one activity, from backend there activities are returned (origin and two links or more links, depending how many time one activity is linked). Again, frontend assigns  `cuid` beforehand and through manipulation reapplies the same `cuid` on return when saving in store, but in this scenario backend again returns an `array`, but frontend served only one `cuid` for one added activity. 

To make all this work, a lot of hacky updates and conditions should be applied on `vuex` store, especially in [resource helper](https://github.com/ExtensionEngine/tailor/blob/develop/client/store/helpers/resource.js) and mutations as well. 

Also, one major drawback, IMHO, that `cuid`'s are not persisted. 

Luckily, `uuid`'s are available  in the database and they can be easily replace `cuid`. Using `uuid`'s would easily solved this problem and other potential bulk operations. 

This PR demonstrates how that could work. Commit just swaps cuid with uuid in several key places, all major actions with activities work out of the box. 

Does this makes sense?

### DIFF
--- a/client/components/catalog/Container.vue
+++ b/client/components/catalog/Container.vue
@@ -14,7 +14,7 @@
     <div v-show="!searching" class="row course-list">
       <course-card
         v-for="course in orderedCourses"
-        :key="course._cid"
+        :key="course.uid"
         :course="course">
       </course-card>
       <infinite-loading ref="infiniteLoading" @infinite="loadMore">
--- a/client/components/course/Outline/Activity.vue
+++ b/client/components/course/Outline/Activity.vue
@@ -49,10 +49,10 @@
         </div>
       </div>
       <insert-activity
-        :anchor="{ id, _cid, parentId, courseId, type, position }"
+        :anchor="{ id, uid, parentId, courseId, type, position }"
         @expand="toggle(true)"/>
     </div>
-    <div v-if="!isCollapsed({ _cid }) && hasChildren">
+    <div v-if="!isCollapsed({ uid }) && hasChildren">
       <draggable
         :list="children"
         :options="{ handle: '.activity' }"
@@ -60,7 +60,7 @@
         <activity
           v-for="(subActivity, index) in children"
           v-bind="subActivity"
-          :key="subActivity._cid"
+          :key="subActivity.uid"
           :index="index + 1"
           :level="level + 1"
           :activities="activities"
@@ -86,7 +86,7 @@ export default {
   mixins: [reorderMixin],
   inheritAttrs: false,
   props: {
-    _cid: { type: String, required: true },
+    uid: { type: String, required: true },
     id: { type: Number, default: null },
     parentId: { type: Number, default: null },
     courseId: { type: Number, required: true },
@@ -119,13 +119,13 @@ export default {
       return isEditable(this.type);
     },
     isSelected() {
-      return this.focusedActivity._cid === this._cid;
+      return this.focusedActivity.uid === this.uid;
     },
     isHighlighted() {
       return this.isHovered || this.isSelected;
     },
     isExpanded() {
-      return !this.isCollapsed({ _cid: this._cid });
+      return !this.isCollapsed({ uid: this.uid });
     },
     hasSubtypes() {
       return !!size(this.config.subLevels);
@@ -134,7 +134,7 @@ export default {
       return (this.children.length > 0) && this.hasSubtypes;
     },
     showOptions() {
-      return this._cid === this.outlineState.showOptions;
+      return this.uid === this.outlineState.showOptions;
     },
     children() {
       const level = this.level + 1;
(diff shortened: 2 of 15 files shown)


kronicker:
@zcuric The whole reason for having `cid` is to have front-end defined unique id.
That allows us to have a model in the store even before it is persisted in the database so e.g. when you add the activity it is immediately shown on the UI before we receive confirmation from the API alongside `id` and `uid` generated by the database.

zcuric:
@kronicker `uuid` can be also generated on frontend and persisted in database. There is no point of having two unique ids. And `uuid` is also collision resisted.  From current state of tailor store and database, relying on `cuid` is pointless and all "magic" with `cuid` can be removed, i.e. helpers, mappers and other stuff. 

Happy path for frontend also can be perserved.

kronicker:
@zcuric I agree on this, we could ditch `cuid` for front-end generated `uuid` with fallback on postgres default generation.
@ExtensionEngine/tailor, thoughts?

vladimyr:
Not all entities in db have uuid, right? So we'll end up mixing cuids and uuids inside client codebase?

zcuric:
@vladimyr That's true. In that case, all entities would be updated to be aligned with the reset of the db. 
My proposal is to have a one unique identifier across the board.

2Pacalypse-:
Sorry in advance for going on a tangentially related rant, but as an outsider, I'd just like to say that having client-side generated IDs for resource creation is one of the most perplexing pieces of code in Tailor. I don't think I saw anyone on the internet generate a resource on the client side before the network request finishes. The benefits of this are very marginal and sometimes even detrimental to the user experience (eg. when creating a repository, a new repository is shown in UI as "created" before the loader even finishes loading).

And the downsides of maintaining a client-side ID are self-evident; now the client application must be aware of two separate IDs and keep them in sync, which is not always that easy. For example, what happens when a resource is created on the client-side (with a `_cid`) and the actual network request fails, never persisting the resource in the database? I don't think this case is even handled currently in Tailor (I might be wrong on this, so please do correct me if I am).

My suggestion would be to remove the `_cid` logic completely and just wait for the server to actually create the resource before showing it to the user (or in case of an error, not show it). To make the UX better while waiting, and material design specification might have more to say on this, the following should be done (as determined by some user studies, I think):
- Wait 250-300ms before showing the spinner
- If after that time the data still isn't ready, show the spinner
- If the spinner is shown, show it for at least X ms (I don't actually remember what X is)

Sorry again if I'm just missing something here that would invalidate this whole post; please do let me know if that's the case 😄

zcuric:
@2Pacalypse- My line of thinking is the same. 

> My suggestion would be to remove the _cid logic completely and just wait for the server to actually create the resource before showing it to the user (or in case of an error, not show it). 

I second this. For store unique keys we have `uuid` in case of updating.

zcuric:
I would attach two links that are relatable to this topic. 
- [Vuex best practices for complex objects
](https://forum.vuejs.org/t/vuex-best-practices-for-complex-objects/10143/46)
- [Normalizing State Shape - Redux approach](https://redux.js.org/recipes/structuring-reducers/normalizing-state-shape)

kronicker:
@2Pacalypse- Even though I see some benefits of the current approach, I think it is maybe it's starting to create tripping point for the maintainers (especially those not so involved) that might not be worth the gains, especially if we would implement some more _advanced_ UX handling like the one you mentioned so I'm seconding you on this.
Btw, could you link to the specs that you paraphrased?

2Pacalypse-:
I was actually quoting a friend of mine there who worked at the material design team, who in turn was paraphrasing an internal user study done by them, so I don't think there's a publicly available link for that study 😄

However, research in this area can be conflicting at times. For example, there's another method to show loading progress that we can use, called skeleton screens. Material design specification is also suggesting to use them in certain cases, eg. when loading a list of some kind. And here are two user studies comparing the three methods of showing the loading progress (blank page, spinner, skeleton screen) and coming to different conclusions:
- https://www.viget.com/articles/a-bone-to-pick-with-skeleton-screens
- https://uxdesign.cc/what-you-should-know-about-skeleton-screens-a820c45a571a

In any case, to not go completely off-topic by talking about UI/UX of the progress loaders, the main point that I'd like to make is that the problem of waiting and *time perception* that the `_cid` is trying to solve, is almost always handled at the UI/UX level, and not by changing the logic of your application.

vladimyr:
> the main point that I'd like to make is that the problem of waiting and _time perception_ that the `_cid` is trying to solve, is almost always handled at the UI/UX level, and not by changing the logic of your application.

Assuming acceptable waiting times and being ok with slowing down users during _batch crud sessions_. Also this eliminates possibility of working offline even temporarily due to connection loss.

zcuric:
> Assuming acceptable waiting times and being ok with slowing down users during _batch crud sessions_. Also this eliminates possibility of working offline even temporarily due to connection loss.

Heavy operations, like cloning or even linking a big tree, are expect to last longer and user should get feedback - progress bar, spinner etc.
Working offline doesn't work either way. If you try to use tailor without server running you'll have silent fail.

vladimyr:
> Heavy operations, like cloning or even linking a big tree, are expected to last longer and user should get feedback - progress bar, spinner etc.

I'm not talking about one heavy operation but sequence of quick crud actions that will get slowed down due to network lag compared to current state of affairs ☝️ 

> Working offline doesn't work either way. If you try to use tailor without server running you'll have silent fail.

True, I ignored the fact that we don't have client-server reconciliation mechanism in play currently...

PR #313, opened 2019-07-16
Use Treeview on Activity copy

ibebic:
Screen Shot 2019-07-26 at 12 04 32


ilijaveselica:
I would suggest removing left margin and reducing font size, it will be benefitial in cases when activities have 3 or more levels or activity titles are long

ibebic:
> I would suggest removing left margin and reducing font size, it will be benefitial in cases when activities have 3 or more levels or activity titles are long

👍 Thanks for the suggestion, haven't thought of that. I decreased margins a bit, but instead of decreasing the font I increased the modal width for an additional 100px. It doesn't look too large even on smaller screens and the additional space is (I think) more than enough.

underscope:
@ibebic Please update once #340 gets merged

underscope:
Please merge latest develop and similar to the new creation dialog try to scope stuff you need for copying within copy dialog / store.

PR #321, opened 2019-07-25
System settings access fix

franosincic:


client/router.js
```diff
@@ -113,6 +113,8 @@ let router = new Router({
 router.beforeEach((to, from, next) => {
   if (to.matched.some(it => it.meta.auth) && !store.state.auth.user) {
     next({ path: '/login', query: { redirect: to.fullPath } });
+  } else if (to.matched.some(it => it.meta.restricted) && !store.getters.isAdmin) {
+    next({ name: from.name });
```
vladimyr: Do we always have from? How about fallback to home if from isn't available for some obscure reason?

client/router.js
```diff
@@ -69,7 +69,7 @@ let router = new Router({
   }, {
     path: '/system-settings',
     component: SystemSettings,
-    meta: { auth: true },
+    meta: { auth: true, restricted: true },
```
vladimyr: :bulb: I would either call this `require/ensureAdmin` or refactor it to accept array of allowed roles.

PR #326, opened 2019-07-27
Upgrade ESLint configuration

vladimyr:
This PR upgrades ESLint (semi)standard and vue plugin to latest version; addresses ESLint reports 🏄


.eslintrc.js
```diff
@@ -47,11 +64,12 @@ module.exports = {
...
         'UNIQUE',
-        'BINDING',
+        'TWO_WAY_BINDING',
+        'OTHER_DIRECTIVES',
         'EVENTS',
-        'CONTENT',
         'GLOBAL',
-        'OTHER_ATTR'
+        'OTHER_ATTR',
+        'CONTENT'
```
underscope: Should we consider bumping CONTENT up ? I would expect v-html to be defined as one of the top ones.... 🤔

PR #328, opened 2019-07-29
Use Vuetify v-switch/v-checkbox on switch/checkbox meta

ibebic:
Screen Shot 2019-07-29 at 11 07 58
Screen Shot 2019-07-29 at 11 23 46


client/components/common/Meta/Checkbox.vue
```diff
@@ -1,24 +1,19 @@
...
-          v-model="value"
-          @change="$emit('update', meta.key, value)"
-          :id="meta.key"
-          :name="meta.key"
-          type="checkbox">
-      </label>
-      <p class="description">{{ meta.description }}</p>
-    </div>
+    <v-checkbox
+      v-model="value"
```
abasic: There is no need for a v-model here nor below. A combination of `input-value` prop and a `change` handler will be enough for our requirements.

client/components/common/Meta/Switch.vue
```diff
@@ -30,17 +24,11 @@ export default {
...
-$lever-checked: lighten($checked, 25%);
-$lever-unchecked: lighten($unchecked, 25%);
-
 .control {
   padding: 3px 8px;
 
-  &:hover {
-    background-color: #f5f5f5;
+  /deep/ label.v-label {
+    margin-bottom: 0;
```
abasic: I don't think this has any effect on the styling since I can not find a `label` in the `v-switch` component.

ibebic: `label` section of the checkbox/switch components is used to display the optional `description` parameter of the meta component. When added it looks something like this:
Screen Shot 2019-07-30 at 20 35 56
Without the added css, the label is not perfectly aligned with the checkbox/switch.

zcuric:
@abasic can we merge this?

vladimyr:
> @vladimyr can we merge this?

As far as I'm concerned this is one truly fine PR :+1:

rkusan00:
NOTE: This has been merged in #351

Issue #53, opened 2018-03-29
Add 'deleted_at' property to content repository when deleting it


ibebic:
## I'm submitting a...

- [ ] Bug report  
- [ ] Regression (a behavior that used to work and stopped working in a new release)
- [x] Feature request
- [ ] Documentation issue or request

## Context
If would be useful to be able to see via the published content when a content repository is deleted from the authoring platform. Please add a 'deleted_at' tag and publish info on deletion (but only if that content repository was already published)

PR #331, opened 2019-07-29
Add "detachedAt" property to deleted repositories

DorianSyruss:
This PR solves issue #53. Upon repo deletion, publishes the new version of catalog with "deprecatedAt" attribute added for the deleted repository. The publishing should only happen if the deleted repo was already published.


server/course/course.controller.js
```diff
@@ -54,6 +54,11 @@ function patch({ body, course, user }, res) {
 
 function remove({ course, user }, res) {
   return course.destroy({ context: { userId: user.id } })
+    .then(course => {
+      const deprecatedAt = course.dataValues.deleted_at;
```
DorianSyruss: If anyone knows why is only snake case "deleted_at" field populated and the change only get's reflected in dataValues (on returned instance meta deleted_at doesn't get populated, nor aliased), help would be appreciated. @vladimyr

server/shared/publishing/helpers.js
```diff
@@ -56,6 +56,18 @@ function updateRepositoryCatalog(repository, publishedAt) {
   });
 }
 
+function deprecateRepository(repository, deprecatedAt) {
```
DorianSyruss: I've separated concerns of repo deprecation and repo update with this method, even though we could probably refactor "updateRepositoryCatalog" method to deprecate repos under certain conditions but I'm not sure we should.

server/course/course.controller.js
```diff
@@ -54,6 +54,11 @@ function patch({ body, course, user }, res) {
...
+      if (!deprecatedAt) return;
+      publishingService.deprecateRepository(course.dataValues, deprecatedAt);
```
kronicker: Move this call to `afterDestroy` hook.

server/course/course.controller.js
DorianSyruss: This is resolved, by passing ```returning: true``` in options instance gets returned with exact data.

server/shared/publishing/helpers.js
```diff
@@ -41,11 +41,21 @@ function publishActivity(activity) {
   });
 }
 
-function updateRepositoryCatalog(repository, publishedAt) {
+function getRepositoryCatalog() {
   return storage.getFile('repository/index.json').then(buffer => {
-    let catalog = (buffer && JSON.parse(buffer.toString('utf8'))) || [];
+    return (buffer && JSON.parse(buffer.toString('utf8'))) || [];
```
zcuric: Can you make this return more readable? Split it, maybe.

server/script/detachDeletedRepos.js
```diff
@@ -0,0 +1,29 @@
+const { Course } = require('../shared/database');
+const each = require('lodash/each');
+const omit = require('lodash/omit');
+const find = require('lodash/find');
+const storage = require('../shared/storage');
+const { getRepositoryAttrs, getRepositoryCatalog } = require('../shared/publishing/helpers');
```
abasic: Multiple destructures should be on top.

underscope:
@DorianSyruss Please merge develop

server/course/course.model.js
```diff
@@ -67,6 +68,15 @@ class Course extends Model {
     });
   }
 
+  static hooks() {
+    return {
+      afterDestroy(course) {
+        return Course.findByPk(course.id, { paranoid: false })
```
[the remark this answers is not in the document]
DorianSyruss: Exactly, could have done instance.reload but It's the same I guess, and already saw this pattern on tailor.

PR #333, opened 2019-07-30
Updated poi config to support eslintignore file

abasic:
- removed old `poi` `eslint` package
 - `ignore` flag is set to true in `poi` config for `eslint-loader`

### DIFF
--- a/poi.config.js
+++ b/poi.config.js
@@ -101,6 +101,13 @@ module.exports = {
       .plugin('dotenv')
       .use(require.resolve('dotenv-webpack'));
 
+    config
+      .module
+      .rule('eslint')
+      .use('eslint-loader')
+        .loader('eslint-loader')
+        .tap(options => Object.assign(options, { ignore: true }));
+
     if (mode !== 'production') return;
     config
       .plugin('minimize')
--- a/package-lock.json
+++ b/package-lock.json
@@ -1510,9 +1510,9 @@
           "dev": true
         },
         "ajv": {
-          "version": "6.9.2",
-          "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.9.2.tgz",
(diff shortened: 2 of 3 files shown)


poi.config.js
```diff
@@ -101,6 +101,13 @@ module.exports = {
       .plugin('dotenv')
       .use(require.resolve('dotenv-webpack'));
 
+    config
+      .module
+      .rule('eslint')
+      .use('eslint-loader')
+        .loader('eslint-loader')
+        .tap(options => Object.assign(options, { ignore: true }));
```
vladimyr: If I understand code of `eslint-loader` correctly this gets passed down to `eslint`'s `CLIEngine`: https://github.com/webpack-contrib/eslint-loader/blob/2a72daa5b78e3ab0a1d9861210f21e7bb4440948/index.js#L22

That being said, docs on following page: https://eslint.org/docs/developer-guide/nodejs-api#cliengine clearly say that:
>ignore - False disables use of .eslintignore, ignorePath and ignorePattern (default: true). 

So what does this do really? 🤔

kronicker: In build process for production, for some reason, `ignore` would be resolved to `false`, thus _ignoring_ `ignorePath` and/or `ignorePattern`.

vladimyr: Hmm, this thing http://npmfs.com/package/@poi/plugin-eslint/12.0.0/index.js does not toggle it so what does?

kronicker: I know, that's what was baffling me as well.
After more than an hour of debugging, we couldn't figure out what was happening, we only came to the conclusion that setting `ignore: true` explicitly resolves the issue...

vladimyr: I assume you did run `webpack:inspect-config` in prod mode? No luck there?

kronicker: Yes, but the config seemed fine.

abasic:
@vladimyr @kronicker What should we do with this PR?

vladimyr:
> @vladimyr @kronicker What should we do with this PR?

My vote goes for closing it because:
a) we had only one complaint we never fully investigated
b) hopefully #350 will eliminate it :crossed_fingers:

PR #336, opened 2019-07-30
Use Vuetify v-date-picker on date meta

ibebic:


ibebic:
Closing for now, currently used DateTime picker is simpler than the separate Vuetify Date and Time pickers, just needs a few styling tweaks to be more in line with the current general design.

PR #338, opened 2019-07-31
Use Vuetify buttons on colorpicker meta

ibebic:
Screen Shot 2019-07-31 at 11 25 41

NOTE: This is just a temporary solution until upgrading to Vuetify v2 which has its own Color picker component.


zcuric:
@abasic can we merge this?

abasic:
Closing this since Tailor is being actively moved to Vuetify v2.

PR #339, opened 2019-07-31
Tweak FileUpload meta styling

ibebic:
~~NOTE: This is just a temporary styling tweak until upgrading to Vuetify v2 which has its own `file input` component.~~

EDIT: Used `v-file-input` component for upload:

ezgif-2-ee602445435c

~~TODO: Validation currently done through input `accept` attribute. Improved validation is perhaps needed.~~ - validation through `accept` attribute is enough for now considering the fact that `FILE` type validation currently consists only from accepted file extensions.


client/components/common/FileUpload.vue
```diff
@@ -17,12 +17,15 @@
         :class="[sm ? 'v-btn v-btn--small' : 'btn btn-material btn-sm upload-button']">
         {{ label }}
       </label>
-      <span
-        v-else
-        @click="downloadFile"
-        class="file-name">{{ fileName }}
-      </span>
-      <span v-if="fileKey" @click="deleteFile" class="mdi mdi-delete delete"></span>
+      <span v-else @click="downloadFile" class="file-name">{{ fileName }}</span>
```
abasic: Can you please try and make this into a flat Vuetify button and see what it looks like.

client/components/common/FileUpload.vue
```diff
@@ -17,12 +17,15 @@
...
+      <v-icon
```
abasic: This should be a `Veutify` `v-button` icon instead of just a `v-icon`.

client/components/common/FileUpload.vue
ibebic: I made the changes and it does look better:
Screen Shot 2019-07-31 at 13 28 10

kjuej:
@ibebic Please truncate the file name label.
Screen Shot 2019-10-16 at 11 20 10

underscope:
@ibebic When you get a chance please revisit remaining issues and migrate to v2

client/components/common/FileUpload.vue
```diff
@@ -52,6 +40,12 @@ export default {
   data() {
     return { uploading: false };
   },
+  computed: {
+    acceptedFileTypes() {
+      const { ext } = this.validate;
```
underscope: If ext is not set this will result with err

client/components/common/FileUpload.vue
```diff
@@ -1,46 +1,34 @@
 <template>
-  <div class="file-upload">
```
underscope: Please match new outlined style of other components
image

underscope:
- Tweaked component to match other meta
- Kept initial input btn as is (since it is reused within TE)
- Extracted common logic
- Input btn should be entirely decoupled with included styles (WIP)

PR #341, opened 2019-08-09
Fix multiselect placeholder overlay

vkaracic:
Before:
Screenshot from 2019-08-09 08-36-42

After:
Screenshot from 2019-08-09 08-36-28

I went through all the (multi)select elements that I could find and none of them seem to be broken after this change.


zcuric:
@abasic can you merge this?

PR #342, opened 2019-08-14
Jodit editor updates

borna223:


client/components/content-elements/tce-passage/edit/plugins/table-popups/index.js
```diff
@@ -0,0 +1,49 @@
...
+    afterInitHook.apply(this, arguments);
+    observeTables(this, Jodit);
+    addScrollHandler(this);
+  };
+};
+
+function observeTables(editor, Jodit) {
+  const $$ = get(Jodit, 'modules.Helpers.$$', () => []);
+  const tableProcessor = get(editor, '__plugins.table', {});
+  if (typeof tableProcessor.observe !== 'function') return;
```
abasic: Use `lodash/isFunction` for this.

client/components/content-elements/tce-passage/edit/plugins/table-popups/index.js
```diff
@@ -0,0 +1,49 @@
...
+  node.addEventListener('scroll', handler);
+  editor.events.on('beforeDestruct', () => {
+    node.removeEventListener('scroll', handler);
+  });
+}
+
+function getScrollHandler(editor, popup = {}) {
+  const store = editor.events.getStore(editor.events);
+  const [{ originalCallback } = {}] = store.get(EVENT_NAME, NAMESPACE);
+  if (typeof originalCallback !== 'function') return;
```
abasic: Use `lodash/isFunction` for this.

client/components/content-elements/tce-passage/edit/plugins/table-popups/index.js
```diff
@@ -0,0 +1,49 @@
...
+    observeTables(this, Jodit);
+    addScrollHandler(this);
+  };
+};
+
+function observeTables(editor, Jodit) {
+  const $$ = get(Jodit, 'modules.Helpers.$$', () => []);
+  const tableProcessor = get(editor, '__plugins.table', {});
+  if (typeof tableProcessor.observe !== 'function') return;
+  $$('table', editor.editor).forEach(table => {
```
abasic: Can the `find` method be used here? I assume the observe on the `tableProcessor` will only be called once since there should be only one property on the table object equal to `tableProcessor.__key`. So the code can just find the table it needs and if it exists `observe` it.

borna223: Every single table that was "given" to the editor needs to be observed so some sort of iteration is needed. `$$` helper (which finds all instances of HtmlTableElement) was only used to avoid any possible side effects because Jodit uses it internally.

PR #347, opened 2019-08-29
TCE-file

borna223:


client/components/content-elements/tce-file/edit/Toolbar.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+  <div class="tce-file-toolbar">
+    <form ref="upload" @submit.prevent>
```
abasic: Allow the user to set the `name` that appears as the text of the `button` instead of using the filename.

zcuric:
@abasic I merged latest develop here. Is this mergable?

PR #354, opened 2019-10-03
Use @extensionengine/eslint-config

zcuric:
### This PR
- [x] Installs` @extensionengine/eslint-config`
- [x] Lints


karma.conf.js
```diff
@@ -1,3 +1,6 @@
+
```
kronicker: Extra space

client/utils/revision.js
```diff
@@ -60,6 +60,7 @@ export function getFormatDescription(rev, activity) {
 export function getRevisionAcronym(rev) {
   switch (rev.entity) {
     case 'ACTIVITY':
+      // eslint-disable-next-line no-case-declarations
```
vladimyr: Correct me if I'm wrong but you could just wrap case into a block?
```js
case 'ACTIVITY': {
  const typeArray = rev.state.type.split('_', 2);
  return reduce(typeArray, (acc, val) => acc + val.charAt(0), '');
}
```

client/utils/revision.js
```diff
@@ -75,6 +76,7 @@ export function getRevisionColor(rev) {
   const DEFAULT_COLOR = '#808080';
   switch (rev.entity) {
     case 'ACTIVITY':
+      // eslint-disable-next-line no-case-declarations
```
vladimyr: Same as above :arrow_up:

zcuric: Fixed.

zcuric:
@vladimyr Issues addressed.

PR #358, opened 2019-10-05
Cleanup gitignore

vladimyr:
This PR removes obsolete patterns and alters existing ones to cover **all** standard pm2 & _rc_ configuration patterns. :tada:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,27 +11,24 @@ logs
 *.log
 npm-debug.log*
 
-# Test reports
-client/test/unit/coverage
-client/test/e2e/reports
-selenium-debug.log
-
 # Configuration
 .env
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -11,27 +11,24 @@ logs
...
 extensions/content-elements/*
 !extensions/content-elements/.gitkeep
 
 # Branding
-client/assets/img/*logo*
-!client/assets/img/default-logo*
-client/assets/img/*favicon*
-!client/assets/img/default-favicon.ico
+client/assets/img/*
+!client/assets/img/default-*
```
underscope: Should we take into account that we might add other img assets in the future 🤔

PR #360, opened 2019-10-10
Jodit tooltip control

borna223:
Screen Shot 2019-10-10 at 12 54 28
Screen Shot 2019-10-10 at 12 54 41

### DIFF
--- a/client/components/content-elements/tce-jodit-html/edit/controls/tooltip/index.js
+++ b/client/components/content-elements/tce-jodit-html/edit/controls/tooltip/index.js
@@ -15,8 +15,8 @@ export default {
       const { val } = Jodit.modules.Helpers;
       const form = editor.create.fromHTML(
         `<form class="jodit_form">
-          <textarea name="tooltip"></textarea>
-          <input name="text" type="text">
+          <textarea name="tooltip" placeholder="Tooltip"></textarea>
+          <input name="text" type="text" placeholder="Text">
           <div style="text-align: right">
             <button name="delete" type="button">Delete</button>
             <button name="submit" type="submit">Submit</button>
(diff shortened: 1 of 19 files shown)


client/components/content-elements/tce-jodit-html/edit/controls/tooltip/index.js
```diff
@@ -0,0 +1,63 @@
...
+          <input name="text" type="text" placeholder="Text">
+          <div style="text-align: right">
+            <button name="delete" type="button">Delete</button>
+            <button name="submit" type="submit">Submit</button>
+          </div>
+        </form>`
+      );
+
+      const args = [current, TOOLTIP_TAG_NAME, editor.editor];
+      current = (current && Jodit.modules.Dom.closest(...args)) || false;
```
abasic: The same check is performed here and below: if `current` is truthy. I think the two can be merged by doing the following changes:

- if `current` is not `null` here but `undefined` then we can default to `false` on line 13. If it is `null` then we can keep it as is since it is a falsy value.
- we remove this line and the line above.
- we move the `Jodit.modules.Dom.closest([current, TOOLTIP_TAG_NAME, editor.editor])` to the first line of the `if` block.

I believe these changes make the code more readable.

client/components/content-elements/tce-jodit-html/edit/controls/tooltip/index.js
```diff
@@ -0,0 +1,62 @@
+import Jodit from 'jodit';
+
+const TOOLTIP_CONTROL = 'tooltip';
+const TOOLTIP_TAG_NAME = 'tooltip-element';
+const TOOLTIP_ATTRIBUTE = 'data-tooltip';
+const TOOLTIP_CLASS = 'tce-jodit-tooltip';
+
+class TooltipElement extends HTMLElement {}
+customElements.define(TOOLTIP_TAG_NAME, TooltipElement);
```
kronicker: Why custom element? You're narrowing down browser support. Use `span` instead

kjuej:
@borna223 Suggestion: disable delete button until there's a tooltip to be deleted.
Screen Shot 2019-10-16 at 14 42 38

PR #366, opened 2019-10-23
Move away from `course` nomenclature

zcuric:
### This PR
- [x] Replaces all occurrences of `course` with `repository`
- [x] Renames files and folder from course.*.{js,vue} to repository
- [x] Add migrations to update table names and columns


underscope:
This would create too much mess / conflicts at the moment. We should approach it partially and focus on BE first. Lets discuss tomorrow 😉 🍻

PR #367, opened 2019-10-29
Introduced vuex-persist plugin

ttokic:
This PR includes:

- [x] Introduces [vuex-persist](https://www.npmjs.com/package/vuex-persist) plugin to Tailor

#### Motivation
The motivation for this PR is to replace the manual synchronization of the Vuex state with the local storage. This plugin watches the **state mutations** for the given **Vuex modules** and automatically updates the **local storage**. It basically keeps them synchronized.

This PR introduces ```store/plugins``` directory which should contain the collection of the Vuex plugins, it does not matter if they are custom or 3rd party.


client/store/plugins/vuex-persist.js
```diff
@@ -0,0 +1,12 @@
+import VuexPersistence from 'vuex-persist';
+
+const STORAGE_KEY = process.env.VUEX_PERSIST_STORAGE_KEY;
```
underscope: Doesn't feel like it needs rename / could be destructured. Consider renaming to something not vendor specific / simpler 💅

client/store/modules/auth/mutations.js
```diff
@@ -1,9 +1,9 @@
-export const login = (state, user) => {
-  state.user = user;
+export const login = (state, data) => {
```
underscope: This is a bit unreadable, specially because you are assigning to whole state. Consider destructuring payload `{ user, token }` and assign it explicitly. This will also avoid any unwanted assignments.

client/store/modules/auth/getters.js
```diff
@@ -2,3 +2,5 @@ import get from 'lodash/get';
 import { role } from 'shared';
 
 export const isAdmin = ({ user }) => get(user, 'role') === role.user.ADMIN;
+
+export const token = state => state.token;
```
underscope: Whats the point of this getter ?

client/store/modules/auth/actions.js
```diff
@@ -2,14 +2,14 @@ import api from '@/api/auth';
 
 export const login = ({ commit }, credentials) => {
   return api.login(credentials)
-    .then(user => commit('login', user));
+    .then(data => commit('login', data));
 };
 
-export const logout = () => {
+export const logout = ({ commit }) => {
```
underscope: See comment on api.logout method and cleanup this one accordingly...

client/store/modules/auth/actions.js
```diff
@@ -2,14 +2,14 @@ import api from '@/api/auth';
 
 export const login = ({ commit }, credentials) => {
   return api.login(credentials)
-    .then(user => commit('login', user));
+    .then(data => commit('login', data));
```
underscope: I would prefer destructured data (if we don't have naming for user + token) to see what is being passed as data. One other possible solution would be to pass token within user object.

client/store/modules/auth/getters.js
ttokic: I added that getter because I had to get the `token` and pass it as the argument to `SSEClient` constructor. Feel free to take a look [here](https://github.com/ExtensionEngine/tailor/pull/367/files#diff-24ace95a81a97a92f1e785a22e295893R16).
 
Another reason I needed that getter is because I had to add the same token to the Axios interceptor for the authorization.
References:
 - [Reference_1](https://github.com/ExtensionEngine/tailor/pull/367/files#diff-096103d8490b885f9d5037baea6aea80R58)
- [Reference_2](https://github.com/ExtensionEngine/tailor/pull/367/files#diff-de413e6307dae35c58e236129ed45f76R19)

underscope: Sorry for not clarifying it a bit better 😞, I was referring to the fact that we exposed it as a getter although we are not doing any processing. Can we just pick it up from state?

client/store/plugins/vuex-persist.js
```diff
@@ -0,0 +1,11 @@
+import VuexPersistence from 'vuex-persist';
+
+const MODULES_TO_PERSIST = ['auth'];
+
+export default new VuexPersistence({
+  // The key to store the state on in the storage provider.
+  key: process.env.PERSIST_STORAGE_KEY,
+  modules: MODULES_TO_PERSIST,
```
underscope: What did we gain with extracting this 🤔

poi.config.js
```diff
@@ -85,7 +85,8 @@ module.exports = {
   },
   envs: {
     API_PATH: '/api/v1/',
-    AUTH_JWT_SCHEME
+    AUTH_JWT_SCHEME,
+    PERSIST_STORAGE_KEY: 'TAILOR_USER'
```
underscope: Consider updating key name / value since:
- Other modules can be added later on (with new approach) so I am not sure if TAILOR_USER name makes sense (rather go with something more generic like) APP_STATE / APP_STORE ?
- For me PERSIST_STORAGE_KEY isn't descriptive enough (what is it actually being persisted) so I would rename it to something like VUEX_STORAGE_KEY

client/store/plugins/vuex-persist.js
ttokic: I have extracted it because other modules can be added later on, so the list might end up large.

poi.config.js
ttokic: Do you mind if I add a short comment above `VUEX_STORAGE_KEY: APP_STORE` describing the purpose of this setting. 

It would be something like: 
> The key to store the Vuex state in the storage

client/store/plugins/vuex-persist.js
underscope: 👍 Lets do it once we get there 😉

poi.config.js
underscope: Sure 👍

underscope:
@ttokic Don't forget to merge latest develop as-well 😉

client/store/plugins/vuex-persist.js
ttokic: Sure 👍

client/api/request.js
```diff
@@ -42,8 +19,13 @@ Object.defineProperty(client, 'base', {
   }
 });
 
+client.setStorageInterface = function ({ getToken, clearAuthData }) {
+  this._getAuthToken = getToken;
+  this._clearAuthData = clearAuthData;
+};
```
vladimyr: Previously `Auth` class reported back to `vuex` and now we need bidirectional communication between `vuex/persist` and `request` instance. Token flows from `vuex` to `request` and token invalidation requests go opposite way. Ofc that means we can't use events anymore and next step would be implementing request-reply pattern by using specially crafted pairs of request and reply events but that is too much work for this simple use case so we ended up with alternative solution that is basically passing interfaces around. That is fine but maybe we can make it even better - by better I mean more decoupled - by doing something along these lines.
From `request`'s perspective we can use local `request.token` variable and just get it and set it to `null` when needed (invalidation due to server response). From `vuex`'s perspective we can _decorate_ `request` with `token` property:
```js
Object.defineProperty(request, 'token', {
  get: () => store.state.auth.token,
  set: value => !value && store.dispatch('logout')
});
```
As far as `request` is concerned it is changing and reading local variable while vuex intercepts that calls and acts accordingly.

client/api/request.js
```diff
@@ -42,8 +19,13 @@ Object.defineProperty(client, 'base', {
...
+
 client.interceptors.request.use(config => {
-  const { token } = client.auth;
+  const token = client._getAuthToken();
```
vladimyr: Continuing on my previous comment here we would have:
```js
const { token } = client;
```

client/api/request.js
```diff
@@ -54,7 +36,7 @@ client.interceptors.request.use(config => {
 
 client.interceptors.response.use(res => res, err => {
   if (err.response && [FORBIDDEN, UNAUTHORIZED].includes(err.response.status)) {
-    return client.auth.emit('error', err);
+    client._clearAuthData();
```
vladimyr: And here we would just invalidate it by conventionally setting it to `null`:
```js
client.token = null;
```

client/main.js
```diff
@@ -40,6 +41,11 @@ Vue.use(Timeago, {
   }
 });
 
+request.setStorageInterface({
+  getToken: () => store.state.auth.token,
+  clearAuthData: () => store.dispatch('logout')
+});
```
vladimyr: This eventually gets replaced with:
```js
Object.defineProperty(request, 'token', {
  get: () => store.state.auth.token,
  set: value => !value && store.dispatch('logout')
});
```

client/SSEClient.js
```diff
@@ -10,14 +10,13 @@ function listenerFactory(method) {
 }
 
 class SSEClient {
-  constructor(url) {
+  constructor(url, token) {
```
vladimyr: Can we import `request` here and get token from it?

client/store/plugins/index.js
```diff
@@ -0,0 +1,4 @@
+import vuexPersist from './vuex-persist';
+
+// Collect and export all Vuex plugins
+export default [vuexPersist];
```
vladimyr: Frankly I believe it is obvious so no need for explanatory comment 🙃

client/store/plugins/vuex-persist.js
```diff
@@ -0,0 +1,9 @@
+import VuexPersistence from 'vuex-persist';
+
+export default new VuexPersistence({
+  // The key to store the state in the storage provider.
+  key: process.env.VUEX_STORAGE_KEY,
+  modules: ['auth'],
+  // Use localStorage
+  storage: window.localStorage
```
vladimyr: Same as above because likely library itself has some sort of in-code docs (typings, jsdoc) and editor will show them as needed. :bowtie:

package.json
```diff
@@ -170,7 +171,7 @@
     "sass-loader": "^7.3.1",
     "sequelize-cli": "^5.5.1",
     "strip-json-comments": "^3.0.1",
-    "stylelint": "^11.1.1",
+    "stylelint": "^12.0.0",
```
vladimyr: Let it be 😁 
If we were really pedantic it should be done as separate PR but let's leave it now. 😉

poi.config.js
```diff
@@ -85,7 +85,9 @@ module.exports = {
   },
   envs: {
     API_PATH: '/api/v1/',
-    AUTH_JWT_SCHEME
+    AUTH_JWT_SCHEME,
+    // The key to store the Vuex state in the storage
+    VUEX_STORAGE_KEY: 'APP_STORE'
```
vladimyr: My vote goes for `TAILOR_STATE` or even better `TAILOR_APP_STATE` because:
1. you really want to prefix it with app name cause otherwise it will potentially clash with other locally developed apps running on same port
2. seing state suffix indicates that it came from vuex

poi.config.js
vladimyr: Quoting myself https://github.com/ExtensionEngine/tailor/pull/367#discussion_r353490268:
>My vote goes for TAILOR_STATE or even better TAILOR_APP_STATE because:
>
>1. you really want to prefix it with app name cause otherwise it will potentially clash with other locally developed apps running on same port
>2. seing state suffix indicates that it came from vuex

poi.config.js
```diff
@@ -85,7 +85,9 @@ module.exports = {
   },
   envs: {
     API_PATH: '/api/v1/',
-    AUTH_JWT_SCHEME
+    AUTH_JWT_SCHEME,
+    // The key to store the Vuex state in the storage
+    VUEX_STORAGE_KEY: 'TAILOR_APP_STATE'
```
vladimyr: Personally I would go without that comment because now it is pretty obvious what is going on here 🙃

ttokic: Resolved

PR #376, opened 2019-11-14
Cyclic dependency

borna223:
It is not possible to publish custom content containers because of this. Just a heads up until I find some time to fix it (or someone else does it before me :smiley_cat:).

### DIFF
--- a/server/repository/repository.controller.js
+++ b/server/repository/repository.controller.js
@@ -53,9 +53,10 @@ function patch({ user, repository, body }, res) {
     .then(repository => res.json({ data: repository }));
 }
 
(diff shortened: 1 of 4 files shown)


server/repository/repository.controller.js
```diff
@@ -54,7 +54,7 @@ function patch({ user, repository, body }, res) {
 }
 
 function remove({ user, repository }, res) {
-  return repository.destroy({ context: { userId: user.id } })
+  return repository.destroy({ context: { userId: user.id, publishingService } })
```
underscope: No need to complicate this. Just put publishing here (remove it from hook)

PR #397, opened 2019-11-19
Make sure that transaction is used when cloning

zcuric:
### This PR
- [x] Makes sure that transaction is used when cloning


underscope:
Thanks 😉

PR #412, opened 2019-12-05
Update example config

vladimyr:
This PR swaps previous example config with new one aligning with configuration loading priority established by earlier changes. Additionally readme has been updated together with few spelling & grammar error fixes.


tailor.config.js.example
```diff
@@ -0,0 +1,41 @@
+'use strict';
```
underscope: It would be awesome if this would be a working / real example 🤡

tailor.config.js.example
```diff
@@ -0,0 +1,41 @@
...
+    relationships: [{
+      type: 'prerequisites',
+      label: 'Prerequisites',
+      multiple: true,
+      searchable: true,
+      allowEmpty: true,
+      placeholder: 'Click to select',
+      allowCircularLinks: false,
+      allowInsideLineage: false,
+      allowedTypes: ['ACTIVITY_TYPE_1', 'ACTIVITY_TYPE_2']
```
underscope: Swap with GOAL 🐎

tailor.config.js.example
vladimyr: What am I missing? 😕 I tried it locally and in fact it does work... 🤔

PR #414, opened 2019-12-09
Update readme using script

vladimyr:
This PR introduces script for updating readme file. It reads `package.json` `engines` field and updates engines list inside **Prerequisites** section. :tada:


zcuric:
So, you made a script so you don't have to copy/paste 3 lines? So `vladimyr`. 😄 
LGMT 👍

vladimyr:
> So, you made a script so you don't have to copy/paste 3 lines? So `vladimyr`. 😄

As any reasonable dev would do 😉

PR #419, opened 2019-12-19
Allow custom container summary

kronicker:
This PR adds the option to export `buildSummary` method from custom containers `util.js` to allow the custom container to build a custom container summary in the published content outline.
If no method is supplied generic summary fallback is used.

Example use-case: Custom container has sub-containers that need to be surfaced to container summary in the outline.

### DIFF
--- a/server/shared/publishing/helpers.js
+++ b/server/shared/publishing/helpers.js
@@ -250,11 +250,18 @@ function getRepositoryAttrs(repository) {
 }
 
 function attachContentSummary(obj, { containers, assessments }) {
-  obj.contentContainers = map(containers, it => ({
-    ...pick(it, ['id', 'uid', 'type', 'publishedAs']),
-    elementCount: get(it.elements, 'length', 0)
-  }));
+  obj.contentContainers = map(containers, getContainerSummary);
   obj.assessments = map(assessments, it => pick(it, ['id', 'uid']));
+}
+
+function getContainerSummary(container) {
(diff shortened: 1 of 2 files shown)


server/shared/publishing/helpers.js
```diff
@@ -250,13 +250,20 @@ function getRepositoryAttrs(repository) {
...
 
+function getContainerSummary(container) {
+  const customBuilder = containerRegistry.getSummaryBuilder(container.type);
+  return customBuilder
+    ? customBuilder(container)
+    : {
+      ...pick(container, ['id', 'uid', 'type', 'publishedAs']),
+      elementCount: get(container.elements, 'length', 0)
+    };
+}
```
abasic: This looks kind of messy to me. I propose that one of these two things is done here:

1. The function early returns if the `customBuilder` exists so the highlighted code would look like this:
```js
if (customBuilder) return customBuilder(container);
return {
  ...pick(container, ['id', 'uid', 'type', 'publishedAs']),
  elementCount: get(container.elements, 'length', 0)
}
```
2. The default functionality is extracted to a `defaultSummaryBuilder` (or named something like this method) method so the highlighted code would look like this:
```js
  return customBuilder ? customBuilder(container) : defaultSummaryBuilder(container);
}

function defaultSummaryBuilder({ id, uid, type, publishedAs, elements = [] }) {
  return { id, uid, type, publishedAs, elementCount: elements.length };
}
```
Maybe the lines go over 80 characters in option number two so it might need to be split into additional lines. 
My personal preference is the second option.

PR #423, opened 2019-12-20
Updated README.md with an extensions block

abasic:
- it describes what custom elements and containers are
- it links to templates which can be used to create them

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,7 +203,7 @@ Example:
 
 ## EXTENSIONS
 
-Tailor support creation of custom content elements and custom containers. Content
+Tailor supports creation of custom content elements and custom containers. Content
 elements are pieces of learning that a student would consume (text, image, etc.)
 while containers hold those content elements or other containers. These extensions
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -200,3 +200,14 @@ A string template that will be interpolated on the client using two route
 params, `repositoryId` and `activityId`, into a preview URL for each activity.
 Example:
 `https://my.url.com/#/repository/{repositoryId}/activity/{activityId}/preview`
+
+## EXTENSIONS
+
+Tailor support creation of custom content elements and custom containers. Content
```
underscope: Tailor supports ?

README.md
```diff
@@ -200,3 +200,14 @@ A string template that will be interpolated on the client using two route
...
+elements are pieces of learning that a student would consume (text, image, etc.)
```
underscope: I prefer avoiding learning terminology because that might not be the only ecosystem to use it.

PR #437, opened 2020-02-03
Fix the author of the comment also gets mail

### LINKED ISSUE #407: User gets an email notification for his own comment
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request

## Context
When writing a comment in a repository where the logged in user is an admin or author, that user gets an email notification notifying him of the comment he just wrote.

BaskovicP:
I want to resolve #407 issue: User gets an email notification for his own comment
Before making a list of recipients we filter the author of the coment. So the person who made the comment does not recieve a email but all the others in the repository do.

Please check it out: @kjuej

### DIFF
--- a/server/comment/hooks.js
+++ b/server/comment/hooks.js
@@ -2,7 +2,7 @@
 
 const { broadcast, events } = require('./channel');
 const mail = require('../shared/mail');
-const map = require('lodash/map');
+const reduce = require('lodash/reduce');
 const pick = require('lodash/pick');
 
 exports.add = (Comment, Hooks, db) => {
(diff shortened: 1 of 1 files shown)


server/comment/hooks.js
```diff
@@ -43,6 +43,9 @@ async function sendEmailNotification(comment, db) {
     author: author.profile,
     ...pick(comment, ['id', 'content', 'createdAt'])
   };
-  const recipients = map(repository.repositoryUsers, it => it.user.email);
+  const recipients = reduce(repository.repositoryUsers, (acc, it) => {
```
underscope: Although this is faster consider using `lodash/without` to make it a bit more readable and compact:
```javascript
const collaborators = map(repository.repositoryUsers, 'user.email');
const recipients = without(collaborators, author.email);
```
Also what happens if there are none 😉

BaskovicP:
Thanks @underscope ,it is much more readable like this. As _without_ always returns a new array even if we feed it something strange like undefined the array will be empty so we actually don't need to send any emails. (thats why the check for length)

kjuej:
@BaskovicP @underscope Seems to be working as expected.

server/comment/hooks.js
```diff
@@ -3,6 +3,7 @@
 const { broadcast, events } = require('./channel');
 const mail = require('../shared/mail');
 const map = require('lodash/map');
+const without = require('lodash/without');
```
underscope: When you get a chance please order it 😉

zcuric: 😉 👍

PR #439, opened 2020-02-06
Remove existing comments from fetch

### LINKED ISSUE #408: The SSE for user comments on repositories may show duplicate comments
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request

## Steps to Reproduce

1. Create a repository
2. Create a couple of child components within that repository
3. Login with 2 different users (admins) that have access to that repository
4. With user 1 focus on child component 1, and with user 2 focus on child component 2
5. With user 1 add a comment in the side meta on component 1
6. With user 2 click on component 1 and notice that comment is duplicated

Screen Shot 2019-12-04 at 09 28 16

BaskovicP:
I wish to resolve #408 

Bug flow and causes: 
1. SSE event is triggered -> all the suscribers have the new comment in state
2. The user focuses and clicks on the competency -> fetch is triggered
3. Fetch gets the data from the endpoint that includes the comment already put in the state by the SSE
4. Fetch inserts the fetched comments in the state 
5. The comment is then duplicated

Fix:
On the step 3. we scan for the comments that are already in the state and we filter them from the fetch before its triggered. That way we don't put the comments that are already in the state.

@kjuej @underscope Please check if the solution is satisfactionary and where the improvements are to be made. Thx.

### DIFF
--- a/client/store/helpers/resource.js
+++ b/client/store/helpers/resource.js
@@ -44,8 +44,10 @@ export default class Resource {
    */
   processEntries(items) {
     return reduce(items, (acc, it) => {
-      this.setCid(it);
-      acc[it._cid] = it;
+      if (!this.getCid(it.id)) {
+        this.setCid(it);
+        acc[it._cid] = it;
+      }
       return acc;
     }, {});
   }
--- a/client/store/helpers/resource.js
+++ b/client/store/helpers/resource.js
@@ -44,10 +44,8 @@ export default class Resource {
    */
   processEntries(items) {
     return reduce(items, (acc, it) => {
-      if (!this.getCid(it.id)) {
-        this.setCid(it);
-        acc[it._cid] = it;
-      }
+      this.setCid(it);
+      acc[it._cid] = it;
       return acc;
     }, {});
   }
--- a/client/store/modules/comments/actions.js
+++ b/client/store/modules/comments/actions.js
@@ -1,14 +1,21 @@
 import generateActions from '../../helpers/actions';
+import map from 'lodash/map';
+import pickBy from 'lodash/pickBy';
 import SSEClient from '../../../SSEClient';
 
 const { api, get, save, setEndpoint, update } = generateActions();
 let SSE_CLIENT;
 
+const filterExistingComment = (existingComments, fetchedComments) => {
+  const existingId = map(existingComments, 'id');
+  return pickBy(fetchedComments, it => !existingId.includes(it.id));
+};
+
 const fetch = ({ state, commit }, { id, repositoryId }) => {
   const action = state.repositoryId === repositoryId ? 'fetch' : 'reset';
   if (action === 'reset') commit('setRepository', repositoryId);
   return api.fetch({ activityId: id }).then(items => {
-    commit(action, items);
+    commit(action, filterExistingComment(state.items, items));
     commit('commentsFetched', id);
   });
 };
--- a/client/store/modules/comments/actions.js
+++ b/client/store/modules/comments/actions.js
@@ -1,21 +1,20 @@
+import find from 'lodash/find';
 import generateActions from '../../helpers/actions';
-import map from 'lodash/map';
-import pickBy from 'lodash/pickBy';
+import omitBy from 'lodash/omitBy';
 import SSEClient from '../../../SSEClient';
 
 const { api, get, save, setEndpoint, update } = generateActions();
 let SSE_CLIENT;
 
-const filterExistingComment = (existingComments, fetchedComments) => {
-  const existingId = map(existingComments, 'id');
-  return pickBy(fetchedComments, it => !existingId.includes(it.id));
+const omitExisting = (fetched, existing) => {
+  return omitBy(fetched, it => find(existing, { id: it.id }));
 };
 
 const fetch = ({ state, commit }, { id, repositoryId }) => {
   const action = state.repositoryId === repositoryId ? 'fetch' : 'reset';
   if (action === 'reset') commit('setRepository', repositoryId);
   return api.fetch({ activityId: id }).then(items => {
-    commit(action, filterExistingComment(state.items, items));
+    commit(action, omitExisting(items, state.items));
     commit('commentsFetched', id);
   });
 };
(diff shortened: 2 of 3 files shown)


client/store/modules/comments/actions.js
```diff
@@ -1,14 +1,21 @@
...
+import pickBy from 'lodash/pickBy';
 import SSEClient from '../../../SSEClient';
 
 const { api, get, save, setEndpoint, update } = generateActions();
 let SSE_CLIENT;
 
+const filterExistingComment = (existingComments, fetchedComments) => {
+  const existingId = map(existingComments, 'id');
+  return pickBy(fetchedComments, it => !existingId.includes(it.id));
+};
```
zcuric: I suggest doing this with `lodash/find` instead of `lodash/map`. 

```js
const filterExistingComment = (existingComments, fetchedComments) => {
  return pickBy(fetchedComments, it => !find(existingComments, { id: it.id }));
};
```

client/store/modules/comments/actions.js
```diff
@@ -1,14 +1,21 @@
...
+
 const fetch = ({ state, commit }, { id, repositoryId }) => {
   const action = state.repositoryId === repositoryId ? 'fetch' : 'reset';
   if (action === 'reset') commit('setRepository', repositoryId);
   return api.fetch({ activityId: id }).then(items => {
-    commit(action, items);
+    commit(action, filterExistingComment(state.items, items));
```
zcuric: Following the uppper comment, this would be then 
```js
 commit(action, omitExisting(items, state.items));
```

BaskovicP:
Thanks @zcuric the refactor was done by your suggestion and pushed. Thanks for the tip in the refactor.

kjuej:
@BaskovicP @underscope The issue no longer occurs on this branch.

client/store/modules/comments/actions.js
```diff
@@ -1,14 +1,20 @@
...
+const omitExisting = (fetched, existing) => {
+  return omitBy(fetched, it => find(existing, { id: it.id }));
+};
+
 const fetch = ({ state, commit }, { id, repositoryId }) => {
   const action = state.repositoryId === repositoryId ? 'fetch' : 'reset';
   if (action === 'reset') commit('setRepository', repositoryId);
   return api.fetch({ activityId: id }).then(items => {
-    commit(action, items);
+    commit(action, omitExisting(items, state.items));
```
underscope: You can consider going step further and use differenceWith so you can inline it 😉 

What concerns me is that resource wrapper should set proper _cid for retrieved record if the record has been already mapped/retrieved (see resource within store helpers).  If this was properly set then it wouldn't be duplicated since items are stored within object keyed by _cid. So the answer should be that upon sseAdd (need to manage it within actions since you have api exposed there) you should use setCid function to create mapping 😉

BaskovicP: Thanks for the tip. There was indeed a deeper issue in the resource file. In the processEntries function we would fill the acc with all the items, even those whose cuid is already present (it would just get overwritten which in itself is not a problem the problem arises when you fill the acc with everything). Now we have a check there that prevents inserting into the accumulator. Please check @underscope

client/store/modules/comments/actions.js
```diff
@@ -17,7 +17,10 @@ const subscribe = ({ state, commit }) => {
   if (SSE_CLIENT) SSE_CLIENT.disconnect();
 
   SSE_CLIENT = new SSEClient(`/api/v1${state.$apiUrl}/subscribe`);
-  SSE_CLIENT.subscribe('comment_create', item => commit('sseAdd', item));
+  SSE_CLIENT.subscribe('comment_create', item => {
+    commit('sseAdd', item);
+    api.setCid(item);
```
underscope: This should be prior 😉

client/store/helpers/resource.js
```diff
@@ -44,8 +44,10 @@ export default class Resource {
    */
   processEntries(items) {
     return reduce(items, (acc, it) => {
-      this.setCid(it);
-      acc[it._cid] = it;
+      if (!this.getCid(it.id)) {
```
underscope: This is being called only after fetch, when we know that cid needs to be set. `setCid` will attach mapping if it exists. Am I missing something / Why if ?

client/store/modules/comments/actions.js
BaskovicP: This will be fixed the next commit.

client/store/helpers/resource.js
BaskovicP: For this one:


```vue
// helpers/resource.js
  fetch(params) {
    return this.get('', params)
      .then(({ data: { data } }) => this.processEntries(data));
  }

get(path, params = {}) {
    return axios.get(this.url(path), { params });
  }
```
This part fetches all the comments. It doesn't care if there is something in the state already. 
So it loads comment1,comment2,comment3 even if the comment1 was added in the store and state by the SSE. Then we have the duplication of the comment1. We could also remove this.setCid(it) from the if. But what is the purpose of setting a _cid that is already mapped?

client/store/modules/comments/mutations.js
```diff
@@ -16,7 +15,6 @@ const commentsFetched = (state, activityId) => {
 const sseAdd = (state, comment) => {
   const { id } = comment;
   if (find(state.items, { id })) return;
-  comment._cid = cuid();
```
BaskovicP: Here was actually the main problem where we set the cuid without mapping it.
So this happens:

```js
const sseAdd = (state, comment) => {
  const { id } = comment;
  if (find(state.items, { id })) return;
  comment._cid = cuid();
  Vue.set(state.items, comment._cid, comment);
};
```
1. SSE event is triggered and the item has a new CUID which is not mapped just set
2. User selects the repository and **fetch** is triggered which activates **setCuid()**
 ```js
  getCid(id) {
    return this.mappings[id];
  }
setCid(model) {
    model._cid = this.getCid(model.id) || cuid();
    if (model.id) this.map(model._cid, model.id);
  }

```
3. setCuid()  is now searching for the cuid in mapping and as it doesn't find it so it sets it to a new cuid.
4. Now when we are filling the state and expecting that the same comments (with the same cuid) are overwritten this doesn't happen and we have comment duplication. 

``

PR #440, opened 2020-02-07
Refactor store structure

underscope:
- Scope activities, content elements, comments and revisions under repository
- Improve naming
- Cleanup


client/store/modules/repository/activities/actions.js
```diff
@@ -1,5 +1,5 @@
 import calculatePosition from 'utils/calculatePosition';
-import generateActions from '../../helpers/actions';
+import generateActions from '../../../helpers/actions';
```
zcuric: Maybe aliasing `helpers`?

underscope: I'll definitely do something about it 😉

PR #445, opened 2020-02-15
[CEP-152] TES Relationships 

zcuric:
### This PR
- [x] Add relationships between Teaching elements
- [x] Implements necessary UI 

### 🎫 JIRA 
https://extensionengine.atlassian.net/browse/CEP-152

### 🔍 QA Notes
- Add `relationships` to `tesMeta` in `.tailorrc.js`, for example: 
```js
  "tesMeta": [{
    "type": "IMAGE",
    relationships: [{
      type: 'relatedImage',
      label: 'Related Image',
      placeholder: 'Select related image',
      multiple: false,
      allowedTypes: ['IMAGE']
    }, {
      type: 'relatedDocuments',
      label: 'Related Document',
      placeholder: 'Select related document',
      multiple: true,
      allowedTypes: ['PDF', 'VIDEO']
    }]
  }]
```
- in editor on content elements with type image and try to add some relationships with other elements of allowed types, as set in config.


kjuej:
@zcuric Why not show only allowed elements in relation selection menu?

zcuric:
@kjuej Because of context. We need to be sure at which place we are adding the relationship.

kjuej:
@zcuric Should elements then be displayed in the same order as they're displayed in the container? Currently they're ordered by creation date (Meaning if you place an element between other elements in the container, it will still be shown last in the relationship selection menu).

kjuej:
@zcuric The functionality seems to be working as expected.

kjuej:
@zcuric Can they be sorted by content containers as well as position?

zcuric:
@kjuej probably, I'll have to check. @kronicker is that a necessity?

zcuric:
@kjuej probably, I'll have to check. @kronicker is that a necessity?

zcuric:
@kjuej They are now sorted by content container position and own position.

client/utils/activity.js
```diff
@@ -38,3 +38,13 @@ export function getAncestors(activities, activity) {
 export function isSameLevel(activityX, activityY) {
   return getLevel(activityX.type).level === getLevel(activityY.type).level;
 }
+
+export function toTreeFormat(activities, targetLevels, parentId = null, level = 1) {
```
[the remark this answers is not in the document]
zcuric: Uf, I messed up 🤡 . I moved the `toTreeFormat` that I changed for needs of the PR , instead of the original. Will address that.

zcuric:
@underscope 

> Docs 🤔 ? (Add to readme and wiki; reach Luka)

Good point. I will.

config/shared/activities.js
```diff
@@ -65,8 +65,8 @@ function getLevel(type) {
 }
 
 function getTesMeta(schemaId, type) {
```
zcuric:
```suggestion
function getElementMeta(schemaId, type) {
```

underscope: 👍 Just getting started on this one 😉

PR #447, opened 2020-02-18
Placed activity assessments inside of a container

abasic:


client/components/content-containers/tcc-assessment-block/edit/index.vue
```diff
@@ -15,60 +15,76 @@
...
-
-const appChannel = EventBus.channel('app');
+import pickBy from 'lodash/pickBy';
 
 export default {
-  name: 'assessments',
+  name: 'assessment-block',
+  props: {
+    container: { type: Object, required: true },
+    tes: { type: Object, required: true }
```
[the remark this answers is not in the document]
abasic: It can't be `contentElements` because in the `ContentContainers/index.vue` file the prop is still named [tes](https://github.com/ExtensionEngine/tailor/blob/develop/client/components/editor/structure/ContentContainers/index.vue#L26). I could change it but I think all of that should be done in a separate PR. If you want it to be a part of this one then I will change it.

client/components/content-containers/tcc-assessment-block/edit/index.vue
```diff
@@ -15,60 +15,76 @@
...
+    savedAssessments() {
+      return filter(this.tes, { activityId: this.container.id });
+    },
+    assessments() {
+      const { savedAssessments: saved, unsavedAssessments: unsaved } = this;
+      return [...saved, ...Object.values(unsaved)];
+    },
+    nextPosition() {
+      const lastItem = last(this.assessments);
+      return lastItem ? lastItem.position + 1 : 1;
```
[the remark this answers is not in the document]
abasic: It was not.

client/components/content-containers/tcc-assessment-block/edit/index.vue
```diff
@@ -15,60 +15,76 @@
...
     saveAssessment(assessment) {
       // TODO: Figure out why save is broken (for update)
-      return assessment.id ? this.update(assessment) : this.save(assessment);
+      const event = assessment.id ? 'updateElement' : 'saveElement';
+      return this.$emit(event, assessment);
+    },
+    clearUnsavedAssessments(assessments) {
+      const ids = assessments.map(it => it._cid);
+      const cond = it => !ids.includes(it._cid);
+      this.unsavedAssessments = pickBy(this.unsavedAssessments, cond);
```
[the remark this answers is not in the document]
abasic: `differenceBy` is only for Arrays.

client/components/content-containers/tcc-assessment-pool/edit/index.vue
```diff
@@ -0,0 +1,140 @@
...
+  watch: {
+    savedAssessments: 'clearUnsavedAssessments'
+  },
+  components: { AddElement, AssessmentItem }
+};
+</script>
+
+<style lang="scss" scoped>
+.assessments {
+  margin: 70px 0 250px;
```
underscope: Change px to rems

client/components/editor/structure/Assessments.vue
```diff
@@ -32,7 +32,7 @@
 <script>
 import { mapActions, mapGetters, mapMutations } from 'vuex';
 import AddElement from 'tce-core/AddElement';
-import AssessmentItem from './AssessmentItem';
+import AssessmentItem from 'tce-core/AssessmentItem';
```
underscope: Shouldn't this file be deleted entirely 🤔 ?

client/components/content-containers/tcc-assessment-pool/edit/index.vue
```diff
@@ -0,0 +1,140 @@
...
+import last from 'lodash/last';
+import map from 'lodash/map';
+import pickBy from 'lodash/pickBy';
+import sortBy from 'lodash/sortBy';
+
+export default {
+  name: 'assessment-block',
+  props: {
+    container: { type: Object, required: true },
+    tes: { type: Object, required: true }
```
underscope: We should update naming to contentElements (and handle legacy)

client/components/content-containers/tcc-assessment-pool/edit/index.vue
```diff
@@ -0,0 +1,140 @@
...
+  data() {
+    return {
+      unsavedAssessments: {},
+      selected: [],
+      allSelected: false
+    };
+  },
+  computed: {
+    savedAssessments() {
+      return filter(this.tes, { activityId: this.container.id });
```
underscope: Why can component get elements which are not part of this container? 🤔

client/components/content-containers/tcc-assessment-pool/edit/index.vue
```diff
@@ -0,0 +1,140 @@
...
+  },
+  methods: {
+    addAssessment(assessment) {
+      const cid = cuid();
+      const data = { ...assessment, _cid: cid };
+      this.unsavedAssessments = { ...this.unsavedAssessments, [cid]: data };
+      this.selected.push(data._cid);
+    },
+    saveAssessment(assessment) {
+      // TODO: Figure out why save is broken (for update)
```
underscope: Old comment / Might not be the case anymore.

client/components/content-containers/tcc-assessment-pool/edit/index.vue
abasic: Because that is how all of the containers currently work, including the [default](https://github.com/ExtensionEngine/tailor/blob/develop/client/components/editor/structure/ContentContainers/Container.vue#L56) ones. We fetch all the content elements and then we pass all of them to each container and then each one of them filters out the ones that are theirs.

client/components/editor/structure/Assessments.vue
abasic: If you are talking about the `Assessments` file then yes. I missed that. If you are talking about the `AssessmentItem`, it is used by both the Exam custom container and this container. And I am not sure if we want to duplicate that. The Exam custom container will probably never be used again so maybe we can remove it as well at some point and then remove `AssessmentItem` from `tce-core`.

underscope: The first one 😉

client/components/content-containers/tcc-assessment-pool/edit/index.vue
underscope: Can we improve it now? / Is there any reasoning behind that 🤔 ?

client/store/modules/repository/actions.js
```diff
@@ -7,7 +7,7 @@ export const initialize = ({ commit, dispatch }, id) => {
   const getRoute = entity => `repositories/${id}/${entity}`;
   const modules = {
     activities: 'activities',
-    tes: 'content-elements',
+    elements: 'content-elements',
```
underscope: We started to use content elements terminology because elements seemed a bit too generic (So I am not sure what to think about this). `contentElements` in some context seem a bit much and `elements` in some other seem a bit too generic.

client/store/modules/editor/getters.js
```diff
@@ -21,8 +21,8 @@ export const contentContainers = (_state, getters, { repository }) => {
 };
 
 export const assessments = (_state, getters, rootState) => {
-  const { items: tes } = rootState.repository.tes;
```
underscope: Since this is now custom container should there be a separate getter for it at all ?

client/components/content-containers/tcc-exam/edit/index.vue
```diff
@@ -33,7 +33,7 @@
         @update="$emit('updateSubcontainer', $event)"
         @delete="$emit('deleteSubcontainer', group, 'group')"
         :group="group"
-        :tes="tes"
+        :elements="elements"
```
underscope: Should we handle backwards compatibility ? 🤔

abasic: I don't think anybody is using this container anymore so I don't we need to worry about it. And also this is a prop for a component which is only used in this container. It passes it on to its child which is also only used here. So we should be good on any breaking changes.

client/store/modules/editor/getters.js
abasic: No there should not. I'll remove it after I merge with develop.

client/store/modules/repository/actions.js
abasic: It is probably best to be more verbose. It is easier for people to know what it refers to, especially people less familiar with the project.

abasic: I made the change inside the store so all the `getters` and `actions` reference `contentElements` but I did not make the changes throughout the app. I was not sure if that would be too excessive or not.

client/store/index.js
```diff
@@ -38,7 +38,6 @@ export default new Vuex.Store({
   getters: {
     ...mapGetters('repositories', ['repositories', 'repositoryQueryParams']),
     ...mapGetters('activities', ['activities']),
-    ...mapGetters('tes', ['tes']),
```
abasic, line 41: @underscope @vladimyr 

I removed this because, from what I can tell, it is not used anywhere. I don't think any of them are used except the `repositories/repositories` one. I believe this was written so that we have a nested module structure but also have these getters globally available. If that is true then I would remove all of them including the repositories one.  The code that generates these getters is convoluted and I do not see a reason to keep it for just one. And if we do keep it I would recommend changing its name. It being named `mapGetters` is deceiving since that is the name of a `Vuex` method.

underscope: 👍 It's a relict of the past. Feel free to do so.

underscope:
@abasic Removal of unsaved items is not working. Let me know if I should wrap it up.

client/components/common/tce-core/QuestionContainer/index.vue
```diff
@@ -39,7 +39,6 @@
       <controls
         @edit="edit"
         @save="save"
-        @remove="remove"
```
abasic, line 42: The controls do not emit a remove event so I removed the hander.

client/components/content-containers/tcc-assessment-pool/edit/index.vue
```diff
@@ -9,68 +9,72 @@
...
   },
   methods: {
-    ...mapActions('repository/tes', ['save', 'update', 'remove']),
-    ...mapMutations('repository/tes', ['add']),
     addAssessment(assessment) {
-      const data = { ...assessment, _cid: cuid() };
-      this.add(data);
-      this.selected.push(data._cid);
+      const cid = cuid();
+      this.$emit('addElement', { ...assessment, _cid: cid });
```
abasic, line 66: I reverted the local state for unsaved assessments i introduced in this PR and moved it to the store where it was originally. The reason was the support for canceling. It needed to show a confirmation modal before removing the changes. With the store it is supported for every deletion event. With a local state I would need to do it inside the custom container or the `tce-question-container`. I believe the store is a better approach because it provides a singular interface for all custom containers and it does not introduce additional dependencies to components (example event bus). 

Looking at it now the usage of local state was a mistake. I believe the usage of local state for elements (maybe other things as well) in custom containers should be discouraged to keep the various implementation of custom containers streamlined and similar.

client/store/helpers/actions.js
```diff
@@ -20,6 +20,8 @@ export default function ($apiUrl) {
     return api.fetch(opts).then(res => commit('reset', res));
   };
 
+  const add = ({ commit }, model) => commit('add', model);
```
abasic, line 23: Added this so custom containers can add elements to the store without them being saved.

client/components/content-containers/tcc-assessment-pool/edit/index.vue
```diff
@@ -80,46 +84,47 @@ export default {
     isSelected(assessment) {
       return this.selected.includes(assessment._cid);
     },
+    clearSelected() {
```
abasic, line 87: The selected array was not being cleared on deletion.

abasic:
@underscope I fixed the issue and made some changes. I wrote a couple of comments so you know why I did them.

kjuej:
@abasic  when creating an assessment, and clicking cancel before saving, the assessment does not get deleted.

client/store/helpers/actions.js
underscope: @abasic Don't we have add mutation already. What's the drive for action?

abasic: @underscope The drive was to avoid calling the mutation directly but go through the full flow of calling an action that calls the mutation. I know it is not necessary to get the desired result but wanted to keep it inline with all the other event handlers that are there.

underscope:
@kjuej

kjuej:
@underscope Seems to be working as expected.

PR #449, opened 2020-02-20
Updated custom container support

abasic:
This PR updates the custom container support by making the following changes:

- The `util` file and all the methods inside of it are no longer required. If the `fetch` method is missing then the container uses the default fetch and statics resolution methods. This allows for more streamlined support for containers that have the same, simple structure that the default containers have. Basically, it allows for different UI for default containers.
- Updated container fetching logic so it only tries to fetch the supported containers for that activity.
- Custom containers are now treated as templates instead of just containers. What this allows is using the same container template for different types of containers. The benefit of this is that a container template can be imported multiple times with different configurations. The `type` property of a custom container is replaced with a `templateId` property. The `templateId` needs to be added to the configuration of a container as the `templateId` property in the `Tailor` config file.
- Content elements of all the containers are now fetched, instead of just the elements of the root containers and the first level of sub-containers. This allows the creation of custom containers with multiple levels of depth (sub-containers having sub-containers).
- Updated the default schema to support custom containers.


client/components/content-containers/tcc-exam/info.js
```diff
@@ -1,4 +1,4 @@
 module.exports = {
-  type: 'EXAM',
+  templateId: 'EXAM',
```
underscope, line 2: Are we handling deprecation? (somebody has custom container without templateId; default to type).

client/components/editor/structure/ContentContainers/index.vue
```diff
@@ -57,15 +59,16 @@ export default {
     displayHeading: { type: Boolean, default: false },
     type: { type: String, required: true },
     label: { type: String, required: true },
+    templateId: { type: String, default: null },
```
underscope, line 60: Consider ordering these props by importance (items/containers, type, templateId, parentId...)

client/store/modules/editor/getters.js
```diff
@@ -20,9 +21,23 @@ export const contentContainers = (_state, getters, { repository }) => {
   }, {});
 };
 
+export const contentContainers = (_state, getters, { repository }) => {
+  const { rootContainerGroups } = getters;
+  const { items: activities } = repository.activities;
+  const parents = flatMap(rootContainerGroups);
+  const children = findChildren(activities, parents);
```
underscope: Consider putting parents as first argument (You are trying to find children of those parents so it feels like that's the primary param)

client/store/modules/editor/getters.js
```diff
@@ -20,9 +21,23 @@ export const contentContainers = (_state, getters, { repository }) => {
...
+};
+
 export const assessments = (_state, getters, rootState) => {
   const { items: tes } = rootState.repository.tes;
   const { activity } = getters;
   if (!activity) return [];
   return filter(tes, { activityId: activity.id, type: 'ASSESSMENT' });
 };
+
+function findChildren(activities, parents) {
```
underscope: You should have this one within activity utils.

server/shared/publishing/helpers.js
```diff
@@ -178,9 +194,9 @@ function fetchDefaultContainer(container) {
     }));
 }
 
-function fetchCustomContainers(parent) {
+function fetchCustomContainers(parent, types) {
```
underscope: Would it make sense to inherit it from configuration based on parent type (rather than passing additional params)

server/shared/publishing/helpers.js
```diff
@@ -151,23 +151,39 @@ function publishAssessments(parent) {
...
+    if (containerRegistry.getContentFetcher(templateId)) {
+      acc.custom.push(templateId);
+    } else {
+      acc.defaults.push(type);
+    }
+    return acc;
+  }, { defaults: [], custom: [] });
+}
+
+function fetchDefaultContainers(parent, types) {
```
underscope: See comment below.

server/shared/publishing/helpers.js
```diff
@@ -151,23 +151,39 @@ function publishAssessments(parent) {
...
+  const where = { type: types };
+  return parent.getChildren({ where }).map(fetchDefaultContainer);
```
underscope: It feels like we can turn this into one query 🤔

server/shared/publishing/helpers.js
```diff
@@ -151,23 +151,39 @@ function publishAssessments(parent) {
...
       const publishedAs = get(config, 'publishedAs', 'container');
-      return { ...it, publishedAs };
+      const { templateId } = config;
+      return { ...it, templateId, publishedAs };
     });
     return containers.concat(mappedContainers);
   }, []);
 }
 
+function getGroupedContainerTypes(typeConfigs) {
```
underscope: Confusing name/doc might be helpful. typeConfigs 🤔 Feels like it might be nicer if we just pass activity and rename method to something like getActivityContainerTypes ?

server/shared/publishing/helpers.js
```diff
@@ -151,23 +151,39 @@ function publishAssessments(parent) {
...
+  return typeConfigs.reduce((acc, { type, templateId }) => {
+    if (containerRegistry.getContentFetcher(templateId)) {
```
underscope: getContentFetcher 🤔

server/shared/publishing/helpers.js
```diff
@@ -151,23 +151,39 @@ function publishAssessments(parent) {
...
-  const isCore = it => !containerRegistry.getStaticsResolver(it.type);
-  const coreTypes = typeConfigs.filter(isCore).map(it => it.type);
+  const { defaults, custom } = getGroupedContainerTypes(typeConfigs);
 
-  return Promise.all([
-    parent.getChildren({ where: { type: coreTypes } }).map(fetchDefaultContainer),
-    fetchCustomContainers(parent)
-  ])
-  .reduce((containers, groupedContainers) => {
+  const fetches = [];
```
underscope: fetches 🤔

server/shared/publishing/helpers.js
```diff
@@ -151,23 +151,39 @@ function publishAssessments(parent) {
...
+  if (defaults.length) fetches.push(fetchDefaultContainers(parent, defaults));
+  if (custom.length) fetches.push(fetchCustomContainers(parent, custom));
+  if (!fetches.length) return [];
+  return Promise.all(fetches).reduce((containers, groupedContainers) => {
     const mappedContainers = groupedContainers.map(it => {
       const config = find(typeConfigs, { type: it.type });
       const publishedAs = get(config, 'publishedAs', 'container');
-      return { ...it, publishedAs };
+      const { templateId } = config;
```
underscope: Consider inlining it (don't greatly benefit from this destructuring)

server/shared/content-plugins/containerRegistry.js
```diff
@@ -9,41 +9,42 @@ const EXTENSIONS_LIST = '../../../extensions/content-containers/index';
 class ContainerRegistry extends BaseRegistry {
   constructor() {
     super('container', containerList, EXTENSIONS_LIST);
+    this._contentFetcher = {};
```
underscope: 🤔 Can we improve the naming

client/content-plugins/ContainerRegistry.js
```diff
@@ -2,7 +2,8 @@ import ComponentRegistry from './ComponentRegistry';
 import containerList from 'shared/core-containers';
 import { getContainerName } from 'tce-core/utils';
 
-const ATTRS = ['type', 'version'];
-const options = ['container', containerList, ATTRS, getContainerName];
+const ATTRS = ['type', 'templateId', 'version'];
+const getCondition = templateId => it => it.templateId === templateId;
+const options = ['container', containerList, ATTRS, getContainerName, getCondition];
```
underscope: A bit bloated now. We should probably transform it to object or group some params.

client/content-plugins/ElementRegistry.js
```diff
@@ -3,6 +3,7 @@ import elementList from 'shared/core-elements';
 import { getComponentName } from 'tce-core/utils';
 
 const ATTRS = ['name', 'type', 'subtype', 'version', 'schema', 'initState', 'ui'];
-const options = ['element', elementList, ATTRS, getComponentName];
+const getCondition = type => it => it.subtype === type || it.type === type;
+const options = ['element', elementList, ATTRS, getComponentName, getCondition];
```
underscope: See prev comment

client/content-plugins/ComponentRegistry.js
```diff
@@ -37,23 +42,21 @@ export default class ComponentRegistry {
     const element = isExtension
       ? (await import(`extensions/content-${_type}s/${path}`)).default
       : (await import(`components/content-${_type}s/${path}`)).default;
-    const type = isQuestion(element.type)
-      ? processAnswerType(element.subtype)
-      : element.type;
-    const componentName = this._getName(type);
+    const id = element.templateId || getType(element);
```
underscope: 🤔 Named method explaining the rationale / logic

client/content-plugins/ComponentRegistry.js
```diff
@@ -37,23 +42,21 @@ export default class ComponentRegistry {
...
   }
 
   all() {
     return cloneDeep(this._registry);
   }
 
-  get(type) {
-    if (!type) return this.all();
+  get(type, defaultToAll = true) {
+    if (!type) return defaultToAll ? this.all() : null;
```
underscope: For what is this being used?

client/components/content-containers/tcc-exam/info.js
abasic: No but we can. The reason I did not do it because AICPA was the only project at the time that was using it plus this is not an officially released feature yet.

server/shared/publishing/helpers.js
abasic: I am not sure that it would. Both this method for fetching custom containers and the method for fetching the default containers get their types from the fetchContainers method. So, if we remove it from one of them we need to remove it from the other one. And I do not see what do we gain by doing that.

server/shared/publishing/helpers.js
abasic: I see that you commented this one as well. They way we should do this if we want the types fetched in these methods is to build a type look up on server start in the `config/shared/activities.js` file. This is something that should probably be done no matter what path we take.

server/shared/publishing/helpers.js
abasic: True. If we pass a Content Elements include to the `getChildren()` method we can fetch them all at once.

server/shared/publishing/helpers.js
abasic: You are right. I'll make that change.

client/content-plugins/ComponentRegistry.js
abasic: In the [AddElement](https://github.com/ExtensionEngine/tailor/blob/15077f809605b0f25e002d24905e70d244b18629/client/components/common/tce-core/AddElement/index.vue#L106) component and in the system settings for listing all the elements. But, looking at it now, I should have just made a separate method that returns all of them.

server/shared/publishing/helpers.js
abasic: We don't really do what I wrote anywhere (as far as I can tell). So for now, I will remove the finding of the container types from the `fetchContainers` method and move it to the sub-methods. And then we can see if we want to go in another way.

underscope:
@abasic Is this still WIP ?

abasic:
@underscope it is not but you added it so I did not want to remove it.

abasic:
@underscope I have merged the latest code, added the legacy support, and tested it out.

PR #457, opened 2020-03-08
Fixed issue with metadata validation

underscope:
- Validate only current meta

@kronicker ☝️


underscope:
@kjuej Meta validation should be broken on current develop. Please test this one (applicable for input, textarea and file meta)

PR #464, opened 2020-03-10
add search on tag filter drop down menu

boskoradalj:


client/components/catalog/TagFilter.vue
```diff
@@ -6,22 +6,25 @@
...
-        <v-list-item-action class="mr-2">
-          <v-checkbox :value="it.isSelected" />
-        </v-list-item-action>
-        <v-list-item-content class="text-left">
-          <v-list-item-title>{{ it.name }}</v-list-item-title>
-        </v-list-item-content>
-      </v-list-item>
-      <v-list-item v-if="!options.length">
-        <v-list-item-content class="text-left">
-          <v-list-item-title>No available tags</v-list-item-title>
```
underscope: Message needs to be shown if there are no tags or search results

client/components/catalog/TagFilter.vue
```diff
@@ -6,22 +6,25 @@
...
-        </v-list-item-content>
-      </v-list-item>
+      <v-text-field
+        v-model="search"
+        :hide-details="true"
+        prepend-inner-icon="mdi-magnify"
+        placeholder="Search..."
+        solo />
+      <div class="item-container">
```
underscope: Consider improving naming => options-container. If you adopt the previous comment then you can remove it completely.

client/components/catalog/TagFilter.vue
```diff
@@ -32,23 +35,28 @@ import map from 'lodash/map';
...
     options() {
       return map(this.tags, it => {
         const isSelected = !!find(this.tagFilter, { id: it.id });
         return { ...it, isSelected };
       });
+    },
+    filteredTags() {
+      const { options, search } = this;
+      return search
+        ? options.filter(it => it.name.toLowerCase().includes(search.toLowerCase()))
```
underscope: [See](https://github.com/ExtensionEngine/tailor/blob/develop/client/components/repository/Outline/index.vue#L82) how we filter activities

client/components/catalog/TagFilter.vue
```diff
@@ -6,22 +6,25 @@
...
-        <v-list-item-content class="text-left">
-          <v-list-item-title>{{ it.name }}</v-list-item-title>
-        </v-list-item-content>
-      </v-list-item>
-      <v-list-item v-if="!options.length">
-        <v-list-item-content class="text-left">
-          <v-list-item-title>No available tags</v-list-item-title>
-        </v-list-item-content>
-      </v-list-item>
+      <v-text-field
```
underscope: v-text field probably shouldn't be part of list 🤔 and list items should be direct children.

client/components/catalog/TagFilter.vue
```diff
@@ -6,22 +6,25 @@
...
+        prepend-inner-icon="mdi-magnify"
+        placeholder="Search..."
+        solo />
+      <div class="item-container">
+        <v-list-item
+          v-for="tag in filteredTags"
+          :key="tag.id"
+          @click="$emit('update', tag)">
+          <v-list-item-action class="mr-2">
+            <v-checkbox v-model="tag.isSelected" />
```
underscope: You are mutating tag state (computed) 🚨 and emitting update on event

client/components/catalog/Container.vue
```diff
@@ -135,7 +131,7 @@ export default {
     }
   },
   created() {
-    this.fetchTags();
+    this.fetchTags({ addNewTag: false });
```
underscope: Use something like associated/assigned as flag rather than the current functionality you are adding it for.  It should be falsy by default (no need to specify it).

client/store/modules/repositories/actions.js
```diff
@@ -45,8 +45,11 @@ const pin = ({ commit, getters }, { id, pin }) => {
   });
 };
 
-const fetchTags = ({ commit }) => {
-  return tagApi.fetch().then(tags => commit('fetchTags', tags));
+const fetchTags = ({ commit }, payload) => {
```
underscope: No need to modify this one since store will only have assigned ones (non assigned are only being used / and fetched within modal).

client/components/catalog/Card/Tags/index.vue
```diff
@@ -55,6 +56,13 @@ export default {
       const tagCount = get(this.repository, 'tags.length', TAG_LIMIT) - 1;
       const length = [15, 6, 5][clamp(tagCount, 0, 2)];
       return truncate(tag, { length });
+    },
+    async loadDialog() {
```
underscope: No need for this, you can just do it upon dialog show 😉

client/components/catalog/TagFilter.vue
```diff
@@ -5,40 +5,57 @@
         <v-icon color="primary lighten-4">mdi-tag-outline</v-icon>
       </v-btn>
     </template>
-    <v-list>
+    <v-sheet class="pa-4 primary darken-1">
+      <v-text-field
+        v-model="search"
+        label="Search Tags"
+        dark
```
underscope: Inline booleans

client/components/catalog/TagFilter.vue
```diff
@@ -5,40 +5,57 @@
...
+        flat
+        solo-inverted
+        hide-details
+        clearable
+        clear-icon="mdi-close-circle-outline" />
```
underscope: Should come before boolean attrs

client/components/catalog/TagFilter.vue
```diff
@@ -5,40 +5,57 @@
...
-      </v-list-item>
-      <v-list-item v-if="!options.length">
-        <v-list-item-content class="text-left">
-          <v-list-item-title>No available tags</v-list-item-title>
+          <v-list-item-title>{{ tag.name }}</v-list-item-title>
         </v-list-item-content>
       </v-list-item>
     </v-list>
+    <v-list v-else>
+      <v-list-item>No tags found</v-list-item>
```
underscope: No need to create list / list item. can be just div.

client/components/catalog/TagFilter.vue
```diff
@@ -5,40 +5,54 @@
         <v-icon color="primary lighten-4">mdi-tag-outline</v-icon>
       </v-btn>
     </template>
-    <v-list>
+    <v-sheet class="pa-4 primary darken-1">
```
underscope: Dont forget to set tile attr 😉

client/components/catalog/TagFilter.vue
```diff
@@ -48,7 +62,11 @@ export default {
 .v-list {
   max-height: 18.75rem;
   overflow-y: auto;
-  padding: 1rem 0;
+  border-radius: 0 !important;
```
underscope: No need to do so just use tile attr 😉

client/components/catalog/TagFilter.vue
boskoradalj: from some reason does not apply tile style

server/tag/tag.model.js
```diff
@@ -21,6 +21,21 @@ class Tag extends Model {
...
+  static getAssociated(user) {
+    const Repository = this.sequelize.model('Repository');
+    const User = this.sequelize.model('User');
+    const includeRepository = {
+      model: Repository,
+      as: 'repositories',
+      attributes: ['id'],
+      required: true
+    };
+    if (!user.isAdmin()) {
```
underscope: You have to check if user is passed ;)

server/tag/tag.model.js
```diff
@@ -21,6 +21,21 @@ class Tag extends Model {
...
+      includeRepository.include = [{ model: User, attributes: ['id'], where: { id: user.id } }];
+    }
+    return includeRepository;
```
underscope: Make this func do fetching rather then returning query params.

client/store/modules/repositories/actions.js
```diff
@@ -46,7 +46,7 @@ const pin = ({ commit, getters }, { id, pin }) => {
 };
 
 const fetchTags = ({ commit }) => {
-  return tagApi.fetch().then(tags => commit('fetchTags', tags));
+  return tagApi.fetch({ associated: true }).then(tags => commit('fetchTags', tags));
```
underscope: Over the limit, move .then to the new line 😉

client/components/catalog/TagFilter.vue
underscope: You have outside border radius which is applied by v-menu component (setting radius on top corners of v-sheet and bottom for list; these ones are ok). But there is a border radius at the bottom of v-sheet (bordering with v-list). If you set tile attr that one will be gone.

image

client/components/catalog/TagFilter.vue
```diff
@@ -5,40 +5,56 @@
...
+      <v-text-field
+        v-model="search"
+        label="Search Tags"
+        clear-icon="mdi-close-circle-outline"
+        dark flat solo-inverted hide-details
+        clearable />
+    </v-sheet>
+    <v-list
+      v-if="filteredTags.length"
+      :tile="false">
```
underscope: Just saw that this one is true by default No need to switch it off.

client/components/catalog/TagFilter.vue
```diff
@@ -5,40 +5,56 @@
...
-        </v-list-item-content>
-      </v-list-item>
-      <v-list-item v-if="!options.length">
-        <v-list-item-content class="text-left">
-          <v-list-item-title>No available tags</v-list-item-title>
+          <v-list-item-title>{{ tag.name }}</v-list-item-title>
         </v-list-item-content>
       </v-list-item>
     </v-list>
+    <div v-else class="no-result">
```
underscope: No need for new class. You can just set it to => `white py-3`

kjuej:
@boskoradalj There's a vee-validate error displayed in the console each time you type in a new tag and save it:
Screen Shot 2020-03-27 at 08 59 14

kjuej:
@boskoradalj Should i get an error message if i try to enter a tag that already exists on the Course Card? (Currently nothing really happens, no new tag is saved, and the modal gets closed upon saving)

kjuej:
@boskoradalj When tag search is used, all other previously selected tags in the dropdown list no longer show the checked mark:
Screen Shot 2020-03-27 at 09 11 33

kjuej:
@boskoradalj The search input doesn't get cleared upon closing the tag filter dropdown list (upon reopening it, the search input has the same value it had before closing).

kjuej:
@boskoradalj Suggestion: 
Maybe keep the clear icon on the right side when typing values in search input?
Screen Shot 2020-03-27 at 09 34 31

client/components/catalog/TagFilter.vue
```diff
@@ -1,44 +1,61 @@
 <template>
-  <v-menu :close-on-content-click="false" offset-y>
+  <v-menu
+    @update:return-value="search=''"
```
underscope: missing spaces before/after `=`

client/components/catalog/TagFilter.vue
```diff
@@ -1,44 +1,61 @@
...
+    offset-y>
     <template v-slot:activator="{ on }">
       <v-btn v-on="on" icon text>
         <v-icon color="primary lighten-4">mdi-tag-outline</v-icon>
       </v-btn>
     </template>
-    <v-list>
+    <v-sheet
+      class="pa-4 primary darken-1"
+      tile>
```
underscope: Should come before class.

kjuej:
@boskoradalj @underscope Everything seems to work as expected.

PR #473, opened 2020-04-03
Add fix for broken tag filer 

boskoradalj:


kjuej:
Everything seems to be working as expected now.

PR #491, opened 2020-04-20
Improve HTML meta

kronicker:
This PR 
- adds the ability to pass Quill editor options (e.g. `toolbar.modules`) to   Quill editor used in HTML meta
- removes unused prop `quillOptions` in favor of `meta.options`
- fixes broken reference

Note: I think we could've made sensible defaults and make it more strict, but this way we don't change existing behavior and add flexibility while we don't deprecate it in favor of Jodit Meta 🤞 

### UI changes
<table>
<tr>
<td><strong>Before</strong></td>
<td><strong>After<strong></td>
</tr>
<tr>
<td>
Screenshot 2020-04-23 at 00 52 21
</td>
<td>
Screenshot 2020-04-23 at 01 01 06
</td>
</tr>
</table>

### DIFF
--- a/client/components/common/Meta/Html.vue
+++ b/client/components/common/Meta/Html.vue
@@ -19,7 +19,7 @@
 import { quillEditor as QuillEditor } from 'vue-quill-editor';
 import some from 'lodash/some';
 
-const quillOptions = () => ({
+const defaultOptions = () => ({
   modules: {
     toolbar: [
       ['bold', 'italic', 'underline'],
@@ -42,7 +42,7 @@ export default {
     };
   },
   computed: {
-    options: ({ meta }) => ({ ...quillOptions(), ...meta.options })
+    options: ({ meta }) => ({ ...defaultOptions(), ...meta.editorOptions })
   },
   methods: {
     update(quill) {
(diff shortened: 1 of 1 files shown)


kronicker:
@vladimyr @underscope I've updated the UI just a bit to be more aligned with other meta inputs. Let me know what do you think?

client/components/common/Meta/Html.vue
```diff
@@ -86,10 +88,10 @@ export default {
     border-color: currentColor;
   }
 
-  .quill-input-label {
+  .quill-label {
     position: absolute;
-    top: -23px;
-    color: rgba(0, 0, 0, 0.6);
+    top: -21px;
+    font-size: 14px;
```
vladimyr, line 94: I'm ok with it if pixels are used elsewhere but in general, we shouldn't use them for font sizing.

vladimyr:
@kronicker Looks sexy 😎 I especially like this new convention of advertising UI changes through GFM tables, those are some serious pro moves 🤓

client/components/common/Meta/Html.vue
kronicker: I've initially set changes in `rem`s but the figured resto of the style is in pixels so I changed back to pixels. 🤷‍♂️

PR #493, opened 2020-04-23
Reroute to catalog after repository removal

kjuej:


underscope:
@kjuej

kjuej:
@underscope Works as expected.

PR #500, opened 2020-04-30
Element toolbar bugfix

borna223:
Fixes this (matching question is selected in editor but "Numerical Response" is displayed in toolbar):
Screenshot 2020-04-30 at 11 28 01


underscope:
@kjuej

kjuej:
@borna223 @underscope Works as expected.

PR #505, opened 2020-05-06
Show active users

boskoradalj:


client/components/editor/structure/ContentContainers/Container.vue
```diff
@@ -30,17 +30,31 @@
...
+          :element-style="getElementStyle(item.contentId)">
+          <div
+            :class="{ inactive: !hasActiveUsers(item.contentId) }"
+            class="active-users-wrapper">
+            <active-users
+              v-if="getActiveUsers('element', item.contentId)"
+              :users="getActiveUsers('element', item.contentId)"
+              :size="26"
+              vertical
+              tooltip-right />
```
MiroDojkic: This doesn't seem to be use in the `ActiveUsers` component?

client/components/editor/structure/ContentContainers/Container.vue
```diff
@@ -89,4 +115,17 @@ export default {
   min-height: 2.25rem;
   margin-bottom: 0.5rem;
 }
+
+// /deep/ .active-users-wrapper {
```
MiroDojkic: What's the reason behind this being commented out?

client/components/editor/structure/ContentContainers/Container.vue
```diff
@@ -30,17 +30,31 @@
...
+          :element="item"
+          :element-style="getElementStyle(item.contentId)">
+          <div
+            :class="{ inactive: !hasActiveUsers(item.contentId) }"
+            class="active-users-wrapper">
+            <active-users
+              v-if="getActiveUsers('element', item.contentId)"
+              :users="getActiveUsers('element', item.contentId)"
+              :size="26"
+              vertical
```
MiroDojkic: `vertical` layout seems to play nicely with the positioning implemented by [commented out CSS](https://github.com/ExtensionEngine/tailor/pull/505/files?file-filters%5B%5D=.vue#diff-ff0314a01aa905cd12b0dc35c3705e35R119).
Does it make sense to use it without it?

client/components/editor/structure/ContentContainers/Container.vue
MiroDojkic: In case we end up using this `/deep/` should be replaced with `::v-deep`.
I'd also suggest scoping this to the first ancestor element.

client/components/editor/structure/ContentContainers/Container.vue
```diff
@@ -77,9 +92,20 @@ export default {
       const isFirstChild = newPosition === -1;
       const context = { items, newPosition, isFirstChild, insert: true };
       this.insertElement({ element, context });
+    },
+    hasActiveUsers(elementId) {
+      const activeUsers = this.getActiveUsers('element', elementId);
+      return activeUsers.length;
```
MiroDojkic: Since the name of the prop implies `Boolean` type, it might be more appropriate to return `!!activeUsers.length`.

client/components/editor/structure/ContentContainers/Container.vue
```diff
@@ -30,17 +30,31 @@
         <content-element
           :set-width="setWidth"
           :dragged="dragged"
-          :element="item" />
+          :element="item"
+          :element-style="getElementStyle(item.contentId)">
+          <div
+            :class="{ inactive: !hasActiveUsers(item.contentId) }"
```
MiroDojkic: This class was used to fade-in the component. Is this behavior intentionally removed?
In case we're not keeping in, we might as well remove the class and the computed property.

MiroDojkic: In case we're keeping it, would `hidden` be more appropriate to describe the effect?

client/components/repository/index.vue
```diff
@@ -66,6 +75,7 @@ export default {
   async created() {
     const { repositoryId } = this;
     await this.initialize(repositoryId);
+    this.setupActivityUsersApi(`/repository/${repositoryId}/active-users`);
```
MiroDojkic: Why is this needed?
I might be missing something but as far as I can see, `activeUsers` module's actions are not composed of the ones generated by `generateActions` (from `client/store/helpers/actions.js`)

client/components/repository/index.vue
```diff
@@ -90,6 +101,11 @@ export default {
   display: flex;
   flex-direction: column;
 
+  .repository-header {
```
MiroDojkic: We could use [Vuetify's flex utility classes](https://vuetifyjs.com/en/styles/flex/#flex) for this. 💄 🙂

client/components/repository/index.vue
```diff
@@ -1,29 +1,33 @@
...
+          v-for="tab in tabs"
+          :key="tab.name"
+          :to="{ name: tab.route }"
+          active-class="tab-active"
+          ripple exact
+          class="px-4">
+          <v-icon class="pr-2">mdi-{{ tab.icon }}</v-icon>{{ tab.name }}
+        </v-tab>
+      </v-tabs>
+      <active-users :users="getActiveUsers('repository', repositoryId)" />
```
MiroDojkic: Didn't test this, but curried getter can't cache results and will execute on every re-render.
I don't think it will have an effective impact in this case, but we can help this by moving curried getter call to computed property:
```js
computed: {
  activeUsers() {
    return this.getActiveUsers('repository', this.repositoryId);
  }
}
```

client/components/editor/index.vue
```diff
@@ -19,27 +22,36 @@
...
     ...mapGetters('editor', ['activity', 'contentContainers'])
   },
-  methods: mapActions('repository', ['initialize']),
+  methods: {
+    ...mapActions('activeUsers', { setupActivityUsersApi: 'setEndpoint' }),
+    ...mapActions('repository', ['initialize'])
+  },
   async created() {
     const { repositoryId: currentRepositoryId, repository: storeRepository } = this;
+    this.setupActivityUsersApi(`/repository/${currentRepositoryId}/active-users`);
```
MiroDojkic: Same [comment](https://github.com/ExtensionEngine/tailor/pull/505/files?file-filters%5B%5D=.vue#diff-a9c8a39f12211587a3c676e72c3b5c34R78) applies here.

client/components/editor/index.vue
```diff
@@ -2,6 +2,9 @@
   <div class="editor-container">
     <template v-if="!isLoading">
       <toolbar :element="selectedElement" />
+      <template slot="active-users">
+        <active-users :users="getActiveUsers('activity', activityId)" />
```
MiroDojkic: Same [comment](https://github.com/ExtensionEngine/tailor/pull/505/files?file-filters%5B%5D=.vue#diff-a9c8a39f12211587a3c676e72c3b5c34R20) applies here.

server/active-user/events.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+const Events = {
```
MiroDojkic, line 3: It would be cool if we could share these between client and server. 🤔

server/shared/sse/channel.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  get id() {
+    return this._id;
+  }
+
+  add(connection) {
+    this._connections.set(connection.id, connection);
+    connection.on('close', () => this.remove(connection));
+    return this;
```
MiroDojkic, line 19: What is the reason behind returning `this` from `add`, `remove` and `send` methods?
It seems redundant since we're already calling methods on the same instance which is in fact returned from methods. 😁

server/shared/sse/index.js
```diff
@@ -0,0 +1,111 @@
...
+
+  static channel(channelId) {
+    channelId = String(channelId);
+    return this._channels.get(channelId);
+  }
+
+  join(channelId) {
+    channelId = String(channelId);
+    this._ensureChannel(channelId);
+    const channel = this.constructor.channel(channelId);
```
MiroDojkic, line 83: What's the reason for using this instead of `SSEConnection.channel()`?

client/components/common/ActiveUsers.vue
```diff
@@ -0,0 +1,78 @@
...
+</template>
+
+<script>
+export default {
+  name: 'active-users',
+  props: {
+    users: { type: Array, default: () => [] },
+    size: { type: Number, default: 30 },
+    vertical: { type: Boolean, default: false },
+    tooltipRight: { type: Boolean, default: false }
```
MiroDojkic: Are we using this prop anywhere?

client/components/common/ActiveUsers.vue
```diff
@@ -0,0 +1,78 @@
...
+  <div :class="{ vertical }" class="active-users">
+    <v-avatar
+      v-for="{ id, email, palette, profileImage } in activeUsers"
+      :key="id"
+      :color="palette.background"
+      :style="{ boxShadow: `0 0 0 2px ${palette.border}` }"
+      :size="size">
+      <img v-if="profileImage" :src="profileImage">
+      <span v-else :style="{ color: palette.text }">
+        {{ email[0].toUpperCase() }}
```
MiroDojkic: Looks like a good use case for a [filter](https://vuejs.org/v2/guide/filters.html). 🙂

client/components/common/ActiveUsers.vue
```diff
@@ -0,0 +1,78 @@
...
+  props: {
+    users: { type: Array, default: () => [] },
+    size: { type: Number, default: 30 },
+    vertical: { type: Boolean, default: false },
+    tooltipRight: { type: Boolean, default: false }
+  },
+  data: () => ({ activeUsers: [] }),
+  methods: {
+    setActiveUsers(users) {
+      this.activeUsers = users;
```
MiroDojkic: What's the reason for duplicating data passed through the prop?
If we need that, I'd suggest we deep clone prop instead of just copying the reference.
If not, please remove the related logic.

client/components/common/ActiveUsers.vue
```diff
@@ -0,0 +1,78 @@
...
+      },
+      immediate: true
+    }
+  }
+};
+</script>
+
+<style scoped lang="scss">
+.active-users {
+  display: flex;
```
MiroDojkic: [Vuetify flex utility classes](https://vuetifyjs.com/en/styles/flex/#flex) to the rescue? 😄

client/components/common/ActiveUsers.vue
```diff
@@ -0,0 +1,78 @@
...
+  align-items: center;
+
+  div {
```
MiroDojkic: We could use `.v-avatar` instead of `div` in selectors to be more specific.

client/components/common/ActiveUsers.vue
```diff
@@ -0,0 +1,78 @@
...
+    margin-left: -5px;
+    transition: all 0.2s;
+
+    &:hover {
+      margin-top: -5px;
+      margin-left: 5px;
+      z-index: 1;
+    }
+  }
+  &.shadow-grey div { box-shadow: 0 0 0 1px #d0d0d0; }
```
MiroDojkic: These selectors seem unused?

client/components/common/ActiveUsers.vue
```diff
@@ -0,0 +1,78 @@
...
+.active-users {
+  display: flex;
+  align-items: center;
+
+  div {
+    margin-left: -5px;
+    transition: all 0.2s;
+
+    &:hover {
+      margin-top: -5px;
```
MiroDojkic: This component doesn't provide any `click` behavior. Does it make sense to disable hover effect then?

client/components/common/ActiveUsers.vue
```diff
@@ -0,0 +1,78 @@
...
+};
+</script>
+
+<style scoped lang="scss">
+.active-users {
+  display: flex;
+  align-items: center;
+
+  div {
+    margin-left: -5px;
```
MiroDojkic: This margin will make the ancestor component take less width based on the width of the children. That will cause the first avatar to be slightly cut off on the left side.
We can fix this by applying `margin-left: 0;` for the `.v-avatar:first-of-type`.

I'd also suggest we add some padding to the ancestor component. 💄

client/components/common/mixins/activeUsers.js
```diff
@@ -0,0 +1,50 @@
...
+    this.subscribeToActiveUsers(this.repositoryId);
+    this.fetchActiveUsers(this.repositoryId);
+  },
+  beforeRouteLeave(to, from, next) {
+    if (!trackedRoutes.includes(to.name)) {
+      this.unsubscribeFromActiveUsers(this.context);
+    }
+    // Remove context when leaving route except when navigating
+    // to course route (Outline component)
+    if (to.name === 'repository') this.removeContext(this.context);
```
MiroDojkic: Consider making `trackedRoutes` an object in order to use the constant instead of plain string here.

client/components/editor/ContentElement.vue
```diff
@@ -51,6 +70,24 @@ export default {
...
+      this.removeActiveUserContext(this.context);
+    }
+  },
+  created() {
+    EventBus.on('element:focus', element => {
+      this.isFocused = !!element && (element.id === this.element.id);
+    });
+  },
+  beforeDestroy() {
+    this.removeActiveUserContext(omit(this.context, 'created'));
```
MiroDojkic: `created` is already omitted on the server-side (activeUser.controller). Do we need to omit it here as well?

client/store/modules/activeUsers/actions.js
```diff
@@ -0,0 +1,84 @@
...
+
+const Events = {
+  Add: 'active_user:add',
+  Remove: 'active_user:remove',
+  RemoveSession: 'active_user:remove_session'
+};
+
+const subscribe = ({ state, commit, rootState }) => {
+  const { repositoryId } = rootState.route.params;
+  // const token = localStorage.getItem('JWT_TOKEN');
```
MiroDojkic: Is there a reason to keep this line?

client/store/modules/activeUsers/actions.js
```diff
@@ -0,0 +1,84 @@
...
+  const token = rootState.auth.token;
+  const params = { token };
```
MiroDojkic: Suggestion, since this is the only place where the `token` is used:
```js
const params = { token: rootState.auth.token };
```

client/store/modules/activeUsers/actions.js
```diff
@@ -0,0 +1,84 @@
...
+  const url = urlJoin(baseUrl, api.url.root(repositoryId), '/subscribe');
```
MiroDojkic: How about we move the `/subscribe` path to `api.url` and compose it with the `root` path (in order to keep API URLs in one place), then use it here:
```js
const url = urlJoin(baseUrl, api.url.subscribe(repositoryId));
```

client/store/modules/activeUsers/actions.js
```diff
@@ -0,0 +1,84 @@
...
+  add,
+  fetch,
+  remove,
+  subscribe,
+  unsubscribe
+};
+
+function assignPalette(user, usedPalettes) {
+  const availablePalette = find(palette, p => !usedPalettes.includes(p.id));
+  const colorPalette = availablePalette || sample(palette);
```
MiroDojkic: If we're eventually going to end up with a random recurring palette when the number of users is greater than the number of palettes, does it make sense to just randomly assign the palette in the first place?

client/utils/palette.js
```diff
@@ -0,0 +1,10 @@
+export default [
```
MiroDojkic: Would `avatarPalette.js` describe the content better/more specifically?

client/store/modules/activeUsers/getters.js
```diff
@@ -0,0 +1,45 @@
+import filter from 'lodash/filter';
+import find from 'lodash/find';
+import get from 'lodash/get';
+import map from 'lodash/map';
+import orderBy from 'lodash/orderBy';
+import pick from 'lodash/pick';
+
+export const activeUsers = state => {
```
MiroDojkic, line 6: Couple of minor change suggestions in order to improve readability:
```js
export const activeUsersByEntity = state => {
  const result = { repository: {}, activity: {}, element: {} };
  const activeUsers = Object.values(state.activeUsers);
  activeUsers.forEach(({ contexts, ...user }) => {
    contexts.forEach(context => {
      setUserActivityForContext(result, user, context);
    });
  });
  return result;
};
```

client/store/modules/activeUsers/getters.js
```diff
@@ -0,0 +1,45 @@
...
+
+function mapContext(activeUsers, user, context) {
+  const { repositoryId, activityId, elementId, created } = context;
+  const userData = { ...pick(user, ['id', 'email', 'palette']), created };
+  setEntityActivity(activeUsers, 'repository', repositoryId, userData);
+  if (activityId) setEntityActivity(activeUsers, 'activity', activityId, userData);
+  if (elementId) setEntityActivity(activeUsers, 'element', elementId, userData);
+}
+
+function setEntityActivity(activeUsers, entity, entityId, user) {
```
MiroDojkic, line 33: A couple of ideas to make this more readable
```js
function setEntityActivity(activeUsers, entity, entityId, user) {
  const activityData = pick(user, ['id', 'email', 'palette', 'created']);
  const activeUsersOnEntity = activeUsers[entity][entityId];
  if (!activeUsersOnEntity) {
    activeUsers[entity][entityId] = [activityData];
    return;
  }
  if (find(activeUsersOnEntity, { id: user.id })) return;
  activeUsersOnEntity.push(activityData);
}
```

client/store/modules/activeUsers/mutations.js
```diff
@@ -0,0 +1,75 @@
...
+import omit from 'lodash/omit';
+import pick from 'lodash/pick';
+import remove from 'lodash/remove';
+import { setEndpoint } from '../../helpers/mutations';
+import Vue from 'vue';
+
+const save = (state, user) => {
+  const { activeUsers } = state;
+  user.contexts.forEach(context => {
+    setUserActivity(Vue, activeUsers, user, context);
```
MiroDojkic: Why not use `Vue.set` inside `setUserActivity` instead of passing it as an arg? 🤔

client/store/modules/activeUsers/mutations.js
```diff
@@ -0,0 +1,75 @@
...
+  const { activeUsers } = state;
+  setUserActivity(Vue, activeUsers, user, context);
+};
+
+const sseRemove = (state, { user, context: sourceContext }) => {
+  const existingUser = state.activeUsers[user.id];
+  if (!existingUser) return;
+  const index = existingUser.contexts.findIndex(targetContext => {
+    return isContextEqual(sourceContext, targetContext);
+  });
```
MiroDojkic:
```js
targetContext => isContextEqual(sourceContext, targetContext)
```

server/active-user/store.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+module.exports = new (class extends Map {
```
MiroDojkic, line 3: Nice!

server/shared/sse/channel.js
boskoradalj: we should ask Ilija or Dario :)

server/shared/sse/index.js
boskoradalj: question for Ilija od Dario

MiroDojkic:
@ExtensionEngine/tailor IMHO this feature is a couple of UX tweaks short of being complete! 😄 

1. Delete icon overlaps active users when the user is editing content element:  
Screen Shot 2020-06-02 at 18 06 30.  
One option that seems straightforward would be moving delete icon to element toolbar? 👀 

2. `Active users` on content element looks broken:  
Screen Shot 2020-06-02 at 18 39 42.  
Even when avatars are arranged vertically, it looks like a component collides with the elements layout/container. A potential solution I could think of is to position active users above element container:
Screen Shot 2020-06-02 at 18 20 20.  
This would also allow a greater number of active users to be shown within the width of the element.

3. Active content element's **border** color is the same as the fallback avatar's **background**. 
If we're showing an avatar image, the border color of the element (probably 😅) wouldn't be matched by any of the avatar's colors.
Would it be a good idea if an element's border would have the color of the avatar's border (which is always displayed)?

4. Does it make sense to filter out the current user from the `active users` component?

client/components/common/mixins/activeUsers.js
```diff
@@ -0,0 +1,54 @@
...
+  watch: {
+    sseId: {
+      handler() {
+        if (!this.sseId) return;
+        this.addContext(this.context);
+      },
+      immediate: true
+    }
+  },
+  mounted() {
```
MiroDojkic: Would it be more appropriate to use the `created` hook so we don't risk sending multiple requests on re-renders?

client/components/common/mixins/activeUsers.js
```diff
@@ -0,0 +1,54 @@
...
+    this.subscribeToActiveUsers(this.repositoryId);
```
MiroDojkic, line 42: `subscribe` actions doesn't receive additional argument but instead gets `repositoryId` from `rootState.route`.

client/components/editor/index.vue
```diff
@@ -1,7 +1,11 @@
 <template>
   <div class="editor-container">
     <template v-if="!isLoading">
-      <toolbar :element="selectedElement" />
+      <toolbar :element="selectedElement">
+        <template slot="active-users">
+          <active-users :users="activeUsers" />
+        </template>
+      </toolbar>
```
MiroDojkic: This can be simplified:
```js
<toolbar :element="selectedElement">
  <active-users #active-users :users="activeUsers" />
</toolbar>
```

client/components/editor/structure/ContentContainers/Container.vue
```diff
@@ -30,17 +30,28 @@
...
-          :element="item" />
+          :element="item"
+          :element-style="getElementStyle(item.contentId)">
+          <div
+            class="active-users-wrapper">
+            <active-users
+              v-if="getActiveUsers('element', item.contentId)"
+              :users="getActiveUsers('element', item.contentId)"
+              :size="26" />
+          </div>
```
MiroDojkic: Seems like we could simplify this by removing wrapping `div` altogether by passing class directly to `ActiveUsers` component. Would that work?

client/components/editor/structure/ContentContainers/Container.vue
```diff
@@ -89,4 +108,10 @@ export default {
   min-height: 2.25rem;
   margin-bottom: 0.5rem;
 }
+
+::v-deep .active-users-wrapper {
```
MiroDojkic: Is `v-deep` redundant here?

client/store/modules/activeUsers/actions.js
```diff
@@ -0,0 +1,79 @@
+import api from '@/api/activeUsers';
+import find from 'lodash/find';
+import forEach from 'lodash/forEach';
+import generateActions from '@/store/helpers/actions';
+import palette from 'utils/avatarPalette';
+// import sample from 'lodash/sample';
```
MiroDojkic: Looks like we can remove this.

client/store/modules/activeUsers/actions.js
```diff
@@ -0,0 +1,79 @@
...
+    });
+};
+
+const unsubscribe = ({ commit }) => {
+  feed.disconnect();
+  commit('resetActiveUsers');
+  commit('setSseId', null);
+};
+
+const fetch = ({ commit }, repositoryId) => {
```
MiroDojkic: These lambdas can be written without blocks and return statements.

client/utils/avatarPalette.js
```diff
@@ -0,0 +1,8 @@
+export default [
+  { id: 0, background: '#86c3b5', text: '#fff', border: '#EC407A' },
```
MiroDojkic: Are we still using background and text colors?

kjuej:
@bradalj hovering a user with no given First name and last name:
Screen Shot 2020-07-01 at 14 49 21

PR #509, opened 2020-05-13
Fixed issues with repository graph

underscope:
- Fixed bug with adding activity
- Fixed bug with activity selection (repo node)


kjuej:
@underscope Seems to be working as expected.

PR #525, opened 2020-05-16
fix repository list with tag filter applied

boskoradalj:
…ot contain tag applied with tag filter selection

### DIFF
--- a/client/store/modules/repositories/mutations.js
+++ b/client/store/modules/repositories/mutations.js
@@ -1,4 +1,5 @@
 import { add, fetch, remove, reset, setEndpoint } from '@/store/helpers/mutations';
+import filter from 'lodash/filter';
 import find from 'lodash/find';
 import Vue from 'vue';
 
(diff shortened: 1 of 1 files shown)


kjuej:
@boskoradalj Seems to be working as expected.

client/store/modules/repositories/mutations.js
```diff
@@ -57,6 +58,9 @@ const addTag = (state, { tag, repositoryId }) => {
 const removeTag = (state, { tagId, repositoryId }) => {
   const repository = find(state.items, { id: repositoryId });
   repository.tags = repository.tags.filter(it => it.id !== tagId);
+  if (!state.tagFilter.length) return;
```
zcuric, line 61: You can move this on line 59, to return early as possible, if conditions are not met.

boskoradalj: If I move this check on the top it will stop execution of removing repository tags when there are no tag filters applied. This logic has been added to handle special use case

PR #531, opened 2020-05-22
TCE-CORE improvements

zcuric:
### This PR
- [x] Removes redundant `component` suffix because it adds unnecessary context to custom elements, for example: `Glossary item component`. Confusing for an end user. If needed it should be added explicitly. 
- [x] Passes down `$attrs` from to `AddElement` to enable setting large button and label from custom elements

### DIFF
--- a/client/components/common/tce-core/EmbeddedContainer.vue
+++ b/client/components/common/tce-core/EmbeddedContainer.vue
@@ -4,7 +4,8 @@
     @update="reorderItem"
     :elements="embeds"
     :supported-types="types"
-    :enable-add="!isDisabled && enableAdd">
+    :enable-add="!isDisabled && enableAdd"
+    v-bind="$attrs">
     <template v-slot:list-item="{ element, isDragged }">
       <contained-content
         @save="save(element, 'data', $event)"
--- a/client/components/common/tce-core/EmbeddedContainer.vue
+++ b/client/components/common/tce-core/EmbeddedContainer.vue
@@ -4,8 +4,7 @@
     @update="reorderItem"
     :elements="embeds"
     :supported-types="types"
-    :enable-add="!isDisabled && enableAdd"
-    v-bind="$attrs">
+    :enable-add="!isDisabled && enableAdd">
     <template v-slot:list-item="{ element, isDragged }">
       <contained-content
         @save="save(element, 'data', $event)"
--- a/client/components/common/tce-core/AddElement/index.vue
+++ b/client/components/common/tce-core/AddElement/index.vue
@@ -86,8 +86,8 @@ export default {
   name: 'add-element',
   inject: {
     $teRegistry: '$teRegistry',
-    buttonLarge: { default: false },
-    buttonLabel: { default: 'Add content' }
+    addBtnLarge: { default: false },
+    addBtnLabel: { default: 'Add content' }
   },
   props: {
     show: { type: Boolean, default: false },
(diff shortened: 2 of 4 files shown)


underscope:
@kjuej

client/components/common/tce-core/EmbeddedContainer.vue
```diff
@@ -4,7 +4,8 @@
     @update="reorderItem"
     :elements="embeds"
     :supported-types="types"
-    :enable-add="!isDisabled && enableAdd">
+    :enable-add="!isDisabled && enableAdd"
+    v-bind="$attrs">
```
underscope, line 8: It might be unintuitive from parent (EmbeddedContainer) to set large/label and expect only for a button to change. Consider adding props to the element list (prefixed with addBtn) rather than just using v-bind.

@kronicker Thoughts?

kronicker: @underscope I agree, just props would imply they are modifiers for the container itself, not the add button.

zcuric: @kronicker @underscope I've used provide/inject instead of prop drilling, please check the latest commit.

zcuric: Idea is with provide/inject is that you provide the label and size in the custom content element. I think this is a good approach. For example: 
```js
export default {
  name: 'te-video-playlist',
  provide: {
    buttonLarge: true,
    buttonLabel: 'Add video'
  }
}
```
I'm open for suggestions regarding naming the values.

kronicker: Provide/inject is not a good practice for props since they're (on purpose) not reactive so if the prop changes it wouldn't be propagated. Their main purpose is dependency injection for some advanced plugin patterns or similar use cases. Check the [docs](https://vuejs.org/v2/api/#provide-inject).

zcuric: @kronicker I've checked the docs before and this is a valid use case for using provide/inject because non of these props are reactive and don't need reactivity. 

Prop drilling is anti pattern and to pollute components with unrelated props is not good practice.

kronicker: Hmmm... Still not sure about it since I know vue team advocated against using this interface for props 🤔 
@underscope @vladimyr ?

zcuric: Can I get the source for `vue team advocated against`? Custom content elements and custom containers fall into the category of advance libs/components/plugins.

kjuej:
@underscope @zcuric Since the "component" phrase was removed, should it not appear here as well? 
Screen Shot 2020-05-22 at 17 07 06
The rest seems to function as expected.

client/components/common/tce-core/EmbeddedContainer.vue
underscope: Hey Zdravko,

I understand the concern of 'prop drilling' and I respect the fact that you wanted to optimize this. 

We should be able to consume `tce-core` components as building blocks for various app segments without the understanding of what they consist of (and whether we can tweak them based on injected configuration). All component aspects should be fully configurable using props and therefore I have to fully agree with Toma regarding this one (without even entering the argument is DI good solution for this problem, as suggested in some articles, and at what point we can consider something as 'prop drilling').

A better alternative for this one might be to create a version with slot for add, so you can configure it without the need to pass the arguments.

In addition to that please extract placeholder change into separate PR since it is unrelated and brings up other concerns.

Thanks!

zcuric: @underscope I have to disagree with both of you. My concern is also that we will end up in slot/prop drilling situation.

While I understand that `tce-core` components should be easily consumed, I think also that these components should have and api to be modified from the extension itself to provide more customisation. `provide/inject` to seems like a perfect fit, IMHO. I could be wrong. This way developer is not concerned with top level props and slots, just with providing the correct data. 

But anyways, I'm closing this for now. I'll think about this more and try to find a better solution.

PR #537, opened 2020-05-26
Remove word component from ElementPlaceholder

zcuric:
### This PR
- [x] Removes "component" from ElementPlaceholder name
- [x] Adds "component" to core elements name


kjuej:
@zcuric Accordion element missing "component'.

PR #545, opened 2020-06-23
Metadata input extensions

### LINKED ISSUE #523: Support metadata input extensions
## I'm submitting a...
- [ ] Bug report  
- [x] Feature request

## Context
Similar to the content elements and custom containers, developers should be able to plug in metadata input extensions.

kronicker:
This PR:
- Enables adding meta extensions and using them
- Extracts and organizes the structure of existing core meta inputs

This PR resolves #523 .


kjuej:
@kronicker Seems to be working as expected!

client/components/common/tce-core/MetaInput.vue
```diff
@@ -0,0 +1,69 @@
...
+  },
+  computed: {
+    originalType: vm => (vm.meta.type || '').toUpperCase(),
+    deprecatedType: vm => deprecatedTypes[vm.originalType],
+    type: vm => get(vm.deprecatedType, 'type', vm.originalType),
+    metaInput: ({ meta, deprecatedType }) => {
+      return !deprecatedType ? meta : {
+        ...meta,
+        ...deprecatedType.config
+      };
```
abasic, line 45:
```suggestion
      if (!deprecatedType) return meta;
      return { ...meta, ...deprecatedType.config };
```

client/components/common/tce-core/MetaInput.vue
```diff
@@ -0,0 +1,69 @@
...
+    },
+    component: vm => getMetaName(vm.type)
+  },
+  methods: { get },
+  watch: {
+    type: {
```
abasic, line 51: If my understanding is correct, the `type` of this meta input never changes. So there should be no need for a watcher. This logic could get triggered in the `mounted` handler.

client/components/common/tce-core/MetaInput.vue
```diff
@@ -0,0 +1,69 @@
...
+      },
+      immediate: true
+    }
+  }
+};
+</script>
+
+<style lang="scss" scoped>
+::v-deep .title {
+  color: #808080;
```
abasic, line 64: Is there a variable for this?

client/components/common/tce-core/MetaInput.vue
```diff
@@ -0,0 +1,69 @@
...
+  font-family: $font-family-secondary;
+  font-size: 14px !important;
```
abasic, line 66: I am not a fan of `!important`. So, if we could do without it, that would be great.

client/components/meta-inputs/meta-datetime/Edit/index.vue
```diff
@@ -6,7 +6,7 @@
       :label="meta.label"
       :placeholder="meta.placeholder" />
     <time-picker
-      v-if="date && (meta.type === 'DATETIME')"
+      v-if="date && !meta.hideTime"
```
abasic, line 9: Something that I like to do is split these conditions if one of them never (or rarely changes). In this case that would be the `meta.hideTime`. I would wrap this `time-picker` component with a template tag and toggle it with the `meta.hideTime` condition.

client/components/common/tce-core/MetaInput.vue
kronicker: This style is c/p but from what I see there is `font-size` with `!important` on `v-application .title` in main stylesheet so I think that's the reason why this is here.

client/components/common/tce-core/MetaInput.vue
kronicker: This is just to guard for all possible uses of meta inputs where if it's not keyed properly can end up just changing `meta` prop so I've added watcher with `immediate` flag instead of just `created` hook.

PR #547, opened 2020-06-27
Swap `joi` with `yup`

vladimyr:
Unfortunately `joi` went the way of the dodo together with `hapi`: https://old.reddit.com/r/node/comments/hfidax/hapi_will_not_be_supported_anymore_by_the_end_of, therefore, removing last bits of it from our codebase.


kjuej:
@vladimyr @underscope Seems to be working as expected.

Issue #532, opened 2020-05-25
Changing carousel height has no effect


kjuej:
## I'm submitting a...

- [x] Bug report  
- [ ] Regression (a behavior that used to work and stopped working in a new release)
- [ ] Feature request
- [ ] Documentation issue or request

## Context
Unable to change carousel element's height
<img width="1189" alt="Screen Shot 2020-05-25 at 09 47 32" src="https://user-images.githubusercontent.com/34127147/82790735-d7a82900-9e6c-11ea-921e-7dc06f2f625a.png">

PR #551, opened 2020-07-03
Set v-carousel height explicitly

hrvojevu:
This PR sets `v-carousel` `height` property explicitly thus fixing the issue https://github.com/ExtensionEngine/tailor/issues/532 in which setting the height of an element did nothing, since default `v-carousel` height would be set to `500px`.

### DIFF
--- a/client/components/content-elements/tce-carousel/edit/index.vue
+++ b/client/components/content-elements/tce-carousel/edit/index.vue
@@ -22,7 +22,7 @@
     <v-carousel
       v-else
       v-model="activeItem"
-      :height="`${height}px`"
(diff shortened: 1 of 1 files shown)


client/components/content-elements/tce-carousel/edit/index.vue
```diff
@@ -22,6 +22,7 @@
     <v-carousel
       v-else
       v-model="activeItem"
+      :height="`${height}px`"
```
underscope, line 25: I think you can just pass the number (Please re-check 😉)

kjuej:

Issue #534, opened 2020-05-25
Remove current slide always removes the last slide


kjuej:
## I'm submitting a...

- [x] Bug report  
- [ ] Regression (a behavior that used to work and stopped working in a new release)
- [ ] Feature request
- [ ] Documentation issue or request

## Context
Remove current slide button on carousel element removes the last slide no matter which slide is currently selected

<img width="845" alt="Screen Shot 2020-05-25 at 14 09 34" src="https://user-images.githubusercontent.com/34127147/82811579-60d15700-9e91-11ea-841c-68bec0fad61b.png">

PR #553, opened 2020-07-07
Sort carousel item indices numerically

hrvojevu:
This PR:
- sorts `tce-carousel` items indices numerically and removes reversing of the array
- uses last index in `indices` array when new slide is created which fixes the issue where last slide would be removed instead of current one
- fixes issue https://github.com/ExtensionEngine/tailor/issues/534


kjuej:

PR #564, opened 2020-07-28
Fix activities reorder on latest schema config

### LINKED ISSUE #563: Activities reodering not working with latest schema config structure
## I'm submitting a...
- [ ] Bug report
- [x] Regression
- [ ] Feature request

## Context
With the latest schema config, activities reorders throws an error.

## Expected Behavior
I'm able to reorder activities on the outline.

## Current Behavior
Reordering activities on the outline throws an error and fails (no change after page reloads).

kronicker:
This PR:
- fixes activities reorder on the latest schema config
- resolves #563

### DIFF
--- a/config/shared/activities.js
+++ b/config/shared/activities.js
@@ -11,7 +11,9 @@ const isString = require('lodash/isString');
 const map = require('lodash/map');
 const mergeConfig = require('../utils/mergeConfig');
 const parseSchemas = require('./schema-parser');
+const reduce = require('lodash/reduce');
 const union = require('lodash/union');
+const uniq = require('lodash/uniq');
 
 /* eslint-disable */
 const defaultConfiguration = require('./activities-rc');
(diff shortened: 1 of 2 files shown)


config/shared/activities.js
```diff
@@ -127,10 +128,16 @@ function getElementConfig(schemaId, type) {
 function getSiblingLevels(type) {
   const schemaId = getSchemaId(type);
   if (!schemaId) return [{ type }];
-  const levels = getOutlineLevels(schemaId);
-  const { level } = find(levels, { type }) || {};
-  if (!level) return [{ type }];
-  return filter(levels, { level });
+  const outline = getOutlineLevels(schemaId);
+  const activityConfig = find(outline, { type });
+  if (!activityConfig) return [{ type }];
```
kronicker, line 133: I carried this logic over from the previous implementation, but this case could only happen if the activity is either not in schema config (which I don't know should we even handle) or more likely if it's not an outline activity.
The latter case assumes only reorder between the same types of activities which is currently correct since containers are grouped per type but might be limiting in the future and also off if this would be used for non-type-grouped sub-containers in custom implementations...
Thoughts @ExtensionEngine/tailor ?

config/shared/activities.js
```diff
@@ -127,10 +128,16 @@ function getElementConfig(schemaId, type) {
...
+  const isRoot = activityConfig.rootLevel;
```
underscope, line 134: Consider simplifying to something like:
```javascript
const subLevels = filter(outline, it => it.subLevels.includes(type)).flat();
if (!activityConfig.rootLevel) return subLevels;
return filter(outline, it => it.rootLevel).concat(subLevels);
```

PR #566, opened 2020-07-29
Feature/editor sidebar activity tree updates

abasic:
- the search is now always visible
- the editable activities are marked with an icon
- can only navigate to them by clicking the icon
- removed dense prop

Video:

tailor_sidebar


client/components/editor/Sidebar/index.vue
```diff
@@ -91,6 +87,14 @@ export default {
     metadata() {
       const { repository, selectedElement } = this;
       return getElementMetadata(get(repository, 'schema'), selectedElement);
+    },
+    toolbarClasses() {
+      const { selectedElement } = this;
+      if (!selectedElement) return;
+
+      const classes = ['toolbar-visible'];
+      if (selectedElement.parent) classes.push('toolbar-composite');
```
abasic: I do not know when the `toolbar-composite` class is applied. By that, I mean which content elements trigger it. If anybody knows please, write in a comment.

underscope: Table (General support for composite element handling when you can have nested element and parent element toolbar visible at the same time)

abasic: I forgot about the table element 😄 . I'll test with it as well.

abasic:
@underscope 
tailor_activity_tree_hover

client/components/editor/Sidebar/Navigation.vue
```diff
@@ -1,6 +1,6 @@
 <template>
-  <div>
-    <v-sheet color="transparent" class="navigation-header">
+  <div class="navigation-container">
+    <v-sheet color="transparent" class="navigation-header" elevation="1" tile>
```
underscope: class => last

client/components/editor/Sidebar/Navigation.vue
```diff
@@ -11,13 +11,23 @@
         class="my-3 mx-1" />
     </v-sheet>
     <v-treeview
-      @update:active="navigateTo"
+      ref="activityTree"
```
underscope, line 16: Consider adding 'no matches found' message

client/components/editor/Sidebar/Navigation.vue
```diff
@@ -11,13 +11,23 @@
...
       :items="activityTree"
-      :active.sync="active"
+      :active="active"
       :search="search"
-      open-all activatable hoverable dense>
-      <template v-slot:label="{ item: { name, selectable } }">
-        <span :class="{ selectable }">{{ name }}</span>
+      open-all
+      hoverable>
```
underscope: Consider inlining booleans (like before)

client/components/editor/Sidebar/Navigation.vue
```diff
@@ -11,13 +11,23 @@
...
+      <template v-slot:label="{ item: { id, name, selectable } }">
+        <div @click.stop="navigateTo(id)" :class="{ selectable }">
+          {{ name }}
+          <v-btn
```
underscope: Consider replacing btn with icon only (The whole section is selectable so it felt weird to have an additional nested element with the same action and additional hover effect). Thoughts? 🤔

client/components/editor/Sidebar/index.vue
```diff
@@ -91,6 +87,14 @@ export default {
     metadata() {
       const { repository, selectedElement } = this;
       return getElementMetadata(get(repository, 'schema'), selectedElement);
+    },
+    toolbarClasses() {
```
underscope: Returning object like before seems simpler 🤔 ?

underscope:
@abasic What's the state of this one?

client/components/editor/Sidebar/Navigation.vue
abasic: The issue with an empty message is that the `v-treeview` component does not have native support for it. Does not have any API capabilities to notify the parent component so there is no clean way of knowing if the tree view is empty or not. One of the ways that I found to do it is to watch its computed property like so:
```js
mounted() {
  this.$watch(
    () => this.$refs.activityTree.excludedItems.size,
    val => (this.showEmptyMessage = this.totalSize === val)
  );
}
```
but I am not really a fan of watching properties of components such as these that are not exposed to the outside since they can change in minor versions even and break the app. If you are ok with this type of approach I will continue with this approach.

client/components/editor/Sidebar/Navigation.vue
abasic: It was a button because it was the only one clickable. Not the wrapper is. So I will remove that button but the `div` should be changed to a `button` since it is clickable. Do you agree?

client/components/editor/Sidebar/index.vue
abasic: I prefer keeping the template simple (not much logic in it) so that is why I moved the logic to the script part. But I will move it back to the template.

client/components/editor/Sidebar/Navigation.vue
abasic: Found another way

PR #567, opened 2020-07-29
Change filesystem file URL

lukasaric:
This PR removes hostname/origin prefix in file url.


borna223:
This won't work if API is hosted on different server?

#567 with `Content-Security-Policy: frame-ancestors 'self' ...` response header should do the trick.

underscope:
@vladimyr Resolution ?

PR #568, opened 2020-07-29
X-Frame-Options bugfix

borna223:
Removes `X-Frame-Options` header previously set by `helmet` for FS hosted assets.

Assets within `<frame>`, `<iframe>`, `<embed>` or `<object>` are now rendered correctly.


borna223:
Solved by #567

borna223:
Solved by #567

Issue #540, opened 2020-05-29
Add validation for PDF, Video and Audio component


kjuej:
## I'm submitting a...

- [x] Bug report  
- [ ] Regression (a behavior that used to work and stopped working in a new release)
- [ ] Feature request
- [ ] Documentation issue or request

## Context
I'm able to enter invalid URL links for the mentioned components
<img width="1287" alt="Screen Shot 2020-05-29 at 10 10 05" src="https://user-images.githubusercontent.com/34127147/83237081-e7897b00-a194-11ea-812f-cbefa1052c0a.png">

kjuej:
@lukasaric I should be able to add youtube videos to video component. It seems it's no longer possible due to the validation:
<img width="976" alt="Screen Shot 2020-10-09 at 12 14 41" src="https://user-images.githubusercontent.com/34127147/95571628-0a704980-0a29-11eb-8273-ea2e430681f9.png">

kjuej:
@lukasaric  I should be able to add youtube videos to video component. It seems it's no longer possible due to the validation:
<img width="976" alt="Screen Shot 2020-10-09 at 12 14 41" src="https://user-images.githubusercontent.com/34127147/95576454-c3865200-0a30-11eb-9f71-06ff66bfe4a7.png">

PR #569, opened 2020-07-29
Implement InputAsset element URL input validation 

lukasaric:
- This PR solves the #540 issue.
- It adds URL validation for PDF, AUDIO & VIDEO container elements


kjuej:
@lukasaric I should be able to add youtube videos to video component. It seems it's no longer possible due to the validation:
Screen Shot 2020-10-09 at 12 14 41

PR #570, opened 2020-07-30
Fix unused transaction in Activity.remove method

MiroDojkic:
Transaction in the `Activity.remove` method wasn't passed to its Sequelize calls which would cause them to not rollback in case of an error.

### DIFF
--- a/server/activity/activity.model.js
+++ b/server/activity/activity.model.js
@@ -182,7 +182,8 @@ class Activity extends Model {
 
   remove(options = {}) {
     if (!options.recursive) return this.destroy(options);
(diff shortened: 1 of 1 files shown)


server/activity/activity.model.js
```diff
@@ -229,9 +229,9 @@ class Activity extends Model {
   }
 }
 
-function removeAll(Model, where = {}, soft = false) {
+function removeAll(Model, where = {}, soft = false, transaction) {
```
vladimyr, line 232: Um, I'm entering nitpick mode right now but my OCD begs for joining soft and transaction into optional options arg. 🙃

PR #571, opened 2020-07-31
Support for creating revisions on afterBulkCreate

borna223:


kjuej:
@borna223 Could we differentiate between created and copied content? So that for the copied content, the history revision would say "copied..." instead of "created..."?

PR #576, opened 2020-08-12
Add reinvite button in user edit dialog

hrvojevu:
This PR extends already implemented `reinvite` functionality by adding `Reinvite` button in user edit dialog

### DIFF
--- a/client/components/system-settings/UserManagement/UserDialog.vue
+++ b/client/components/system-settings/UserManagement/UserDialog.vue
@@ -5,10 +5,10 @@
       <div v-if="userData" class="text-left">
         <v-btn
           @click="reinvite"
-          :loading="isLoading"
-          :disabled="isLoading"
-          outlined
+          :loading="isReinviting"
+          :disabled="isReinviting"
           color="primary"
+          outlined
           class="mb-6">
           Reinvite
         </v-btn>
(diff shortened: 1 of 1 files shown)


client/components/system-settings/UserManagement/UserDialog.vue
```diff
@@ -2,6 +2,17 @@
   <tailor-dialog v-model="show" header-icon="mdi-account">
     <template v-slot:header>{{ userData ? 'Edit' : 'Create' }} User</template>
     <template v-slot:body>
+      <div v-if="userData" class="text-left">
+        <v-btn
+          @click="reinvite"
+          :loading="isLoading"
+          :disabled="isLoading"
+          outlined
+          color="primary"
```
underscope, line 11: Attr order

client/components/system-settings/UserManagement/UserDialog.vue
```diff
@@ -2,6 +2,17 @@
   <tailor-dialog v-model="show" header-icon="mdi-account">
     <template v-slot:header>{{ userData ? 'Edit' : 'Create' }} User</template>
     <template v-slot:body>
+      <div v-if="userData" class="text-left">
+        <v-btn
+          @click="reinvite"
+          :loading="isLoading"
+          :disabled="isLoading"
```
underscope, line 9: We should probably rename this flag ?

kjuej:
@hrvojevu @underscope Would it make more sense to place a reinvite button in the actions column (mdi-refresh or something)?

hrvojevu:
@kjuej That is fine, but in that case we would need to add confirmation dialog for reinvite. Additionally, I'm not sure about the `mdi-refresh` icon since it doesn't imply reinvite and in that case we would have to add tooltips.

underscope:
@kjuej Since reinvite should be rare action lets keep it this way for now (don't want to pollute the user list). In terms of verbiage, we should consider moving from 'edit' to 'manage' user terminology.

PR #352, opened 2019-09-27
Use Vue CLI

zcuric:
### This PR
- [x] Introduces Vue CLI to Tailor

#### Motivation

Main motivation for this PR is to align Tailor (and in future all other EE apps) with official tools that are available for our technology set. Also big trigger for this was [this tweet pool on Poi twitter profile](https://twitter.com/poi__js/status/1173248429616578560), which obviously states that Vue community is gravitating to the official tools rather to third-party solutions. 

Although I think Poi handles some things better then Vue CLI and some things in Vue CLI are made rather sloppy, I also think that by using the official tools we guard ourselves from being attached to third-party which could end up unmaintained. 

#### @vladimyr checklist 🗒

- [x] #354 being merged first and required changes done here too
- [x] Collection of former `poi` - now `vue-cli` plugins to get extracted to separate repo in order to decrease maintenance burden. That means keeping both `poi` versions until we migrate all our apps and this newly created `vue-cli` variants.
- [x] All development functionalities (nodemon, debugging both inside and outside of vscode, webpack inspection scripts) need to be implemented using `vue-cli` too being fully transparent to end users
- [x] Build dependencies **only** get thoroughly inspected and cleaned as part of transition
- [x] c/p of build log showing built artefacts needs to be pasted here showing that there isn't significant difference in build output in terms of chunking, sizes and stuff

PR #577, opened 2020-08-12
Import Vuetify components via a-la-carte

MiroDojkic:


zcuric:
@MiroDojkic  @underscope @vladimyr if we merge #352 we can use `vue-cli-vuetify-plugin` that does this out of the box.

underscope:
@MiroDojkic @zcuric Closing then this one until the status of #352 gets resolved. @MiroDojkic Please reopen based on the resolution.

PR #579, opened 2020-08-15
Fix asset resolving

mandryllo:
This PR:
- [x] Refactors `isPrimitive` & `isQuestion` functions
- [x] Remove duplicated requests to s3


underscope:
@mandryllo Let's discuss

underscope:
@mandryllo Should we support assets on composite as well?

mandryllo:
> @mandryllo Should we support assets on composite as well?

We already support [it](https://github.com/ExtensionEngine/tailor/blob/98c261bb9bbece773245ab3386a1c989f7a2d000/server/shared/storage/helpers.js#L129) 👍

PR #581, opened 2020-08-17
Implement activity prepend

### LINKED ISSUE #522: As a User, I should be able to add activity above currently selected activity
## I'm submitting a...

- [ ] Bug report  
- [x] Feature request

## Context
At the moment Tailor only supports adding new activity within structure editor below currently selected activity.

image

Add the ability to add the activity above as-well.

hrvojevu:
As a part of issue https://github.com/ExtensionEngine/tailor/issues/522 this PR:
- updates `calculatePosition` to handle activity with added `prepend` argument
- updates `Menu` and `Toolbar` components to display prepend icon (arrow-up) and display label `Add item above`


client/utils/calculatePosition.js
```diff
@@ -1,18 +1,30 @@
-export default function ({ newPosition, items, isFirstChild, insert = false }) {
+export default function ({
```
underscope: Since we have more actions now we should consider changing signature and instead insert and prepend booleans have action enum with (ADD_BEFORE, ADD_AFTER, and REORDER values; export enum as well). Thoughts? (It seams like the code would be more readable as well).

client/components/repository/common/CreateDialog/index.vue
```diff
@@ -51,6 +51,7 @@ export default {
     levels: { type: Array, required: true },
     anchor: { type: Object, default: null },
     addChild: { type: Boolean, default: false },
+    prependItem: { type: Boolean, default: false },
```
underscope: We can use the action mentioned in the util here as well.

underscope (changes requested):

hrvojevu:
@kjuej please re-test this feature since it has been refactored.

kjuej:
@hrvojevu seems to be working as expected.

client/components/repository/common/ActivityOptions/Menu.vue
```diff
@@ -64,21 +65,26 @@ export default {
   data: () => ({
     showCreateDialog: false,
     showCopyDialog: false,
-    addSublevel: false,
+    addChild: false,
+    action: '',
```
underscope: default to null?

client/components/repository/common/ActivityOptions/Menu.vue
```diff
@@ -103,21 +109,27 @@ export default {
...
+      if (addChild) return 'Add into';
+      if (action === ADD_BEFORE) return 'Add above';
+      if (action === ADD_AFTER) return 'Add below';
     }
   },
   methods: {
     ...mapActions('repository/activities', ['remove']),
-    setCreateContext(levels, addSublevel) {
-      this.setSupportedLevels(levels, addSublevel);
+    setCreateContext(levels, { addChild = false, action = '' } = {}) {
```
underscope: Same as prev `action = ''` ⬆️   
Should we default to add below (and handle`INTO`; see below)?

client/components/repository/common/ActivityOptions/Menu.vue
```diff
@@ -103,21 +109,27 @@ export default {
...
+      this.setSupportedLevels(levels, addChild);
+      this.action = action;
```
underscope: Should ADD_INTO also be handled (so that action context is always properly set; right now action is sometimes used and sometimes not)

client/components/repository/common/ActivityOptions/Menu.vue
```diff
@@ -103,21 +109,27 @@ export default {
           action: () => this.delete(this.activity)
         }
       ];
+    },
+    createDialogHeading: ({ addChild, action }) => {
```
underscope: Create object and return obj[action];

client/components/repository/common/ActivityOptions/Toolbar.vue
```diff
@@ -68,17 +74,24 @@ export default {
   data: () => ({
     showCreateDialog: false,
     supportedLevels: [],
+    action: false,
     addChild: false
   }),
   computed: {
-    options: vm => getOptions(vm)
+    options: vm => getOptions(vm),
+    createDialogHeading: ({ addChild, action }) => {
```
underscope: Reuse obj from Menu component comment

underscope (changes requested):

client/components/repository/common/CreateDialog/index.vue
```diff
@@ -50,7 +52,7 @@ export default {
     repositoryId: { type: Number, required: true },
     levels: { type: Array, required: true },
     anchor: { type: Object, default: null },
-    addChild: { type: Boolean, default: false },
+    action: { type: String, default: '' },
```
underscope: Should be required?

client/components/repository/common/ActivityOptions/Toolbar.vue
```diff
@@ -68,17 +73,25 @@ export default {
...
     showCreateDialog: false,
     supportedLevels: [],
-    addChild: false
+    action: null
   }),
   computed: {
-    options: vm => getOptions(vm)
+    options: vm => getOptions(vm),
+    createDialogHeading: ({ action }) => {
+      const heading = {
```
underscope: Extract (new module within ActivityOptions folder) and reuse.

client/components/repository/common/CopyActivity/index.vue
```diff
@@ -68,13 +69,15 @@ import { SCHEMAS } from 'shared/activities';
...
+const { ADD_INTO } = insertActions;
+
 export default {
   name: 'copy-activity',
   props: {
     repositoryId: { type: Number, required: true },
     levels: { type: Array, required: true },
     anchor: { type: Object, default: null },
-    copyInto: { type: Boolean, default: false },
+    action: { type: String, default: '' },
```
underscope: Should be required.

client/utils/insertActions.js
```diff
@@ -0,0 +1,5 @@
+export default {
```
underscope: Consider renaming this to InsertLocation or InsertActivityLocation (for filename) and using it without destructuring i.e. InsertLocation.BEFORE. Since this is an enum like structure consider renaming the file to PascalCase.

client/utils/calculatePosition.js
```diff
@@ -1,23 +1,30 @@
-export default function ({ newPosition, items, isFirstChild, insert = false }) {
+import insertActions from './insertActions';
+
+const { ADD_AFTER, ADD_BEFORE } = insertActions;
+
+export default function ({ newPosition, items, isFirstChild, action = '' }) {
```
underscope: Should action default to value (or throw err if not provided)

client/utils/calculatePosition.js
```diff
@@ -1,23 +1,30 @@
-export default function ({ newPosition, items, isFirstChild, insert = false }) {
+import insertActions from './insertActions';
```
underscope: While you are in here, can you please document each scenario and rules being made

underscope (commented):
Nice! almost there 👏🏻

client/utils/insertActions.js
hrvojevu: `activityInsertLocation` seems like best option for me but I'm not sure about not destructuring since in that case we would end up with huge if statement e.g. `action === activityInsertLocation.BEFORE`. Also, with this rename it would make sense to rename `action` prop to `location` or `activityLocation`?

client/utils/calculatePosition.js
hrvojevu: `action` should not be defaulted to a value (or defaulted to `REORDER`, which would then need be added to `insertActions`) since this function can be used to calculate item position when reordering, therefor we should not throw error if not provided. Since not providing `action` argument yields the same result as defaulting it to `REORDER` I don't see the point in defaulting it to `REORDER`.

client/utils/calculatePosition.js
hrvojevu: Added `js-doc` for the function. IMO `js-doc` is sufficient in this case. If you disagree, I will add comments for `if` blocks, if that is what you asked in the original comment.

client/components/repository/common/ActivityOptions/Toolbar.vue
```diff
@@ -23,29 +23,34 @@
       :repository-id="activity.repositoryId"
       :levels="supportedLevels"
       :anchor="activity"
-      :add-child="addChild"
-      :heading="`
-        Add ${addChild ? 'into' : 'below'}
-        ${activity.data.name}`" />
+      :action="action"
+      :heading="`${createDialogHeading} ${activity.data.name}`" />
```
underscope: Computed name sounds like a method (I understand that here this is litteraly dialogHeading upon create). Can we just rename it to dialogHeading or find better alternative?

client/utils/calculatePosition.js
```diff
@@ -1,23 +1,42 @@
...
+/**
+ * Calculates item position based on the options provided.
+ * @param {number} newPosition The position of the anchor item by which the new
+ * position will be calculated.
+ * @param {Object[]} items An array of objects in which we are determining
+ * item's new position.
+ * @param {boolean} isFirstChild Boolean value denoting wether the item should
+ * be placed as the first child of its parent.
+ * @param {string} [action=null] A string value determining where the item should
+ * be placed in relation to `newPosition`. Defaults to `null` in case of items reorder.
```
underscope: I would rather add REORDER as a type and default to it for readability (like you mentioned in the previous comment) and consistency sake (it helps to see that REORDER is default within signature and I think it makes sense that all action types are documented / and have definition the same way. Please let me know if I missed something / if it doesn't make sense.

client/utils/calculatePosition.js
underscope: The thing I would like to do then would be to use action type after ADD_AFTER cond (for else) to improve readability of when that rule is applied, if that makes sense. Also, insertActions should be renamed to InsertActions or similar and have type def (and action string type replaced with that type)

client/utils/insertActions.js
underscope: Enum likes should be PascalCase in the codebase ;) That's fine, you can rename it upon import within context to InsertLocation or similar ;) I think when something is extracted to a bit more general section it should have a bit more detailed definition. I think action is still better because of reorder? Thoughts?

underscope:
Please merge with the latest develop  aswell

client/utils/calculatePosition.js
```diff
@@ -1,23 +1,42 @@
...
+
+const { ADD_AFTER, ADD_BEFORE, REORDER } = InsertLocation;
+
+/**
+ * Calculates item position based on the options provided.
+ * @param {number} newPosition The position of the anchor item by which the new
+ * position will be calculated.
+ * @param {Object[]} items An array of objects in which we are determining
+ * item's new position.
+ * @param {boolean} isFirstChild Boolean value denoting wether the item should
```
underscope: Should we just remove it then and use ADD_INTO instead?

hrvojevu: `ADD_INTO` and `isFirstChild` are contextually different. For example: action can be `REORDER` and `isFirstChild` can be `true` in case the item is reordered and is placed in the first position. Same example goes for `ADD_BEFORE` action and `isFirstChild===true`, this means item is added before other item on the first position. For that reason we _need_ two arguments.

underscope: - Makes sense 👍 My impression because of the comment parent-child context was that it is used only for that scenario. 
- Misspelling of wether
- Please reformat param comments in google jsdoc style (https://google.github.io/styleguide/javascriptguide.xml?showone=Comments#Comments)

Thanks!

PR #590, opened 2020-08-27
Use plyrue instead vue-aplayer

### LINKED ISSUE #387: Unable to gradually increase/decrease volume on an audio element
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request

## Context
I can only mute/unmute the audio element. The option to edit volume is shown but i can not click on it.

Screen Shot 2019-11-15 at 14 01 39

zcuric:
### This PR
- [x] Removes vue-aplayer due not being updated for several years now and having bugs with volume changing with `plyrue` which we already have in the code base
- [x] Closes #387 

### NOTE
- If we don't care about author picture (from the code, I see we don't) we are fine.


kjuej:
@zcuric still unable to edit audio when i upload an mp3 link:
Screen Shot 2020-08-27 at 15 45 54

zcuric:
I did manage to solve the bug regarding vue-aplayer volume control, but when you add new audio element in the content it breaks all audio elements volume control.  :/

kjuej:
@zcuric Volume is working as expected now. Would it be possible to center the audio element inside its container 🤔 :
Screen Shot 2020-08-28 at 12 24 51

kjuej:
@zcuric If i add multiple audio elements, change volume on each of them, after page refresh all of them will have the same volume.

zcuric:
@zcuric it makes sense, because volume is reset after page refresh. 😅

zcuric:
@kjuej I did an investigation and it seems to be plyr.js related issue. I don't know is that an issue in the first place, but it's definitely a weird behaviour. @vladimyr

underscope:
Can we extract this outside of tailor? (the same way that @abasic did for tce-file)

PR #586, opened 2020-08-25
Improve email UI

hrvojevu:
As a part one of issue https://github.com/ExtensionEngine/tailor/issues/517 this PR:
- improves email UI by:
	- updating email logo and header
	- updating button color to `#37474F`
	- changing link color to `#C2185B`
	- removing fixed width on buttons
	- setting divider to `2px` width and changing style to`dashed`
	- updating text alignments

Screen Shot 2020-08-25 at 12 07 18 PM
Screen Shot 2020-08-25 at 12 08 20 PM

#### Note:
Discussion email UI will be addressed in separate PR.

Issue #517, opened 2020-05-15
Improve email templates


underscope:
## I'm submitting a...

- [ ] Bug report  
- [ ] Regression (a behavior that used to work and stopped working in a new release)
- [x] Feature request
- [ ] Documentation issue or request

## Context
- Add logo
- Improve discussion template with context (show last N messages)
- Tweak overall UI

hrvojevu:
@underscope here are some suggestions on improved UI. These include changes on the button (width is now `auto`, color is slightly darker to match logo), divider border style changed to `dashed` (because tailoring). To achieve consistency with logo's across the platform (round logo) new svg file was created since, afaik, it is not possible to have completely rounded container (`border-radius: 50%`) in `mjml`. Suggestions are welcome! 

As far as:
> Improve discussion template with context (show last N messages)

Can you elaborate on that? How many messages do we want to show?


#### Example 1

![Screen Shot 2020-08-21 at 1 12 47 PM](https://user-images.githubusercontent.com/7149531/90884472-1c7e2600-e3b0-11ea-8dc5-80e81d2a2728.png)

#### Example 2

![Screen Shot 2020-08-21 at 1 12 35 PM](https://user-images.githubusercontent.com/7149531/90884480-1ee08000-e3b0-11ea-8267-582d4c8b0ee5.png)

#### Example 3

![Screen Shot 2020-08-21 at 1 13 09 PM](https://user-images.githubusercontent.com/7149531/90884493-22740700-e3b0-11ea-8fbd-868c2078be77.png)

#### Example 4

![Screen Shot 2020-08-21 at 1 10 17 PM](https://user-images.githubusercontent.com/7149531/90884284-c0b39d00-e3af-11ea-9b12-2cd0e91bc481.png)

underscope:
👌 Options 1 and 2 look good to me. Consider just using scissors for the logo (omit circle/background) in primary color and using secondary to accent some details (i.e. name; might be a bit much?)

hrvojevu:
@underscope since accent color (`#37474f`) doesn't really differ from default text color, I would leave it as is. In the image below you can see default (current) text color for name on left side and accent color (`#37474f`) on the right. 

![comparison](https://user-images.githubusercontent.com/7149531/91036821-4893e400-e608-11ea-9e2a-5feb79834c74.png)

hrvojevu:
@underscope regarding 

> Improve discussion template with context (show last N messages)

how many discussion messages do we want to display?

[...]

PR #591, opened 2020-08-31
Improve discussion notification email UI

hrvojevu:
As a part 2 of issue https://github.com/ExtensionEngine/tailor/issues/517, this PR:
- removes activity url from discussion notification email
- displays last 3 discussion messages in addition to the last one
- updates discussion email styling
- updates email `renderHtml` function since `mjml2html` messes up `mustache` `{{#}}, {{/}}, {{^}}` tags

Screen Shot 2020-08-31 at 11 10 08 AM

### Note:
Base branch must be changed to `develop` after https://github.com/ExtensionEngine/tailor/pull/586 has been merged in


kjuej:
@hrvojevu It seems if there's less than 5 comments in total, only the last comment will be shown in the notification email. Is this intended?
Screen Shot 2020-09-01 at 11 44 12

kjuej:
@hrvojevu @underscope @vladimyr Seems like the issue above only occurs on gmail. Also the logo is missing on gmail as well.

hrvojevu:
This is the link that gmail produces:

`https://ci4.googleusercontent.com/proxy/mSnVz8PRZQyuYaDvkJpLIHXatki_PiJia-qK2f8d9Q-OxgF1V7kp4P9-KD9enBmliJXhgiKa0XWZAXlS053N1NIa2i5_-C9UxRnxVw=s0-d-e1-ft#http://127.0.0.1:3000/assets/img/default-logo-compact.svg`

I suppose this doesn't work due to localhost and would work in production environment.

server/shared/mail/templates/components/head.mjml
```diff
@@ -20,10 +20,12 @@
       font-size="12px"
       background-color="white" />
     <mj-class name="divider"
-      padding="20px 25px"
+      padding="20px 10px"
       border-width="1px"
       border-color="#d2d2d2"
       border-style="dashed" />
+    <mj-class name="avatar"
+      padding="10px" />
```
underscope: Consider inlining (only 2 attrs)

server/comment/hooks.js
```diff
@@ -39,13 +39,19 @@ async function sendEmailNotification(comment, db) {
     ]
   });
   const { author, repository, activity } = comment;
+  const lastFourComments = await activity.getComments({
```
underscope: Define this more generic like bellow (`previousComments`). Use offset to eliminate 1.

server/shared/mail/templates/comment.mjml
```diff
@@ -3,33 +3,52 @@
...
+            </td>
+            <td style="padding: 0 0 0 20px">
+              <div class="label">{{author.label}}</div>
+              <div>{{content}}</div>
+            </td>
+          </tr>
+        </mj-table>
+       </mj-column>
+    </mj-section>
+    {{#previousComments}}
```
underscope, line 23: Consider extracting into separate component.

server/shared/mail/templates/comment.mjml
```diff
@@ -3,33 +3,52 @@
...
-          font-size="16px">
-          {{content}}
+          <b>{{author.label}}</b> left a comment on <b>{{repositoryName}}</b>, <b>{{topic}}</b>.
         </mj-text>
       </mj-column>
     </mj-section>
     <mj-section padding="0 10px">
-      <mj-column border-top="1px solid #ccc">
+			<mj-column border-bottom="2px dashed #d2d2d2">
+        <mj-table padding="10px">
```
underscope: See if there is a better way of implementing this layout

server/shared/mail/templates/comment.mjml
hrvojevu: I don't think it is possible due to problems described in the render function. If we were to import separate component here, `mustache` render would not catch that component and `mjml2html ` would mess up mustache tags, second mustache render would not be able to render that component properly.

server/shared/mail/templates/comment.mjml
```diff
@@ -3,33 +3,52 @@
...
+          <tr class="content" style="display: flex;">
+            <td>
+              <img class="avatar" src="{{author.imgUrl}}" />
```
underscope: Data URLs (if image is set) won't work in Gmail (https://stackoverflow.com/questions/8580355/can-you-send-images-in-datauri-format-to-gmail)

hrvojevu: Data URLs are not used in this instance, but a URL pointing to localhost. I figure this should work when these images are hosted on publicly available domain.

server/shared/mail/templates/comment.mjml
hrvojevu: Initially I have tried using the common mjml tags (mj-section, mj-column) but afaik it is not possible to achieve this exact layout.

PR #603, opened 2020-10-01
Rewrite `calculateInsertPosition` getter as action

### LINKED ISSUE #596: Item elements not sorted as they should be(when using add element below)
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request

## Context
When using "add element below" feature, the element first gets positioned at the very bottom, then after page refresh gets positioned at the top.
CopyBelowTailor

hrvojevu:
This PR:
- rewrites `calculateInsertPosition` getter as action thus resolving issue https://github.com/ExtensionEngine/tailor/issues/596


kjuej:
@hrvojevu the copy element below functionality still doesn't work. Initially it copies the element above, and then if you refresh the page the copied elements seem to be sorted at random.

PR #604, opened 2020-10-07
Dissalow non-integer keys on exam time input field

### LINKED ISSUE #592: Improve validation for exam time limit
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request

## Context
Exam time limit can have multiple "-","+","e" characters.
Screen Shot 2020-09-01 at 10 52 30

hrvojevu:
This PR:

- adds `keydown` event listener which prevents default input behavior if pressed key is either: `e`, `+` or `-`

This issue occurs due to `e`, `-` and `+` are being interpreted as numbers when used in `number` type input and dispatched as an empty value. Vee-validate doesn't handle this particular use case, more info here: https://github.com/logaretm/vee-validate/issues/1667.


kjuej:
@hrvojevu Maybe the same should be done for punctuation sign?
Screen Shot 2020-10-07 at 12 10 34

client/components/content-containers/tcc-exam/edit/AssessmentGroup.vue
```diff
@@ -120,6 +121,10 @@ export default {
       const ids = assessments.map(it => it.cid);
       const cond = it => !ids.includes(it.cid);
       this.unsavedAssessments = pickBy(this.unsavedAssessments, cond);
+    },
+    disallowNonIntegers(event) {
```
underscope: Would !isInteger && prevent do the trick (to avoid handling on code key level; can be inlined as well)?

PR #624, opened 2020-10-12
Element copying feature

### LINKED ISSUE #45: As a User I should be able to copy element from other activity
## I'm submitting a...

- [ ] Bug report  
- [x] Feature request

## Expected Behavior
Upon clicking on add element button User should be able to navigate through any repo structure and copy existing element.

### LINKED ISSUE #628: Element link improvements
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request
- [x] Improvement

## Improvement

After some time using the Element link feature, there are few improvements to suggest
1. Make sticky footer in Select Element dialog
image

2. When selecting an element inside the Select Element dialog, it would be awesome if there were an indication of the activity types, use existing color coding for activities from the repository.
image

3. Content authors are getting confused when a link has already been added and think they can add more links. Remove the edit icon from the Element sidebar and put the eye icon which will present that the user can only view the selected link and if they need to edit it or remove, they can do it by clicking the x icon.
image

## Bug report

1. The order of elements inside the Select Element dialog is not correct, not as inside the related Activity
image

mcapeta:
This PR enables selecting and copying existing content element (resolves #45) to the current activity.
It alse fixes #628 - element order issue within element selection dialog.


client/components/common/SelectElement/index.vue
```diff
@@ -2,9 +2,26 @@
   <tailor-dialog
     @click:outside="close"
     :value="true"
-    header-icon="mdi-transit-connection-variant"
+    :header-icon="headerIcon"
+    scrollable
```
underscope: Order (after width)

client/components/common/SelectElement/index.vue
```diff
@@ -2,9 +2,26 @@
...
     width="650">
-    <template v-slot:header>{{ heading }}</template>
+    <template v-slot:header>
+      {{ heading }}
+      <v-btn
```
underscope: Add slot instead and expose toggleSelectAll and allElementsSelected (creates error for element link as well).

client/components/common/SelectElement/index.vue
```diff
@@ -2,9 +2,26 @@
...
+        v-if="selectedActivity && !loadingContent"
+        @click="toggleSelectAll"
+        text dark
+        class="float-right">
```
underscope: Or spacer / justify between

client/components/common/SelectElement/index.vue
```diff
@@ -2,9 +2,26 @@
...
+        <template v-if="allElementsSelected">
+          <v-icon class="mr-2">mdi-checkbox-multiple-blank-outline</v-icon>
```
underscope: Instead of repeating the same block consider creating a computed returning icon and label based on state.

client/components/common/SelectElement/index.vue
```diff
@@ -51,15 +69,24 @@ export default {
     selected: { type: Array, default: () => [] },
     heading: { type: String, required: true },
     multiple: { type: Boolean, default: true },
-    allowedTypes: { type: Array, default: () => [] }
+    allowedTypes: { type: Array, default: () => [] },
+    submitLabel: { type: String, default: 'Save' },
+    headerIcon: { type: String, default: null }
```
underscope: Default to mdi-plus or brick representing the element add?

client/components/common/tce-core/AddElement/index.vue
```diff
@@ -17,32 +17,42 @@
         <v-icon>{{ icon }}</v-icon>
       </v-btn>
     </slot>
-    <v-bottom-sheet v-model="isVisible" max-width="1240" inset>
+    <v-bottom-sheet
+      @input="isVisible = $event"
+      :value="isVisible && !showElementBrowser"
+      max-width="1240" inset>
```
underscope: Move inset to the next line (inline only booleans)

client/components/common/tce-core/AddElement/index.vue
```diff
@@ -17,32 +17,42 @@
...
-            <v-btn :value="50" icon>
-              <v-icon>mdi-select-compare</v-icon>
-            </v-btn>
-          </v-btn-toggle>
-          <v-divider class="mr-3" vertical />
-          <div class="width-label px-1 subtitle-1 grey--text text--lighten-4">
-            Element width
-            <span class="px-1">{{ elementWidth }}</span>%
-          </div>
+          <v-btn
```
underscope: Does it make sense to create a slot and control add element visibility from outside (Move this component into AddNewElement and have a new index.vue wrapper that does add new / add existing handling)? ` :value="isVisible && !showElementBrowser"` looks a bit off 🤔

client/components/common/tce-core/AddElement/index.vue
```diff
@@ -137,33 +167,41 @@ export default {
...
-        const data = { width: LAYOUT.FULL_WIDTH };
-        const question = [{ id: uuid(), data, type: 'JODIT_HTML', embedded: true }];
-        element.data = { ...element.data, question, type: subtype };
-      }
-      element.data = { ...element.data, ...initState() };
-      if (element.type === 'REFLECTION') delete element.data.correct;
-      this.$emit('add', element);
+    addElements(elements) {
+      const items = elements.map((it, index) => {
+        const position = this.position + index;
```
underscope: if inserting multiple items between elements it needs to be bellow max as well ;)

client/components/common/SelectElement/index.vue
mcapeta: Seems that creating a slot just moves the problem to the consumers and requires repeatable code in them (for similar behaviour among them, of course). Instead, I've managed to handle toggling ability inside the component by checking whether it allows selecting multiple items and the content is loaded or not. Let me know if you see any drawbacks in this approach 😉.

client/components/common/tce-core/AddElement/index.vue
mcapeta: Reworked `calculatePosition` util with `AddElement`'s component logic, there were certain problems with inserting multiple elements I wasn't aware of 👀 .

client/components/common/tce-core/AddElement/index.vue
mcapeta: Craves for some sort of separation 😄 , but I am kind of running in circles trying to clean this up... Have some ideas but I would like to resolve that in a separate PR. For now, I just moved some template and styles to `AddNewElement` component.

kjuej:
@mcapeta @underscope 
Minor note: I'd like the copy more button to be more noticeable. This way it seems like more of a headline.
Screen Shot 2020-11-12 at 11 58 06

kjuej:
@mcapeta seems like i can only copy elements from 1 repo, is this by design?

kjuej:
@mcapeta don't show deselect/select button if there are no elements to copy.
Screen Shot 2020-11-12 at 15 05 21

kjuej:
@mcapeta unable to add an element in a custom container:
Screen Shot 2020-11-12 at 15 10 47

underscope:
@kjuej Per conversation with Marin, he will proceed by adding the ability to select repo

bubafinder:
@mcapeta 
Disable editing elements while previewing/copying 

image

kjuej:
@mcapeta If a repo is empty (has no children), you're unable to enter it when in "copy element" menu:
CopyElement
Maybe a message should be added stating that the repo is empty once you enter it?

bubafinder:
Order of elements inside activity is not preserved 

image

bubafinder:
If I try to play the video, the element is selected. If some element has interaction (modal, video, audio, ...) don't select them on interaction click.

image

bubafinder:
Use default color for selected elements. Also, I don't prefer different right border color in relation to top, left and bottom color.

In this PR
image

Color that we use in activity
image

bubafinder:
Expected behavior: If the select repository input field is cleared, I expect to see a list of all repositories.
Current behavior: If the select repository input field is cleared, I can see only the current repository in the list, and need to input a few chars to filter repositories.

image

bubafinder:
Don't like this view. Maybe it is better to disable entering into the empty activity.
Also, it would be good to use min-height for the Copy elements modal.

image

bubafinder:
Cant remove selected style of interacted elements with deselecting & mouse focus out.

image

markolovric:
maybe to add a color to "All selected" state?

image

mcapeta:
> Expected behavior: If the select repository input field is cleared, I expect to see a list of all repositories.
> Current behavior: If the select repository input field is cleared, I can see only the current repository in the list, and need to input a few chars to filter repositories.

It is desirable to avoid fetching all available repositories at once for performance reasons.

>Don't like this view. Maybe it is better to disable entering into the empty activity.
Also, it would be good to use min-height for the Copy elements modal.

Disabling user to enter an empty activity could also be performance issue since we would need to fetch entire repository content to filter out empty activities.

Although your suggestions lead to better experience, I think we could rely on the fact that content authors should know where they copy the content from and not implementing your proposals would not affect them too much (open for discussion 😄 ).

bubafinder:
Everything you mentioned makes sense and I think it is more important than my suggestions.

bubafinder:
I managed to run the delete action in the Video Playlist element

image

bubafinder:
Can't see the border-bottom on the last element. Add a small margin-bottom to the last element.

image

bubafinder:
TODO: 
- test add into/bellow/above
- test drag&drop

mcapeta:
> Can't see the border-bottom on the last element. Add a small margin-bottom to the last element.

Cannot reproduce 🤷‍♂️ 
Which browser do you use (this scrollbar looks strange)?

> I managed to run the delete action in the Video Playlist element

This has to be handled within element itself - video playlist is not part of the Tailor codebase, it is custom element and will be improved together with other elements (AFAIK element preview/edit mode are going to be reworked soon).

bubafinder:
> Which browser do you use (this scrollbar looks strange)?

I am using Chrome. In system preferences, I have activated the option to have always visible scroll bar (this is a must for QA, to see the worst case)

> This has to be handled within element itself - video playlist is not part of the Tailor codebase, it is custom element and will be improved together with other elements (AFAIK element preview/edit mode are going to be reworked soon).

👍

bubafinder:
Keep mouse click style only above the "VIEW ELEMENTS" button

image

bubafinder:
The content inside the view is not responsive

image

client/components/common/ContentPreview/Element.vue
```diff
@@ -1,19 +1,40 @@
 <template>
-  <div @click="toggleSelection" class="element-preview-container">
+  <div
+    @click="toggleSelection"
+    :class="['element-preview-container', elementWidth]">
```
abasic: In Tailor the classes that are not dynamic are placed in the plain `class` attribute. Please change it here and everywhere else it was used.

client/components/common/ContentPreview/Element.vue
```diff
@@ -40,17 +62,32 @@ export default {
 .element-preview-container {
   display: flex;
   position: relative;
+  margin: 1rem 0;
+  float: none;
```
abasic: Why is this `float` definition necessary?

client/components/common/SelectElement/SelectRepository.vue
```diff
@@ -0,0 +1,48 @@
...
+    repositories: [],
+    loading: false
+  }),
+  methods: {
+    select(repository) {
+      if (this.repositories.includes(repository)) {
+        this.$emit('selected', repository);
+      }
+    },
+    fetchRepositories: throttle(loader(function (search) {
```
abasic: Is there a reason for using `throttle` instead of `debounce`?

client/components/common/SelectElement/SelectRepository.vue
```diff
@@ -0,0 +1,48 @@
...
+      if (!search || search.length < SEARCH_MIN_LENGTH) return;
+      return repositoryApi.getRepositories({ search }).then(repos => {
+        this.repositories = uniqBy([...this.repositories, ...repos], 'id');
```
abasic: If this is searching based on the specific search parameter, why is it not using just the new ones that have come in based on the search instead of keeping the old ones as well?

client/components/common/SelectElement/SelectRepository.vue
```diff
@@ -0,0 +1,48 @@
...
+    loading: false
+  }),
+  methods: {
+    select(repository) {
+      if (this.repositories.includes(repository)) {
+        this.$emit('selected', repository);
+      }
+    },
+    fetchRepositories: throttle(loader(function (search) {
+      if (!search || search.length < SEARCH_MIN_LENGTH) return;
```
abasic: Should the repositories list be cleared if no search is given or the search can not be run?

client/components/common/SelectElement/SelectRepository.vue
```diff
@@ -0,0 +1,48 @@
...
+  props: {
+    repository: { type: Object, default: null }
+  },
+  data: vm => ({
+    repositories: [],
+    loading: false
+  }),
+  methods: {
+    select(repository) {
+      if (this.repositories.includes(repository)) {
```
abasic: Just a small note on checking objects in arrays in this manner. I assume this works but it would not in the case the `v-combobox` did not return the object that was passed but another one it created. It is usually better to check if objects exist by relying on unique data it contains.

client/components/common/SelectElement/index.vue
```diff
@@ -93,11 +185,23 @@ export default {
     },
     close() {
       this.$emit('close');
+    },
+    openElement(elementId) {
+      const {
+        selectedActivity: { id: activityId },
+        repository: { id: repositoryId }
+      } = this;
+      const params = { activityId, repositoryId };
```
abasic: This might just be me but I think this is not really readable. Lately, I have been personally avoiding such deep destructing just because it can be hard to read and track. Even more when there are renaming's. 
I think the below is more readable and even less lines of code:
```js
const params = {
  activityId: this.selectedActivity.id,
  repositoryId: this.repository.id
};
```

client/components/common/tce-core/AddElement/AddNewElement.vue
```diff
@@ -0,0 +1,137 @@
...
+        <div class="group-heading blue-grey--text text--darken-4">
+          <span>{{ group.name }}</span>
+        </div>
+        <div class="group-elements">
+          <button
+            v-for="element in group.elements"
+            :key="element.position"
+            @click.stop="$emit('add', [element])"
+            :disabled="!isAllowed(element.type)"
+            class="element">
```
abasic: It is not apparent from the class name that this is a button that can be interacted with. I recommend changing to `add-element` or something along those lines.

client/components/common/tce-core/AddElement/AddNewElement.vue
```diff
@@ -0,0 +1,137 @@
...
+            <v-icon v-if="element.ui.icon">{{ element.ui.icon }}</v-icon>
+            <h5 class="body-2">{{ element.name }}</h5>
```
abasic: An `h5` tag should not be inside of a button. It is not semantically correct. The thing to use in this case is `span` (and in most other cases 😄).

client/components/common/tce-core/AddElement/AddNewElement.vue
```diff
@@ -0,0 +1,137 @@
...
+        <slot name="header"></slot>
+      </v-toolbar>
+      <div
+        v-for="group in library"
+        :key="group.name">
+        <div class="group-heading blue-grey--text text--darken-4">
+          <span>{{ group.name }}</span>
+        </div>
+        <div class="group-elements">
+          <button
```
abasic: Is there a reason not to use the `v-btn` component here?

client/components/common/tce-core/AddElement/index.vue
```diff
@@ -17,63 +17,70 @@
...
-              <h5 class="body-2">{{ element.name }}</h5>
-            </button>
-          </div>
-        </div>
-      </div>
-    </v-bottom-sheet>
+          <v-btn
+            @click="showElementBrowser = !showElementBrowser"
+            dark text
+            class="">
```
abasic: Please remove this empty attribute.

client/components/common/tce-core/AddElement/index.vue
```diff
@@ -17,63 +17,70 @@
...
+import flatMap from 'lodash/flatMap';
+import { getPositions } from 'utils/calculatePosition';
+import intersection from 'lodash/intersection';
 import { isQuestion } from '../utils';
+import pick from 'lodash/pick';
 import reduce from 'lodash/reduce';
+import reject from 'lodash/reject';
 import uuid from '@/utils/uuid';
 
+const SelectElement = () => import('components/common/SelectElement');
```
abasic, line 79: @ExtensionEngine/tailor Should these be defined here at the beginning, along with the other ones or directly in the `components` prop of the component? I am raising this since it is the first time I see this in Tailor and am wondering what is considered the better approach?

client/components/common/tce-core/AddElement/index.vue
```diff
@@ -137,33 +152,42 @@ export default {
...
     processedWidth() {
       return this.elementWidth === 50 ? LAYOUT.HALF_WIDTH : LAYOUT.FULL_WIDTH;
+    },
+    allowedTypes() {
+      const { elementWidth, include, layout, library } = this;
+      const elements = flatMap(library, 'elements');
+      if (!layout) return include || [];
+      const allowedElements = elementWidth === DEFAULT_ELEMENT_WIDTH
+        ? elements
+        : reject(elements, 'ui.forceFullWidth');
```
abasic, line 159: Why is this necessary?

client/components/common/tce-core/AddElement/index.vue
```diff
@@ -174,106 +198,14 @@ export default {
...
   watch: {
-    show(val) {
-      if (val) this.isVisible = val;
-    },
     isVisible(val, oldVal) {
       if (!val && oldVal) this.onHidden();
+    },
+    show(val) {
+      if (val) return this.showElementPicker();
+      this.onHidden();
```
abasic:
```suggestion
      return val ? this.showElementPicker() : this.onHidden();
```

client/components/common/tce-core/AddElement/index.vue
```diff
@@ -137,33 +152,42 @@ export default {
...
+      });
+      this.$emit('add', items);
       this.isVisible = false;
     },
-    isElementDisabled(element) {
-      if (this.elementWidth === DEFAULT_ELEMENT_WIDTH) return false;
-      return get(element, 'ui.forceFullWidth', false);
+    buildElement(el) {
+      const { processedWidth: width, activity } = this;
+      const { position, subtype, data = {}, initState = () => ({}) } = el;
```
abasic, line 175: @ExtensionEngine/tailor I think we should make sure that `initState` and `data` have a default value so that it is is not necessary to default here.

client/components/common/SelectElement/index.vue
```diff
@@ -2,88 +2,180 @@
...
+    async selectRepository(repository) {
+      const { currentActivities, currentRepository } = this;
+      this.repository = repository;
+      this.deselectActivity();
+      this.activities = currentRepository.id === repository.id
+        ? currentActivities
+        : await this.fetchActivities(repository);
+    },
+    fetchActivities: loader(function (repository) {
+      return repositoryApi.get(repository.id, { withActivities: true })
```
abasic: There exists an[ activity list route](https://github.com/ExtensionEngine/tailor/blob/develop/server/activity/index.js#L15) in Tailor that already fetches them based on the repository. Is there a reason why that was not used?

client/utils/calculatePosition.js
```diff
@@ -1,47 +1,48 @@
...
-const { ADD_AFTER, ADD_BEFORE, ADD_INTO, REORDER } = InsertLocation;
+const { ADD_AFTER, REORDER } = InsertLocation;
+
+const distributePositions = ({ lower = 0, upper }, count) => {
+  const delta = upper ? (upper - lower) / (count + 1) : 1;
+  return Array.from({ length: count }).map((_, i) => delta * (i + 1) + lower);
+};
+
+function getDeprecationWarning(config) {
+  if (!Object.prototype.hasOwnProperty.call(config, 'isFirstChild')) return;
```
abasic, line 11: Is there a reason why this is not written by just calling `hasOwnProperty` on the `config` object like so:
```js
if (!config.hasOwnProperty('isFirstChild')) return;
```

client/components/common/SelectElement/index.vue
```diff
@@ -2,88 +2,180 @@
...
-      const containers = sortBy(getContainers(activities, activity), 'position');
+      const { containers } = this;
       const elements = await this.fetchElements(containers);
-      this.contentContainers = containers.map(container => ({
-        ...container,
-        elements: elements.filter(element => element.activityId === container.id)
-      }));
+      this.contentContainers = containers.map(container => {
+        const containerElements = elements.filter(it => it.activityId === container.id);
+        return { ...containers, elements: sortBy(containerElements, 'position') };
```
abasic: I think it should be `container` instead of `containers`.

client/components/common/SelectElement/index.vue
```diff
@@ -2,88 +2,180 @@
...
+      });
+    },
+    assignActivity(element) {
```
abasic: This is used when the element is being toggled on or off which can theoretically happen frequently. Is there a reason this can not be called on all of the elements right after they are fetched and ordered by position?

abasic (changes requested):

client/components/common/ContentPreview/Element.vue
mcapeta: Along with this one, there is also an `elementWidth` class binding which sets the `float: left;` to this element. I've put it here to annul that property (since it results in 0 height parent).

client/components/common/SelectElement/SelectRepository.vue
mcapeta: Thought we want function invocation on the leading edge 🤷‍♂️ ... Updating to be debounced...

client/components/common/SelectElement/SelectRepository.vue
mcapeta: Thought it may be nice to keep them all since they're already fetched. Will update to reject them, though 😉 .

client/components/common/SelectElement/SelectRepository.vue
mcapeta: Same as above, it seemed nice to keep existing ones. Updating...

client/components/common/SelectElement/SelectRepository.vue
mcapeta: 👍  Thanks for noting that, I was relying on the fact that `v-combobox` works with references when `return-object` prop is set to `true`. I'll update it to check element existence using `id` as you suggested just to make sure we're not affected by any possible underlying changes 💪 .

client/components/common/tce-core/AddElement/AddNewElement.vue
mcapeta: Nice catch! I missed that whole section obviously - copy-pasting is dev's enemy 🤦‍♂️

client/components/common/tce-core/AddElement/AddNewElement.vue
mcapeta: ^^ Copy pasted, to my shame 😞 .

client/components/common/tce-core/AddElement/index.vue
mcapeta: Sorry for wasting your time on this one 😱 .

client/components/common/tce-core/AddElement/index.vue
mcapeta: ~Linter says it's right in its place 😄 .~

E: Misread your comment. I'm voting for keeping it close to the regular imports and keeping `components` property concise.

client/components/common/tce-core/AddElement/index.vue
mcapeta: If user is willing to add a half width element it's supposed to be allowed to copy only those elements which support half width preview (and are not necessarily half-width elements originally).

client/utils/calculatePosition.js
mcapeta: It's not me, [linter rule](https://eslint.org/docs/rules/no-prototype-builtins) made me to 😄  .

client/components/common/SelectElement/index.vue
mcapeta: Thank you 😉

client/components/common/SelectElement/index.vue
mcapeta: bullseye

client/components/common/SelectElement/index.vue
mcapeta: No reason, I wasn't aware of it. Thanks!

client/components/common/ContentPreview/Element.vue
```diff
@@ -1,19 +1,40 @@
...
+          :class="{ selected: isSelected }"
+          class="content-element" />
+        <v-tooltip open-delay="400" top>
+          <template v-slot:activator="{ on }">
+            <v-btn
+              v-on="on"
+              @click.stop="$emit('element:open', element.uid)"
+              color="pink darken-1"
+              fab small dark
+              :class="{ visible: hover }"
```
underscope: Bindings order

client/components/common/ContentPreview/Element.vue
underscope: Order (positioning before margins)

client/components/common/SelectElement/SelectActivity.vue
```diff
@@ -32,33 +34,37 @@
...
   props: {
-    selectedElements: { type: Array, default: () => [] }
+    selectedElements: { type: Array, default: () => [] },
+    activities: { type: Array, default: () => [] }
   },
   data: () => ({ search: '' }),
   computed: {
-    ...mapGetters('repository', ['activities']),
     groupedSelection: vm => groupBy(vm.selectedElements, 'outlineId'),
+    expandedItems: vm => map(vm.activities, 'id'),
```
underscope: Consider adding Ids in the name (i.e. `expandedActivityIds`)

client/components/common/tce-core/EmbeddedContainer.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <element-list
-    @add="addItem"
+    @add="addItems"
```
underscope, line 3: Backwards comp 🤔 ?

client/components/common/tce-core/EmbeddedContainer.vue
```diff
@@ -21,12 +21,12 @@
 </template>
 
 <script>
+import calculatePosition from 'utils/calculatePosition';
```
underscope, line 24: 🚨 calculatePosition should be part of tce-core (only tce-core and third party packages should be used within tce core; leftover; will be extracted into separate package). Not a must since it was like that before.

client/utils/calculatePosition.js
```diff
@@ -1,47 +1,48 @@
...
  *     item's new position.
- * @param {boolean} isFirstChild Boolean value denoting whether the item should
- *     be placed as the first child of its parent.
+ * @param {boolean} isFirstChild Deprecated: Boolean value denoting whether
+ *     the item should be placed as the first child of its parent.
  * @param {string} [action=REORDER] A string value determining where the item
  *     should be placed in relation to `newPosition`. Defaults to `REORDER` in
  *     case of items reorder.
- * @return {number} Calculated item position.
+ * @param {number} count A number of positions to return.
```
underscope: A number of items with changed position / for which position needs to be determinated ?

client/utils/calculatePosition.js
```diff
@@ -1,47 +1,48 @@
...
-    return ((prev ? prev.position : 0) + current.position) * 0.5;
-  } else if (action === ADD_INTO) {
-    const lastItem = last(items);
-    return lastItem ? lastItem.position + 1 : 1;
+export default function ({ newPosition, items, action = REORDER, count = 1 }) {
+  getDeprecationWarning(...arguments);
+  const arr = [...items];
+  if (action === REORDER) arr.splice(newPosition, count);
+  let index = items.length;
+  if (newPosition !== undefined) {
```
underscope, line 49: Consider documenting logic / scenarios for each case

client/utils/calculatePosition.js
```diff
@@ -1,47 +1,48 @@
...
+  const { position: upper } = items[index] || {};
+  return distributePositions({ lower, upper }, count);
+};
 
 /**
- * Calculates item position based on the options provided.
- * @param {number} newPosition The position of the anchor item by which the new
- *     position will be calculated.
+ * Calculates item position(s) based on the options provided.
+ * @param {number} newPosition The index of the anchor item by which the new
```
underscope, line 25: For add after sake position might be a better name with new changes (can do a improvement PR ?; so it is easier to revisit)

client/utils/calculatePosition.js
```diff
@@ -1,47 +1,48 @@
...
- *     be placed as the first child of its parent.
+ * @param {boolean} isFirstChild Deprecated: Boolean value denoting whether
+ *     the item should be placed as the first child of its parent.
  * @param {string} [action=REORDER] A string value determining where the item
  *     should be placed in relation to `newPosition`. Defaults to `REORDER` in
  *     case of items reorder.
- * @return {number} Calculated item position.
+ * @param {number} count A number of positions to return.
+ * @return {(number|Array)} Single position if `count` is 1 or an array containing
+ * `count` positions.
```
underscope: Indent ?

client/store/modules/repository/activities/actions.js
```diff
@@ -52,11 +52,9 @@ const clone = ({ commit }, mapping) => {
 
 const calculateInsertPosition = ({ state }, { activity, anchor, action }) => {
   const items = getOutlineChildren(state.items, activity.parentId);
+  if (action === ADD_INTO) return calculatePosition({ items, action });
```
underscope: Rather than calling calculatePosition twice determine payload conditionally (it can initially be set to `{ items, action }`) and if `!ADD_INTO` set `newPosition`

client/components/editor/Sidebar/ElementSidebar/ElementMeta/Relationships/Item.vue
```diff
@@ -32,12 +32,14 @@
...
-      @selected="$emit('save', $event)"
+      @selected="select"
       @close="showElementBrowser = false"
       :selected="value"
       :heading="defaultPlaceholder"
       :multiple="multiple"
-      :allowed-types="allowedTypes" />
+      :allowed-types="allowedTypes"
+      use-current-repo
+      header-icon="mdi-transit-connection-variant" />
```
underscope: Order

client/components/editor/Sidebar/ElementSidebar/ElementMeta/Relationships/Item.vue
```diff
@@ -84,6 +86,14 @@ export default {
         message: `Are you sure you want to remove ${label}?`,
         action: () => this.$emit('save', [])
       });
+    },
+    select(elements) {
+      const items = elements.map(it => {
+        if (!it.activity) return it;
+        const { id, activity, activityId: containerId } = it;
```
underscope, line 93: We should probably rename activity or activityId on element (because now it looks like `activityId === activity.id`). Thoughts? 🤔

client/components/editor/ActivityContent/ContainerList.vue
```diff
@@ -12,15 +12,15 @@
...
-      @addElement="addElement"
-      @updateElement="updateElement"
-      @saveElement="saveContentElement"
-      @deleteElement="requestElementDeletion"
-      @insertElement="insertElement"
-      @reorderElement="reorderContentElements"
-      @addSubcontainer="save"
-      @updateSubcontainer="update"
-      @deleteSubcontainer="requestContainerDeletion"
+      v-on="deprecatedListeners"
```
underscope, line 15: 🏆

client/components/editor/ActivityContent/ContainerList.vue
```diff
@@ -91,26 +118,25 @@ export default {
...
       this.save({ type, parentId, position });
     },
-    saveContentElement(element) {
-      const elementChannel = this.$radio.channel(`element:${getElementId(element)}`);
-      return this.saveElement(element)
-        .then(() => {
+    saveContentElements(elements) {
+      castArray(elements).forEach(element => {
+        const elementChannel = this.$radio.channel(`element:${getElementId(element)}`);
+        return this.saveElement(element).then(() => {
```
underscope: Not sure if we should trigger multiple save notifications in this case. Thoughts 🤔 ?

client/components/content-containers/tcc-exam/edit/AssessmentGroup.vue
```diff
@@ -106,17 +106,19 @@ export default {
...
+      assessments.forEach(it => {
+        const uid = uuid();
+        this.$set(this.unsavedAssessments, uid, { ...it, uid });
+      });
     },
     saveAssessment(assessment) {
-      if (assessment.id) return this.$emit('updateElement', assessment);
-      this.$emit('saveElement', assessment);
+      if (assessment.id) return this.$emit('update:element', assessment);
+      this.$emit('save:element', assessment);
```
underscope: Resolve event name using ternary ?

client/components/common/SelectElement/index.vue
```diff
@@ -2,88 +2,176 @@
...
     selected: { type: Array, default: () => [] },
     heading: { type: String, required: true },
+    allowedTypes: { type: Array, required: true },
     multiple: { type: Boolean, default: true },
-    allowedTypes: { type: Array, default: () => [] }
+    submitLabel: { type: String, default: 'Save' },
+    headerIcon: { type: String, default: 'mdi-toy-brick-plus-outline' },
+    useCurrentRepo: { type: Boolean, default: false }
+  },
+  data: () => {
```
underscope: Could be written more compact / less indented  `() => ({.....`

client/components/common/SelectElement/index.vue
```diff
@@ -2,88 +2,176 @@
...
+    return {
+      repository: null,
+      selectedActivity: null,
+      contentContainers: [],
+      selectedElements: [],
+      loadingContent: false,
+      activities: []
```
underscope: Consider grouping selected and fetched props (i.e. repository, activities, contentContainers, selectedActivity, selectedElements)

client/components/common/SelectElement/index.vue
```diff
@@ -2,88 +2,176 @@
...
+    };
+  },
+  computed: {
+    ...mapGetters('repository', {
+      currentRepository: 'repository',
+      currentActivities: 'activities'
+    }),
+    allElementsSelected: vm => vm.selectedElements.length === vm.elements.length,
+    showBackButton: vm => vm.useCurrentRepo ? !!vm.selectedActivity : !!vm.repository,
+    containers() {
```
underscope: having contentContainers and containers is a bit .... Consider renaming this to processedContainers or similar (i.e. other one to selectedActivityContainers) :bowtie:

client/components/common/SelectElement/index.vue
```diff
@@ -93,11 +181,22 @@ export default {
     },
     close() {
       this.$emit('close');
+    },
+    openElement(elementId) {
```
underscope: goTo / openInEditor rather then openElement?

client/components/common/tce-core/AddElement/AddNewElement.vue
```diff
@@ -0,0 +1,123 @@
...
+<script>
+export default {
+  name: 'add-new-element',
+  props: {
+    library: { type: Array, required: true },
+    allowedTypes: { type: Array, required: true }
+  },
+  methods: {
+    isAllowed(type) {
+      if (!this.allowedTypes.length) return true;
```
underscope: Ternary  or || 🤔  ?

client/components/common/tce-core/AddElement/index.vue
```diff
@@ -17,63 +17,67 @@
...
+              background-color="transparent"
+              dark tile borderless mandatory>
+              <v-btn :value="100" icon>
+                <v-icon>mdi-square-outline</v-icon>
+              </v-btn>
+              <v-btn :value="50" icon>
+                <v-icon>mdi-select-compare</v-icon>
+              </v-btn>
+            </v-btn-toggle>
+            <v-divider class="mr-3" vertical />
```
underscope: Order

client/components/common/tce-core/AddElement/index.vue
```diff
@@ -82,23 +86,31 @@ const ELEMENT_GROUPS = [
   { name: 'Nongraded questions', icon: 'mdi-comment-question-outline' }
 ];
 
+const getQuestionData = (element, type) => {
+  const data = { width: LAYOUT.FULL_WIDTH };
+  const question = [{ id: uuid(), data, type: 'JODIT_HTML', embedded: true }];
```
underscope, line 91: Need to create improvement issue for this one, it shouldn't be here... (Ignore for now)

underscope:
ty ❤️ 🏆

client/components/common/tce-core/EmbeddedContainer.vue
mcapeta: Updated to support both single item or an array of items. Should we deprecate single value signature or support both.

Nice catch btw, I just missed the `list-add` slot within `<element-list />`. Writing down all the things that can be customized within Tailor 😄

client/components/editor/ActivityContent/ContainerList.vue
mcapeta: Reworked to show one notification once all elements are successfully saved. Next step is to save element in batch, as discussed (another PR).

client/components/common/ContentPreview/Element.vue
```diff
@@ -1,19 +1,40 @@
...
-      :class="{ selected: isSelected }"
-      is-disabled />
+    <v-hover v-slot:default="{ hover }">
+      <div class="element-wrapper">
+        <content-element
+          v-bind="$attrs"
+          :element="element"
+          :set-width="false"
+          class="content-element"
+          :class="{ selected: isSelected }" />
```
MiroDojkic: Please switch prop order (bound props first).

client/components/common/ContentPreview/Element.vue
```diff
@@ -1,19 +1,40 @@
...
+        <v-tooltip open-delay="400" top>
+          <template v-slot:activator="{ on }">
+            <v-btn
+              v-on="on"
+              @click.stop="$emit('element:open', element.uid)"
+              color="pink darken-1"
+              fab small dark
+              class="open-element-button"
+              :class="{ visible: hover }">
```
MiroDojkic: Sorry for nitpicking but I think the order of bound, static, boolean and then class props would be consistent with the rest of the codebase. 🙃

client/components/common/ContentPreview/Element.vue
```diff
@@ -1,19 +1,40 @@
 <template>
-  <div @click="toggleSelection" class="element-preview-container">
+  <div :class="elementWidth" class="element-preview-container">
```
MiroDojkic: Please consider if this approach would make it more straightforward to infer what's going on with the layout straight from reading the template:
```suggestion
  <div :class="col-xs-`${elementWidth} float-none`" class="element-preview-container">
```

client/components/common/ContentPreview/Element.vue
```diff
@@ -59,4 +96,20 @@ export default {
     word-break: break-all;
   }
 }
+
+.element-wrapper {
+  position: relative;
+  flex-grow: 1;
```
MiroDojkic: Consider using vuetify's `flex-grow-1` util class for this.

client/components/common/SelectElement/SelectActivity.vue
```diff
@@ -32,33 +34,37 @@
...
-    hasSearchResults() {
-      if (!this.search || !this.$refs) return true;
-      const { excludedItems, nodes } = this.$refs.treeview;
-      return excludedItems.size !== Object.keys(nodes).length;
+    noResultsMessage() {
+      const { activities, search, $refs } = this;
+      if (!activities.length) return 'Empty repository';
+      if (!search || !$refs) return '';
+      const { excludedItems, nodes } = $refs.treeview;
+      if (excludedItems.size !== Object.keys(nodes).length) return '';
```
MiroDojkic: Would assigning this condition to boolean variable make it a bit simpler to read?
I.e.
```suggestion
      const hasSearchResults = excludedItems.size !== Object.keys(nodes).length;
      if (excludedItems.size !== Object.keys(nodes).length) return '';
```

client/components/common/SelectElement/SelectActivity.vue
```diff
@@ -81,5 +87,11 @@ export default {
   .v-chip.custom-chip {
     border-radius: 12px !important;
   }
+
+  ::v-deep .v-treeview-node--leaf {
+    > .v-treeview-node__root, .v-treeview-node__content > * {
+      cursor: auto;
+    }
+  }
```
MiroDojkic: Do you think this would make the selector easier to read or is it too cryptic? 🤔 
```suggestion
  ::v-deep .v-treeview-node {
    &--leaf > &__root, &--leaf > &__content > * {
      cursor: auto;
    }
  }
```

client/components/common/SelectElement/SelectRepository.vue
```diff
@@ -0,0 +1,48 @@
...
+    repositories: [],
+    loading: false
+  }),
+  methods: {
+    select(repository) {
+      if (find(this.repositories, { id: repository.id })) {
+        this.$emit('selected', repository);
+      }
+    },
+    fetchRepositories: debounce(loader(function (search) {
```
MiroDojkic, line 39: Is the name `fetch` good enough as in the case of `select` method - if the component's context makes it easy to figure out what it does?
If not, should we rename `select` to `selectRepository`?

client/components/common/SelectElement/index.vue
```diff
@@ -2,88 +2,174 @@
...
   name: 'select-element',
   props: {
     selected: { type: Array, default: () => [] },
     heading: { type: String, required: true },
+    allowedTypes: { type: Array, required: true },
     multiple: { type: Boolean, default: true },
-    allowedTypes: { type: Array, default: () => [] }
+    submitLabel: { type: String, default: 'Save' },
+    headerIcon: { type: String, default: 'mdi-toy-brick-plus-outline' },
+    useCurrentRepo: { type: Boolean, default: false }
```
MiroDojkic: Consider renaming to `onlyCurrentRepo` or something similar to indicate that it's a boolean type of value.

client/components/editor/Sidebar/ElementSidebar/ElementMeta/Relationships/Item.vue
mcapeta: It is confusing, though. If we're going to change what's preserved in db we should migrate data, aren't we? If you thought of renaming properties on FE only, I would personally leave it as is in one place or move it [here](https://github.com/ExtensionEngine/tailor/pull/624/files#diff-62af149e0cab2f7e0067b49fd93a40766741bf76f511f7f476577a3e6abaf1b5R141) where this data actually come from (but still keep in one place).

And if you have any better idea please let me know, maybe I just misunderstood your suggestion 🤷‍♂️ .

client/components/common/SelectElement/index.vue
```diff
@@ -2,88 +2,174 @@
...
+      currentRepository: 'repository',
+      currentActivities: 'activities'
+    }),
+    allElementsSelected: vm => vm.selectedElements.length === vm.elements.length,
+    showBackButton: vm => vm.useCurrentRepo ? !!vm.selectedActivity : !!vm.repository,
+    processedContainers() {
+      const { selectedActivity, activities } = this;
+      if (!selectedActivity || !activities.length) return [];
+      const rootTypes = getContainerTypes(selectedActivity.type);
+      let containers = activities.filter(({ type, parentId }) => {
```
MiroDojkic: Please consider moving arrow functions to component methods to make this computed prop more straightforward to read.

client/components/common/SelectElement/index.vue
```diff
@@ -2,88 +2,174 @@
...
-        elements: elements.filter(element => element.activityId === container.id)
-      }));
+      const { processedContainers } = this;
+      const elements = await this.fetchElements(processedContainers);
+      this.contentContainers = processedContainers.map(container => {
+        const containerElements = elements
+          .filter(it => it.activityId === container.id)
+          .map(element => ({ ...element, activity }));
+        return { ...container, elements: sortBy(containerElements, 'position') };
+      });
```
MiroDojkic: It might be worth to abstract adding elements to a container here:
```suggestion
      this.contentContainers = processedContainers.map(this.addContainerElements);
```

client/components/common/mixins/deprecation.js
```diff
@@ -0,0 +1,21 @@
...
+    '${oldEvent}' listener is deprecated and will no longer be used!
+    Please emit '${newEvent}' instead on your content containers`);
+};
+
+export default {
+  methods: {
+    deprecateEvent(handler, { oldEvent, newEvent, adaptArgs }) {
+      newEvent = newEvent || noCase(oldEvent, { delimiter: ':' });
+      const context = this;
+      return function () {
```
MiroDojkic: Why not use arrow function and avoid the need to bind this?

client/components/common/tce-core/AddElement/AddNewElement.vue
```diff
@@ -0,0 +1,122 @@
...
+        </div>
+        <div class="group-elements">
+          <v-hover
+            v-for="element in group.elements"
+            :key="element.position"
+            v-slot:default="{ hover }">
+            <v-btn
+              @click.stop="$emit('add', [element])"
+              :disabled="!isAllowed(element.type)"
+              :color="hover ? 'pink' : '#333'"
```
MiroDojkic: Should we use `secondary` instead of `pink`, in case we ever decide to change palette?

client/components/common/tce-core/AddElement/AddNewElement.vue
```diff
@@ -0,0 +1,122 @@
...
+  padding: 0 0 1.875rem;
+  background: #fff;
+}
+
+.group-heading {
+  margin: 0 2.5rem 0.375rem;
+  padding-top: 1.25rem;
+  font-size: 1rem;
+  font-weight: 500;
+  line-height: 1.75rem;
```
MiroDojkic: Do we want to leave this to change relative to font size (in the future)? If that's the case, we could use `1.75;` or `1.75em;`.

client/components/editor/ActivityContent/ContainerList.vue
```diff
@@ -91,26 +120,32 @@ export default {
...
+      // TODO: implement endpoint to save multiple elements at once
+      const contentElements = castArray(elements);
+      Promise.map(contentElements, element => {
+        const elementChannel = this.$radio.channel(`element:${getElementId(element)}`);
+        return this.saveElement(element)
+          .then(() => elementChannel.emit('saved'))
+          .catch(err => {
+            elementChannel.emit('error', err);
+            return Promise.reject(err);
+          });
```
MiroDojkic: Can we flatten this promise chain by moving getting channel and emitting event through channel to `saveElement` method?

MiroDojkic (commented):
Here's a couple of small suggestions (mostly format-related), please see if any of these make any sense. 😄

client/components/common/SelectElement/SelectActivity.vue
mcapeta: I wouldn't say it's cryptic, I like it. It is clear enough we're updating cursor behaviour somewhere within `.v-treeview-node` and readability of other selectors is pretty much the same in both cases so I pick concise one 😉 .

client/components/common/SelectElement/index.vue
mcapeta: Initial version was `useOnlyCurrentRepo`. Seems I rejected the wrong one 😄 . Updating as suggested...

client/components/common/mixins/deprecation.js
mcapeta: Because of `arguments` binding. Arrow functions does not have them (i.e. they access arguments of the first non-arrow parent) if I'm not wrong. Leveraging them here to be able to decorate methods with different signatures...
I'll check it one more time before I resolve this one...

client/components/common/SelectElement/index.vue
mcapeta: Some more context variables needed (`activity`, `elements`) so please take a look if you find new solution more readable 😉 .

client/components/editor/ActivityContent/ContainerList.vue
mcapeta: > Here's a couple of small suggestions (mostly format-related), please see if any of these make any sense. 😄

Thank you 😉

mcapeta:
> Here's a couple of small suggestions (mostly format-related), please see if any of these make any sense. 😄

Thank you 😉

client/components/editor/Sidebar/ElementSidebar/ElementMeta/Relationships/Item.vue
underscope: I was thinking that for readability sake of this component within processor that is hydrating state of this component (component that you linked) we do the rename that you did here: `activityId => containerId` (and/or `activity => outlineActivity`) so this is more readable and we know which activity id represents which activity (one is outline parent and other one is container activity). Is good for now, consider it as an improvement.

client/components/common/tce-core/EmbeddedContainer.vue
underscope: Lets support both 👍

client/components/common/SelectElement/index.vue
```diff
@@ -2,102 +2,212 @@
...
-    allowedTypes: { type: Array, default: () => [] }
+    submitLabel: { type: String, default: 'Save' },
+    headerIcon: { type: String, default: 'mdi-toy-brick-plus-outline' },
+    onlyCurrentRepo: { type: Boolean, default: false }
   },
   data: () => ({
-    selectedActivity: null,
-    contentContainers: [],
-    selectedElements: [],
+    items: {
```
underscope, line 91: Grouping this under items seems bit off/too generic. Consider improving this (separate PR); if selection is used to group the selected, should these be something like `selectionOptions` ? 🤔

client/components/common/SelectElement/index.vue
```diff
@@ -2,102 +2,212 @@
...
+    },
+    processedContainers() {
+      const { selection: { activity }, items: { activities } } = this;
+      if (!activity || !activities.length) return [];
+      const containers = sortBy(activities.filter(this.isRootContainer), [
+        this.getTypePosition, 'position', 'createdAt'
+      ]);
+      return flatMap(containers, it => [it, ...this.getSubcontainers(it)]);
+    },
+    elements() {
```
underscope, line 120: processedElements 🤔  / elementSelectionOptions  🤡   ?

client/components/common/SelectElement/index.vue
```diff
@@ -2,102 +2,212 @@
...
+    ...mapGetters('repository', {
+      currentRepository: 'repository',
+      currentActivities: 'activities'
+    }),
+    allElementsSelected: vm => vm.selection.elements.length === vm.elements.length,
+    rootContainerTypes() {
+      const type = this.selection.activity?.type;
+      return type && getContainerTypes(type);
+    },
+    processedContainers() {
```
underscope, line 112: It might be just a function (we have already a lot of `something` containers on the context. It is only used within a single function so it can be eliminated; Separate PR)

client/components/common/SelectElement/index.vue
underscope, line 120: Same as prev one, it can be just a function which will assign to items.elements / selectionOptions.elements. Then some other aspects will be more readable like:
`vm.selection.elements.length === vm.elements.length`
will become =>
`vm.selection.elements.length === vm.selectionOptions.elements.length`.
Can filtering of allowedTypes be moved to => assignElements? That one should probably be more descriptive as well 🤔

client/components/common/SelectElement/index.vue
```diff
@@ -2,102 +2,212 @@
...
+      ]);
+      return flatMap(containers, it => [it, ...this.getSubcontainers(it)]);
+    },
+    elements() {
+      const elements = flatMap(this.items.contentContainers, 'elements');
+      if (!this.allowedTypes.length) return elements;
+      return elements.filter(it => this.allowedTypes.includes(it.type));
+    },
+    toggleButton() {
+      const { allElementsSelected, elements, multiple, selection } = this;
```
underscope, line 126: 🔝 This one as well; see prev comment; elements => selectionOptions.elements

client/components/common/SelectElement/index.vue
```diff
@@ -2,102 +2,212 @@
...
+      if (!multiple || !selection.activity || !elements.length) return;
+      const { SELECT, DESELECT } = TOGGLE_BUTTON;
+      return allElementsSelected ? DESELECT : SELECT;
+    }
+  },
   methods: {
+    getTypePosition({ type }) {
```
underscope, line 133: Not sure if it makes sense here since it is only used once and is one liner.

PR #627, opened 2020-10-13
Snackbar bugfix

borna223:
Screenshot 2020-10-13 at 15 02 29

Caused by: https://github.com/vuetifyjs/vuetify/blob/v2.3.12/packages/vuetify/src/components/VSnackbar/VSnackbar.ts#L105


borna223:
Fixed by https://github.com/ExtensionEngine/tailor/pull/626

PR #639, opened 2020-10-18
Fix "loader" decorator

mcapeta:
Fixes issue with passing arguments to decorated function...


vladimyr:
> Fixes issue with passing arguments to decorated function...

TIL functions can have arguments 🙃
Thanks for cleaning my mess. 👍

PR #641, opened 2020-10-20
Implement input asset empty save and cancel button feature

### LINKED ISSUE #622: Add cancel button when deleting uploaded pdf/video
## I'm submitting a...

- [ ] Bug report  
- [x] Feature request

## Context
After deleting an uploaded pdf or video element, i have no cancel option available. I should be able to cancel my deletion (if i refresh the page without saving the deletion is cancelled anyway).
CancelButton

hrvojevu:
This PR:
- enables empty asset in `tce-pdf` and `tce-video` to be saved
- adds cancel button functionality which acts as undo when asset is deleted in editing mode

### DIFF
--- a/client/components/common/InputAsset.vue
+++ b/client/components/common/InputAsset.vue
@@ -55,14 +55,24 @@
       class="action">
       Edit
     </v-btn>
-    <v-btn
-      v-else
-      @click.stop="save"
-      :disabled="uploading || !hasAsset"
-      text
-      class="action">
-      {{ hasChanges ? 'Save' : 'Cancel' }}
-    </v-btn>
+    <template v-else>
+      <v-btn
+        v-if="hasChanges"
+        @click.stop="save"
+        :disabled="uploading"
+        text
+        class="action">
+        Save
+      </v-btn>
+      <v-btn
+        v-if="hasChanges || url"
+        @click.stop="cancel"
+        :disabled="uploading"
+        text
+        class="action">
+        Cancel
+      </v-btn>
+    </template>
   </v-toolbar-items>
 </template>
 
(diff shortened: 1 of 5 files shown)


kjuej:
@hrvojevu Seems to be working as expected.

client/components/content-elements/tce-video/edit/Toolbar.vue
```diff
@@ -31,9 +31,13 @@ export default {
     url: vm => get(vm.element, 'data.assets.url')
   },
   methods: {
-    save({ url, publicUrl }) {
+    save({ url }) {
       const element = cloneDeep(this.element);
-      set(element.data, 'assets.url', url);
+      if (url) set(element.data, 'assets.url', url);
+      else {
+        set(element.data, 'assets', null);
```
underscope: This might break future functionality, i.e. adding thumbnail or similar.  Please override just url.

client/components/editor/ContentElement.vue
```diff
@@ -47,7 +48,7 @@ export default {
     },
     save: loader(async function (data) {
       const element = cloneDeep(this.element);
-      Object.assign(element.data, data);
+      set(element, 'data', data);
```
underscope: `element` should never be null here in order for set to be required.

client/components/content-elements/tce-pdf/edit/Toolbar.vue
```diff
@@ -31,9 +31,13 @@ export default {
...
   },
   methods: {
-    save({ url, publicUrl }) {
+    save({ url }) {
       const element = cloneDeep(this.element);
-      set(element.data, 'assets.url', url);
+      if (url) set(element.data, 'assets.url', url);
+      else {
+        set(element.data, 'assets', null);
+        set(element.data, 'url', null);
```
underscope: Since data.url is virtual prop it might make sense to update patch action on the content element ctrl to refetch. That will remove the need for cleaning up virtual props (data.url in this case).

client/components/common/InputAsset.vue
```diff
@@ -116,6 +126,12 @@ export default {
       this.isEditing = false;
       const payload = this.file || { url: this.urlInput, publicUrl: this.urlInput };
       this.$emit('input', payload);
+    },
+    cancel() {
+      const isLinked = !isUploaded(this.url);
+      this.urlInput = isLinked ? this.url : null;
+      this.file = isLinked ? null : pick(this, ['url', 'publicUrl']);
```
underscope, line 133: See the comment below regarding virtual prop cleanup.

underscope (changes requested):

kjuej:
@hrvojevu @underscope Working as expected.

PR #644, opened 2020-10-23
Update content element deletedAt field manually on patch

### LINKED ISSUE #640: Unable to restore deleted element via history
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request

## Context
RestoreIssue

hrvojevu:
This PR:
- manually sets `deletedAt` field in patch controller via `setDataValue` since `deletedAt` is not updated correctly with only passing updated properties using `update` method
- solves issue https://github.com/ExtensionEngine/tailor/issues/640

### DIFF
--- a/server/content-element/content-element.controller.js
+++ b/server/content-element/content-element.controller.js
@@ -38,7 +38,10 @@ function patch({ repository, user, body, params: { elementId } }, res) {
   const context = { userId: user.id, repository };
   return ContentElement.findByPk(elementId, { paranoid })
     .then(asset => asset || createError(NOT_FOUND, 'Element not found'))
-    .then(asset => asset.update(data, { context }))
+    .then(asset => {
+      asset.setDataValue('deletedAt', data.deletedAt);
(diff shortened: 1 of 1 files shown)


server/content-element/content-element.controller.js
```diff
@@ -38,7 +38,10 @@ function patch({ repository, user, body, params: { elementId } }, res) {
   const context = { userId: user.id, repository };
   return ContentElement.findByPk(elementId, { paranoid })
     .then(asset => asset || createError(NOT_FOUND, 'Element not found'))
-    .then(asset => asset.update(data, { context }))
+    .then(asset => {
+      asset.setDataValue('deletedAt', data.deletedAt);
```
underscope: @hrvojevu 
It should be only possible to restore assets (update deletedAt) if deleted (please make it a bit more specific, add a check and set it explicitly to null).
Thx!

underscope:
@kjuej

kjuej:

PR #647, opened 2020-10-25
Restored the deprecated tes prop for containers

abasic:
This PR restores the deprecated `tes` prop for containers so that existing custom content containers continue to function without needing to rename it the prop to `elements`. Because the prop is deprecated, a deprecation message is shown if a container is found using the prop.

Also, a small validation service was added that encapsulates the validation messaging logic.

### DIFF
--- a/client/content-plugins/ContainerRegistry.js
+++ b/client/content-plugins/ContainerRegistry.js
@@ -1,14 +1,26 @@
 import ComponentRegistry from './ComponentRegistry';
 import containerList from 'shared/core-containers';
+import get from 'lodash/get';
 import { getContainerTemplateId as getId } from 'shared/activities';
 import { getContainerName as getName } from 'tce-core/utils';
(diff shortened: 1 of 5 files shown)


underscope:
@kjuej Please revisit on develop

client/content-plugins/ContainerRegistry.js
```diff
@@ -1,14 +1,26 @@
...
+import { service as ValidationService } from './validation';
 
-const validator = ({ templateId, type }) => {
-  if (templateId) return;
-  console.warn(`
-    For container ${type} using depricated type identification!
-    Use templateId instead!
-  `);
+const getTemplateMessage = name => `
+  For container ${name} using depricated 'type' identification!
```
vladimyr, line 9: Umm, you mean _deprecated_ 🙃

abasic: /giphy ups

PR #652, opened 2020-10-30
Implement OpenID Connect authentication

### LINKED ISSUE #632: SSO
- [x] Implement OpenID Connect login (#652)
- [x] Move JWT token to cookie (#689)
- [x] Allow signup over OIDC (#697)
- [ ] Add silent auth over OIDC
- [ ] Allow OIDC Login branding
- [ ] Allow user management on IdP with adapters
- [ ] Implement discussed auth improvements

kronicker:
Subtask of #632 

### This PR:
- [x] Adds OIDC config
- [x] Adds OIDC auth strategy
- [x] Adds OIDC routes and controller
- [x] Adds customizable _Login with OAuth_ button to Login screen
- [x] Removes versioning from the root API path
- [x] Allows user with denied access to login with a different account on IdP

### UI
|Before|After|
|---|---|
|Screenshot 2020-11-05 at 11 33 28|Screenshot 2020-11-05 at 11 32 28|

### QA Note:
**🔍 Tested on:**
- [x] Auth0
- [x] Okta

with test accounts. If you need access, please ping me.


.env.example
```diff
@@ -3,7 +3,9 @@ HOSTNAME=localhost
 PORT=3000
 PROTOCOL=http
 REVERSE_PROXY_PORT=8080
-API_PATH=/api/v1/
+
+# Local storage configuration
+STORAGE_STATE_KEY=TAILOR_APP_STATE
```
kronicker: @ExtensionEngine/tailor Does this make sense? I wanted a central place where this is defined so both client and server can be aware of it and the easiest approach was `.env`. That being said I'm not sure if there is value in this being configurable or should I find a better place for it to live and remain _hardcoded_?

kronicker: After offline discussion, I've decided to leave it hardcoded for now and remove it completely in the following PR.

PR #653, opened 2020-10-30
Introduce `meta-combobox` component

### LINKED ISSUE #82: Add tagging support to the Select component
## I'm submitting a...

- [ ] Bug report  
- [x] Feature request

## Context

As the title says, it should be possible to add tags with the Select component (configurable through the meta config as well).

More info: https://vue-multiselect.js.org/#sub-tagging

hrvojevu:
This PR:
- introduces new `meta-combobox` component
- removes `client/components/common/Select.vue` and `vue-multiselect` as they are not used in the project any more


client/components/meta-inputs/meta-select/Edit.vue
```diff
@@ -1,5 +1,6 @@
 <template>
-  <v-select
+  <component
+    :is="meta.tagging ? 'v-combobox' : 'v-select'"
```
underscope: Stylewise chips are tag like so tagging as a configuration prop is a bit confusing to me 🤷‍♂️, consider some other option name or a separate (combobox) component 🤔 Make sure it is documented [here](https://github.com/ExtensionEngine/tailor/wiki/Configurable-inputs) afterwards.

client/components/common/Select.vue
```diff
@@ -1,239 +0,0 @@
-<template>
```
underscope, line 1: 🏆

PR #661, opened 2020-11-05
Fix position issue when using `ADD_INTO` action

### LINKED ISSUE #608: Position issue when copying multiple elements at once
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request

## Context
When using "Copy existing" feature, and copying multiple elements at once, all copied elements have the same position assigned to them.

hrvojevu:
This PR:
- expands `calculateInsertPosition` action and `calculatePosition` to handle use case when `ADD_INTO` action is used
- fixes https://github.com/ExtensionEngine/tailor/issues/608


kjuej:
@hrvojevu 
copyBelowIss
Copy below issue still occurs

client/store/modules/repository/activities/actions.js
```diff
@@ -46,7 +48,8 @@ const clone = ({ commit }, mapping) => {
 };
 
 const calculateInsertPosition = ({ state }, { activity, anchor, action }) => {
-  const items = getOutlineChildren(state.items, activity.parentId);
+  const id = action === ADD_INTO ? anchor.id : activity.parentId;
```
underscope: Consider renaming to parentId

client/store/modules/repository/activities/actions.js
```diff
@@ -46,7 +48,8 @@ const clone = ({ commit }, mapping) => {
...
+  const items = getOutlineChildren(state.items, id);
```
underscope: In the following line, for `ADD_INTO` case, if items are children of anchor, how can the item be found 🤔 ?

kjuej:
@hrvojevu Seems to be working now.

bubafinder:
The order of multiple copied elements is not retained as in the original repository.
For example, Activity elements ordered 1, 2, 3 are copied in order 3, 2, 1.

client/store/modules/repository/activities/actions.js
```diff
@@ -55,7 +57,19 @@ const calculateInsertPosition = ({ state }, { activity, anchor, action }) => {
   return calculatePosition(context);
 };
 
+const calculateCopyInsertPosition = ({ state }, { anchor, action, copyIndex }) => {
+  let items = getOutlineChildren(state.items, anchor.parentId);
+  if (action === ADD_INTO) {
+    items = getOutlineChildren(state.items, anchor.id);
+    return calculatePosition({ items, action });
+  }
+  const newPosition = findIndex(items, { id: anchor.id }) + copyIndex;
```
underscope: Can we use the newly created (from the previous loop) anchor/node instead of adding a copy index 🤔 ?

hrvojevu: @underscope refactored the code to use previously created activity as anchor. I would still keep both `calculateInsertPosition` and `calculateCopyInsertPosition` since using only `calculateInsertPosition` would increase complexity of the action and decrease readability.

client/store/modules/repository/activities/actions.js
```diff
@@ -55,7 +60,19 @@ const calculateInsertPosition = ({ state }, { activity, anchor, action }) => {
   return calculatePosition(context);
 };
 
+const calculateCopyInsertPosition = ({ state }, { anchor, action }) => {
```
underscope: Let's rename it to calculateCopyPosition since it is always used for copy ;)

client/components/repository/common/CopyActivity/index.vue
```diff
@@ -112,25 +114,29 @@ export default {
     },
     async copyActivity(activity) {
       const { id: srcId, repositoryId: srcRepositoryId, type } = activity;
-      const { anchor, action } = this;
+      const { action } = this;
+      const anchor = (action === ADD_AFTER && last(this.copiedActivities)) || this.anchor;
```
underscope, line 119: If the activity being copied has sub-activities, isn't the last going to pick up random last activity (out of that activity and its respective subactivities array). Should we just extract the initial activity upon loading the copiedActivities array?

client/store/modules/repository/activities/actions.js
```diff
@@ -55,7 +60,19 @@ const calculateInsertPosition = ({ state }, { activity, anchor, action }) => {
...
+  let items = getOutlineChildren(state.items, anchor.parentId);
+  if (action === ADD_INTO) {
+    items = getOutlineChildren(state.items, anchor.id);
```
underscope: Why not put this first?

client/store/modules/repository/activities/actions.js
```diff
@@ -55,7 +60,19 @@ const calculateInsertPosition = ({ state }, { activity, anchor, action }) => {
...
+    return calculatePosition({ items, action });
+  }
+  const newPosition = findIndex(items, { id: anchor.id });
```
underscope, line 67: When does use case without anchor exist ?

underscope:
@kjuej Please revisit on develop

PR #663, opened 2020-11-07
Fix duplicated comments

### LINKED ISSUE #658: Last comment repeated multiple times on gmail notification
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request

## Context
I added 3 comments ("comm1", "comm2" and "comm3") and last comment is always repeated multiple times on gmail:
Screen Shot 2020-11-02 at 16 06 20

borna223:
Fixes #658


server/shared/mail/templates/comment.mjml
```diff
@@ -16,14 +16,34 @@
       </mj-column>
     </mj-section>
     <mj-section padding="0 10px">
-			<mj-column border-bottom="2px dashed #d2d2d2">
-        <mj-include path="./components/user-row.mjml"/>
+      <mj-column border-bottom="2px dashed #d2d2d2">
+        <mj-table padding="10px">
```
underscope, line 20: Why returning the extracted common code?

borna223: Because mustache syntax in the extracted file doesn't have the correct context (in the case of `previousComments` array). The first `mustache.render` call repeats the following block for each element of the `previousComments` array:
```xml
<mj-section padding="0 10px">
  <mj-column background-color="#f5f5f5" border-bottom="2px dashed #d2d2d2">
    <mj-include path="./components/user-row.mjml"/>
  </mj-column>
</mj-section>
```

Then `mjml2html` expands `<mj-include>` tags and 2nd `mustache.render` call handles leftover mustache syntax using the global context instead of the "array item" context.

Issue #656, opened 2020-11-02
Minor UI improvements


bubafinder:
## I'm submitting a...

- [ ] Bug report  
- [x] Improvements
- [ ] Regression (a behavior that used to work and stopped working in a new release)
- [ ] Feature request
- [ ] Documentation issue or request

## Minor UI improvements 

1. Improve the hover background contrast for icons on the right side of the search input
![image](https://user-images.githubusercontent.com/41724080/97878783-5e680880-1d1f-11eb-9066-f3560fbdb1ab.png)

2. Improve the hover background contrast for the 'Open settings' icon in the repository card
![image](https://user-images.githubusercontent.com/41724080/97880359-5e690800-1d21-11eb-9be1-741da20ef6be.png)

3. Order tags by name in tag dropdown
![image](https://user-images.githubusercontent.com/41724080/97880834-ebac5c80-1d21-11eb-928c-771466a33f86.png)

4. Add selected style for searched Activity item
![image](https://user-images.githubusercontent.com/41724080/97880947-10083900-1d22-11eb-9116-cb6ff65b0ccc.png)

5. Improve Activity background color/contrast for default and hover/selected state
![image](https://user-images.githubusercontent.com/41724080/97881642-f1567200-1d22-11eb-8da8-a092835c6fc1.png)

6. Improve the hover style of the repository card. Only 'Edited [time] by [user]' text is changed in the card (bolded). Maybe to change the weight of the title as well? This is still open for discussion
![image](https://user-images.githubusercontent.com/41724080/97995380-b9186780-1de6-11eb-931f-ef799cb99c06.png)

7. Also opened for discussion. Because the date format is M/D/Y, remove the 24-hours clock and ad AM/PM to the time format. Please make sure that the same date&time format is used on the whole platform.
![image](https://user-images.githubusercontent.com/41724080/97997015-cb93a080-1de8-11eb-8c6a-70506e7ddbdb.png)

underscope:
@markolovric Closing this one, please reopen new one for the next milestone with remaining requests (two separate tickets)

PR #673, opened 2020-11-12
Improve search input icon hover contrast

hrvojevu:
This PR:
- updates icons hover background opacity to improve background contrast per points 1 and 2 of issue https://github.com/ExtensionEngine/tailor/issues/656

Screen Shot 2020-11-12 at 6 37 52 PM

### DIFF
--- a/client/components/catalog/Card/index.vue
+++ b/client/components/catalog/Card/index.vue
@@ -25,7 +25,7 @@
               <v-btn
                 v-on="on"
                 @click.stop="navigateTo('repository-info')"
-                color="blue-grey darken-2"
+                color="blue-grey darken-1"
                 icon
                 class="repo-info mr-2">
                 <v-icon>mdi-settings</v-icon>
--- a/client/components/catalog/Card/index.vue
+++ b/client/components/catalog/Card/index.vue
@@ -25,7 +25,7 @@
               <v-btn
                 v-on="on"
                 @click.stop="navigateTo('repository-info')"
-                color="blue-grey darken-1"
+                color="blue-grey lighten-2"
                 icon
                 class="mr-2">
                 <v-icon>mdi-settings</v-icon>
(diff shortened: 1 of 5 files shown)


bubafinder:
Please use the same background color as under Pin and Tag icon.

image

client/components/catalog/SelectOrder.vue
```diff
@@ -6,7 +6,7 @@
           <template v-slot:activator="{ on: tooltip }">
             <v-btn
               v-on="{ ...menu, ...tooltip }"
-              icon text
+              icon
```
underscope, line 10: Color is defined on icon rather than on btn, if you move it here it should solve hover as well ;)

client/components/catalog/TagFilter.vue
```diff
@@ -10,7 +10,7 @@
         open-delay="800"
         top>
         <template v-slot:activator="{ on: tooltip }">
-          <v-btn v-on="{ ...menu, ...tooltip }" icon text>
+          <v-btn v-on="{ ...menu, ...tooltip }" icon>
```
underscope: Same as prev

client/components/catalog/Container.vue
```diff
@@ -14,7 +14,7 @@
               <v-btn
                 v-on="on"
                 @click="onFilterChange(togglePinned)"
-                icon text
+                icon
```
underscope, line 18: Same as prev.

client/components/catalog/Card/index.vue
```diff
@@ -25,7 +25,7 @@
               <v-btn
                 v-on="on"
                 @click.stop="navigateTo('repository-info')"
-                color="blue-grey darken-1"
+                color="blue-grey lighten-2"
```
underscope: We intentionally blended this icon a bit. Don't know about the @markolovric but being repeated on all cards now it looks a bit too much to me.

PR #677, opened 2020-11-17
Ensure that customResolver is returning a promise

### LINKED ISSUE #676: Error is thrown when handleStatics method doesn't return a promise
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request

## Failure Information (for bugs)
Error `resolveStatics(...).then is not a function` is thrown in the `revision.controller.js` (line 23) when content-element/util.js [handleStatics](https://github.com/ExtensionEngine/tailor-element-template/blob/master/util.js) doesn't return a promise.

For example: 
```js
function handleStatics(element, defaultHandler, generalHandler) {
	return element;
}
```

## Steps to Reproduce
1. To reproduce this error create a [custom content element](https://github.com/ExtensionEngine/tailor-element-template) or edit the existing one by adding `util.js` file with custom `handleStatics` handler
2. Edit `handleStatics` function to return non-promise value (for example: `return element;`)
3. Create newly added (or edited) content element inside activity
4. Edit content element a few times
5. Go to the `history` page and choose revision related to your content element
6. By this point, you should receive a blank history screen and the error described above

ikovac:
Closes #676

### DIFF
--- a/server/shared/content-plugins/elementRegistry.js
+++ b/server/shared/content-plugins/elementRegistry.js
@@ -24,7 +24,11 @@ class ElementsRegistry extends BaseRegistry {
   }
 
   getStaticsHandler(type) {
-    return this._staticsHandler[type];
+    const handler = this._staticsHandler[type];
+    if (!handler) return;
+    return (...args) => {
+      return Promise.resolve(handler(...args));
+    };
   }
 }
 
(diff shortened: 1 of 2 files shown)


abasic:
@ikovac 
Something like this should not really happen since resolving statics should always be async. But let's handle it just in case. Handling for this should be in the registry that can be found [here](https://github.com/ExtensionEngine/tailor/tree/develop/server/shared/content-plugins). There are other methods as well that should be async and there is also an element registry. The handling (some sort of async wrapper) should be introduced during the assignment of the method to the resolver or builder. Let me know if you have any questions.

ikovac:
@abasic Thank you for the review. I moved this logic inside the element registry.

PR #679, opened 2020-11-17
Feature: Improve image loading speed

ikovac:
Avoid converting an image to base64 at the mounted hook and load an image with the original image URL.
Images load much faster. In the example below, a large image (1.8MB) is used and the result is that loading speed is improved by more than 2 seconds.

## Current
The image took much longer to show
Kapture 2020-11-17 at 09 58 11

## New
The image loads much faster
Kapture 2020-11-17 at 09 53 37

!! This PR needs further discussion. Maybe there is some good reason behind base64 conversion.


underscope:
@ikovac What's the state of this one (why it has do not merge on)?

ikovac:
@underscope I put the `do not merge` label because I'm waiting for a response. If the logic behind this PR is OK, I will improve a few other things related to the image loading.

underscope:
Looks good, please proceed and cleanup other aspects (i.e. naming of load func param). Thx!

PR #667, opened 2020-11-10
Comment improvements

borna223:
Resolves #657 with the exception of:
> 2. After the user receives an email for the user comment and clicks on the View button, scroll to the related comment on the opened page

Not sure if worth doing now if async loading of old comments is in plan?

PR #681, opened 2020-11-19
Use 12-hour time format with AM/PM instead of 24-hours

hrvojevu:
This PR:
- updates 24-hours date time format to 12-hour format with AM/PM on all places where 24-hours format was used, per point 7 of issue https://github.com/ExtensionEngine/tailor/issues/656


bubafinder:
@underscope It would be good to merge #667 to the develop first and then merge it to this branch, we have new date & time position there

underscope:
@markolovric Lets make that happen within #667

PR #683, opened 2020-11-20
Task: Remove hover effect from the chip

ikovac:
**This PR:**
Remove hover effect from the chip component. The logic behind this PR is that if the chip is not clickable, it should not have a hover effect.


underscope (commented):
https://github.com/vuetifyjs/vuetify/issues/10115

client/components/common/Chip.vue
```diff
@@ -0,0 +1,18 @@
+<template>
```
underscope: Implement `readonly` prop and override the current v-chip  (don't introduce new component) or use a workaround in the comment.

client/components/catalog/TagFilterSelection.vue
```diff
@@ -5,7 +5,7 @@
       :key="tag.id"
       @click:close="$emit('close', tag)"
       close
-      class="mx-2">
+      class="readonly mx-2">
```
underscope: These ones are not readonly (actionable)

client/components/catalog/Card/Tags/index.vue
```diff
@@ -7,7 +7,7 @@
         @click:close="showDeleteConfirmation(id, name)"
         color="blue-grey lighten-4"
         label light small close
-        class="mr-2">
+        class="readonly mr-2">
```
underscope: These ones are not readonly (actionable)

underscope (changes requested):
Remove from other actionable items aswell 😉 :bowtie:

client/components/catalog/Card/Tags/index.vue
ikovac: Correct me if I'm wrong. My view of it was that if a chip is not clickable it should not have `readonly` class. A chip may contain clickable actions, but the chip itself is not clickable. 
For example, let's look at these two chips:

1. This chip is not clickable but it contains clickable action:
Screenshot 2020-11-24 at 11 01 17
2. This chip is clickable and it doesn't have `readonly` class:
Screenshot 2020-11-24 at 11 01 55

underscope: You have documentation/style guide here for each usage type (Right sidebar)
https://material.io/components/chips#input-chips

ikovac: @underscope Thanks for your comment.
All chips in the MD specification have a hover effect, so I wasn't really sure is that relevant. 
Here, in MD implementation: https://material.angular.io/components/chips/examples even basic chips got a hover effect.
I'm going to fix this and remove the `readonly` class from actionable chips 👍

bubafinder:
As I can see, the color of the x is changed
image

underscope:
@markolovric That's ok, that's an action

underscope:
Thx @ikovac 👌

PR #685, opened 2020-11-25
Added new logo

underscope:
- Compact version same as full
- Missing favicon


underscope:
@markolovric Missing favicon and compact version

PR #688, opened 2020-11-26
Sidebar CE comments

lukasaric:
- This PR enables CE comments in editor/outline sidebar.
- Also implements routing & toggling CE discussions by clicking indicator button within sidebar comment


client/components/common/tce-core/Events/DiscussionEvent.js
```diff
@@ -1,5 +1,6 @@
 export default {
   SAVE: 'comment:save',
   REMOVE: 'comment:remove',
-  SET_LAST_SEEN: 'comment:setLastSeen'
+  SET_LAST_SEEN: 'comment:setLastSeen',
+  TOGGLE: 'comments:toggle'
```
underscope: Try to be a bit more descriptive. It is not a general toggle.

client/components/common/tce-core/Discussion/Comment.vue
```diff
@@ -1,81 +1,143 @@
...
-    </v-avatar>
-    <div class="comment-body pl-3">
-      <div class="header">
-        <span class="author">
-          {{ author.fullName || author.email }}
-          <v-icon v-if="isEdited" size="16" class="ml-1 pb-1">
-            mdi-pencil-outline
-          </v-icon>
+  <div>
+    <div v-if="showUnseenSeparator" class="unseen-separator">
```
underscope: Unseen separator should be within thread component

client/components/common/tce-core/Discussion/Comment.vue
```diff
@@ -1,81 +1,143 @@
...
+          </v-tooltip>
+          <v-tooltip left>
+            <template v-slot:activator="{ on }">
+              <v-btn
+                v-if="elementTag"
+                v-on="on"
+                @click="toggleElementDiscussion"
+                color="teal"
+                text x-small>
+                Element
```
underscope: When viewed within the list, to me it looks like it relates to the element below. To me it feels like it should be part of header.

client/components/common/tce-core/Discussion/Comment.vue
```diff
@@ -1,81 +1,143 @@
...
+              <span v-on="on">
+                <timeago
+                  :datetime="comment.createdAt"
+                  :auto-update="60"
+                  class="time" />
+              </span>
+            </template>
+            <span>{{ comment.createdAt | formatDate('M/D/YY h:mm A') }}</span>
+          </v-tooltip>
+          <v-tooltip left>
```
underscope: Extract (ElementLink ?) Not sure if it should toggle ? 🤔

client/components/common/tce-core/Discussion/Comment.vue
```diff
@@ -1,81 +1,143 @@
...
-          </v-icon>
+  <div>
+    <div v-if="showUnseenSeparator" class="unseen-separator">
+      <v-divider class="my-3" />
+      <v-chip @click="showSeenMarker = !showSeenMarker" small outlined>
+        Unseen
+      </v-chip>
+      <div class="d-flex justify-center">
+        <transition name="slide-fade">
+          <v-btn
```
underscope: Should be part of thread not comment

client/components/common/tce-core/Discussion/Thread.vue
```diff
@@ -19,7 +20,10 @@ export default {
     items: { type: Array, required: true },
     user: { type: Object, required: true },
     showAll: { type: Boolean, default: false },
-    minDisplayed: { type: Number, default: 5 }
+    minDisplayed: { type: Number, default: 5 },
+    showAllComments: { type: Boolean, default: false },
```
underscope: showAll and showAllComments ? 🤔

client/components/common/tce-core/Discussion/Comment.vue
```diff
@@ -1,81 +1,143 @@
...
+            color="teal"
+            text x-small
+            class="seen-marker">
+            <v-icon class="mr-1" x-small>mdi-check</v-icon>
+            Mark All as Seen
+          </v-btn>
+        </transition>
+      </div>
+    </div>
+    <li class="comment">
```
underscope: List item outside of list

client/components/common/tce-core/Discussion/Comment.vue
lukasaric: I sent some new UI/UX proposal changes in offline discussion.

client/components/common/tce-core/Discussion/Thread/ElementLink.vue
```diff
@@ -0,0 +1,58 @@
...
+      if (!isEditor) return this.$router.push(editorRoute);
+      if (isSameRoute) this.$router.replace({ query: {} });
+      this.$router.push({ query: editorRoute.query });
+    }
+  }
+};
+</script>
+
+<style lang="scss" scoped>
+.v-btn {
```
underscope: Positioning should be controlled within parent

client/components/common/tce-core/Discussion/Thread/ElementLink.vue
```diff
@@ -0,0 +1,58 @@
...
+    <span>Go to element</span>
+  </v-tooltip>
+</template>
+
+<script>
+import find from 'lodash/find';
+
+export default {
+  name: 'element-link',
+  inject: ['$teRegistry'],
```
underscope: Label should be passed as a prop, no need to inject registry here, pass from element wrapper.

client/components/common/tce-core/Discussion/Thread/ElementLink.vue
```diff
@@ -0,0 +1,58 @@
...
+  props: {
+    activityId: { type: Number, required: true },
+    contentElement: { type: Object, default: () => ({}) },
+    contentElementId: { type: Number, default: null }
```
underscope: Why both?

client/components/common/tce-core/Discussion/Thread/ElementLink.vue
```diff
@@ -0,0 +1,58 @@
...
+  },
+  computed: {
+    elementLabel() {
+      const { type } = this.contentElement;
+      return find(this.$teRegistry._registry, { type })?.name;
+    },
+    isSameRoute: vm => vm.contentElement.uid === vm.$route.query.elementId,
+    isEditor: vm => vm.$route.name === 'editor',
```
underscope: If the parent is passing this, it is only executed once. Would like to move away in these types of components from hooking to router (should be done only within top level containers if possible)

client/components/common/tce-core/Discussion/Thread/ElementLink.vue
```diff
@@ -0,0 +1,58 @@
...
+      name: 'editor',
+      params: { activityId },
+      query: { elementId: contentElement.uid }
+    })
+  },
+  methods: {
+    linkToElement() {
+      const { isEditor, editorRoute, isSameRoute } = this;
+      if (!isEditor) return this.$router.push(editorRoute);
+      if (isSameRoute) this.$router.replace({ query: {} });
```
underscope: Why twice (replace + push). Relict of toggle?

client/components/common/tce-core/Discussion/Thread/index.vue
```diff
@@ -0,0 +1,57 @@
+<template>
+  <ul class="discussion-thread">
+    <div v-for="(comment, index) in visibleItems" :key="comment.uid">
```
underscope: Wrong markup under ul

client/components/common/tce-core/Discussion/Thread/index.vue
```diff
@@ -0,0 +1,57 @@
...
+        @markSeen="$emit('markSeen')"
+        v-bind="{ comment, user, unseenComments, seenMarker }"
+        :is-first-comment="index === 0" />
+      <thread-comment
+        @update="onUpdate"
+        @remove="$emit('remove', comment)"
+        v-bind="{ comment, user }">
+        <template v-slot:element-link>
+          <element-link
+            v-if="containAllComments && comment.contentElementId"
```
underscope: containAllComments 🤔 ? `isActivityThread` ?

client/components/common/tce-core/Discussion/Thread/Comment.vue
```diff
@@ -108,12 +117,14 @@ export default {
...
   display: flex;
-  margin-bottom: 1.25rem;
+  position: relative;
+  margin-bottom: 0.75rem;
   font-family: Roboto, Arial, sans-serif;
 
   &-avatar {
     width: 2.5rem;
     margin-top: 0.375rem;
+    overflow: hidden;
```
underscope: 🤔

client/components/common/tce-core/Discussion/Thread/Comment.vue
```diff
@@ -108,12 +117,14 @@ export default {
 <style lang="scss" scoped>
 .comment {
   display: flex;
-  margin-bottom: 1.25rem;
+  position: relative;
+  margin-bottom: 0.75rem;
```
underscope: Control from outside 🤔 ?

client/components/common/tce-core/Discussion/Thread/Comment.vue
```diff
@@ -35,6 +39,7 @@
         <span>{{ comment.createdAt | formatDate('M/D/YY h:mm A') }}</span>
       </v-tooltip>
     </div>
+    <slot name="element-link"></slot>
```
underscope: What's the purpose / why not just include boolean prop for visibility toggle (since slot will be used only for that).

client/components/common/tce-core/Discussion/Thread/Separator.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+  <div class="separator-container">
+    <div v-if="showUnseenSeparator" class="unseen-separator">
```
underscope: Should be handled from outside

client/components/common/tce-core/Discussion/Thread/Separator.vue
```diff
@@ -0,0 +1,67 @@
...
+      <v-divider class="my-3" />
+      <v-chip @click="showSeenMarker = !showSeenMarker" small outlined>
```
underscope: Rounded btn?

client/components/common/tce-core/Discussion/Thread/Separator.vue
```diff
@@ -0,0 +1,67 @@
...
+        Unseen
+      </v-chip>
+      <div class="d-flex justify-center">
+        <transition name="slide-fade">
+          <v-btn
+            v-if="showSeenMarker"
```
underscope: Should the whole block be under v-if ?

client/components/common/tce-core/Discussion/Thread/Separator.vue
```diff
@@ -0,0 +1,67 @@
...
+            @click="$emit('markSeen')"
+            color="teal"
+            text x-small
+            class="seen-marker">
+            <v-icon class="mr-1" x-small>mdi-check</v-icon>
```
underscope: x-small should be inherited/size set by x-small btn?

client/components/common/tce-core/Discussion/index.vue
```diff
@@ -24,18 +24,21 @@
       v-if="thread.length"
       @update="$emit('update', $event)"
       @remove="$emit('remove', $event)"
+      @markSeen="$emit('markSeen')"
       :items="thread"
       :user="user"
-      :min-displayed="commentsShownLimit"
       :show-all="showAll"
+      :min-displayed="commentsShownLimit"
+      :contain-all-comments="containAllComments"
```
underscope: Confusing name... 🤔

client/components/common/tce-core/Discussion/Thread/Separator.vue
```diff
@@ -0,0 +1,67 @@
...
+    <v-divider v-else-if="!isFirstComment" class="mt-0 mb-2 mx-1" />
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'thread-separator',
+  props: {
+    comment: { type: Object, required: true },
+    unseenComments: { type: Array, default: () => [] },
```
underscope: No need to pass this

client/components/common/tce-core/Discussion/Thread/Separator.vue
```diff
@@ -0,0 +1,67 @@
...
+    seenMarker: { type: Boolean, default: false },
+    isFirstComment: { type: Boolean, required: true },
+    user: { type: Object, required: true }
+  },
+  data: () => ({ showSeenMarker: false }),
+  computed: {
+    isAuthor: vm => vm.user.id === vm.comment.author.id,
+    showUnseenSeparator() {
+      const { comment, seenMarker, unseenComments, isAuthor } = this;
+      return seenMarker && !isAuthor && unseenComments[0]?.id === comment.id;
```
underscope: Isn't it easier for the unseenComments figure out on the upper level first unseen and pass what component needs to do 🤔 ?

client/store/modules/repository/comments/mutations.js
```diff
@@ -8,12 +8,16 @@ import {
   update
 } from '@/store/helpers/mutations';
 
-const markSeenComments = (state, { activityUid, elementUid, lastCommentAt }) => {
+const markSeenComments = ({ seen }, payload) => {
+  const { activityUid, elementUid, lastCommentAt, unseenElementComments } = payload;
```
underscope: unseenElementComments 🤔 ? whats the need for this one?

client/store/modules/repository/comments/index.js
```diff
@@ -6,7 +6,8 @@ const state = {
   items: {},
   seen: {
     activity: {},
-    contentElement: {}
+    contentElement: {},
+    elementComments: []
```
underscope: Why separate store?

client/components/common/tce-core/ElementDiscussion.vue
```diff
@@ -93,9 +96,14 @@ export default {
...
     setLastSeen(timeout) {
-      const { uid: elementUid, lastCommentAt } = this;
-      const options = { elementUid, lastCommentAt, timeout };
-      this.editorBus.emit(DiscussionEvent.SET_LAST_SEEN, options);
+      const { uid: elementUid, lastCommentAt, unseenComments, events } = this;
+      const options = {
+        elementUid,
+        lastCommentAt,
+        timeout,
+        unseenElementComments: unseenComments
```
underscope: Why not only timestamp of last comment?

client/components/editor/ActivityContent/index.vue
```diff
@@ -73,7 +73,10 @@ export default {
       const { elements, seen, activityId } = this;
       return transform(elements, (acc, it) => {
         const comments = this.getComments({ activityId, contentElementId: it.id });
-        const lastSeen = seen.contentElement[it.uid] || 0;
+        const lastSeen = {
```
underscope: Why do we need both, isn't for an element only relevant the bigger out of two?

client/components/editor/ActivityContent/index.vue
```diff
@@ -139,8 +142,9 @@ export default {
     scrollToElement(id, timeout = 500) {
       setTimeout(() => {
         const elementId = `#element_${id}`;
-        const element = this.$refs.activityContent.querySelector(elementId);
-        element.scrollIntoView();
+        const element = this.$refs.activityContent?.querySelector(elementId);
```
underscope, line 144: When is the scrollToElement called prior to render? (`?.`)

client/components/editor/ActivityContent/index.vue
```diff
@@ -170,6 +174,12 @@ export default {
       [[removeSelection, false], [isSelected, true]].forEach(([items, isSelected]) => {
         items.forEach(({ elementId, ...user }) => this.selectElement(elementId, user, isSelected));
       });
+    },
+    $route(route) {
```
underscope: This is prob resulting with prev check? Need to wait for the mount.

client/components/common/tce-core/Discussion/index.vue
```diff
@@ -57,20 +60,24 @@ export default {
...
+    scrollTarget: { type: String, default: 'discussion' },
+    unseenComments: { type: Array, default: () => [] },
+    seenMarker: { type: Boolean, default: false }
   },
   data: () => ({ showAll: false, comment: initCommentInput() }),
   computed: {
     thread: vm => orderBy(vm.comments, ['createdAt'], ['asc']),
     commentsCount: vm => vm.thread.length,
     showAllToggle: vm => vm.commentsShownLimit < vm.thread.length,
+    isPostDisabled: vm => !vm.comment.content?.trim(),
```
underscope: Describe what computed is rather than what it results with for this use case; i.e. isCommentInputEmpty / bit to large, try to figure out better one...

client/components/repository/common/ActivityDiscussion.vue
```diff
@@ -4,30 +4,38 @@
       @save="saveComment"
       @update="saveComment"
       @remove="remove"
-      v-bind="{ comments, user, showHeading, scrollTarget: 'editor' }"
-      show-notifications />
+      @markSeen="setLastSeenComment"
+      v-bind="{ comments, user, showHeading, unseenComments, seenMarker }"
```
underscope: If both have to resolve unseen comments just pass the comments array and last seen timestamp and then resolve within the thread.

client/components/common/tce-core/Discussion/Thread/Comment.vue
```diff
@@ -58,15 +63,19 @@
...
-import TextEditor from './TextEditor';
+import TextEditor from '../TextEditor';
 
 export default {
   name: 'thread-comment',
   props: {
     comment: { type: Object, required: true },
-    user: { type: Object, required: true }
+    user: { type: Object, required: true },
+    containAllComments: { type: Boolean, default: false }
```
underscope: Leftover?

client/components/common/tce-core/Discussion/Thread/ElementLink.vue
```diff
@@ -0,0 +1,58 @@
+<template>
+  <v-tooltip left>
+    <template v-slot:activator="{ on }">
+      <v-btn
+        v-on="on"
+        @click="linkToElement"
+        :color="isSameRoute ? 'secondary' : 'teal'"
```
underscope: isShown ?

client/components/common/tce-core/Discussion/Thread/ElementLink.vue
```diff
@@ -0,0 +1,58 @@
...
+        text x-small>
+        {{ elementLabel }}
+        <v-icon x-small class="ml-1">mdi-arrow-top-right-thick</v-icon>
+      </v-btn>
+    </template>
+    <span>Go to element</span>
```
underscope: View/Show element

client/components/common/tce-core/Discussion/Thread/ElementLink.vue
```diff
@@ -0,0 +1,58 @@
...
+
+<script>
+import find from 'lodash/find';
+
+export default {
+  name: 'element-link',
+  inject: ['$teRegistry'],
+  props: {
+    activityId: { type: Number, required: true },
+    contentElement: { type: Object, default: () => ({}) },
```
underscope: Why default to `{}`

client/components/common/tce-core/Discussion/Thread/index.vue
```diff
@@ -0,0 +1,57 @@
+<template>
+  <ul class="discussion-thread">
+    <div v-for="(comment, index) in visibleItems" :key="comment.uid">
+      <thread-separator
+        @markSeen="$emit('markSeen')"
+        v-bind="{ comment, user, unseenComments, seenMarker }"
+        :is-first-comment="index === 0" />
```
underscope: Can be solved by moving later (Probably does not make sense to show unseen separator if all are unseen ?)

client/components/common/tce-core/Discussion/Thread/Separator.vue
```diff
@@ -0,0 +1,67 @@
...
+            color="teal"
+            text x-small
+            class="seen-marker">
+            <v-icon class="mr-1" x-small>mdi-check</v-icon>
+            Mark All as Seen
+          </v-btn>
+        </transition>
+      </div>
+    </div>
+    <v-divider v-else-if="!isFirstComment" class="mt-0 mb-2 mx-1" />
```
underscope: *

client/components/common/tce-core/Discussion/Thread/index.vue
```diff
@@ -0,0 +1,57 @@
...
+
+export default {
+  name: 'discussion-thread',
+  props: {
+    items: { type: Array, required: true },
+    showAll: { type: Boolean, default: false },
+    minDisplayed: { type: Number, default: 5 },
+    unseenComments: { type: Array, default: () => [] },
+    containAllComments: { type: Boolean, default: false },
+    seenMarker: { type: Boolean, default: false },
```
underscope: seen-confirmation

client/components/common/tce-core/ElementDiscussion.vue
```diff
@@ -70,10 +71,12 @@ export default {
...
     unseenComments() {
       const { comments, user, lastSeen } = this;
-      return comments.filter(it => (
-        it.author.id !== user.id &&
-        new Date(it.createdAt).getTime() > lastSeen
-      ));
+      return comments.filter(it => {
+        const createdAt = new Date(it.createdAt).getTime();
+        return it.author.id !== user.id &&
+          createdAt > lastSeen.element &&
```
underscope: Should be only one .....

client/components/editor/ActivityContent/index.vue
lukasaric: It's not called prior to render. It's sometimes called after the component has been unmounted (e.g, when you open editor with `elementId` query and navigate away before `isLoading` & its timeout are resolved/done. )
The only alternative way I know to this approach would be to preserve `timeoutId` and clear it on `beforeDestroy`.
@underscope thoughts? 🤔

client/components/common/tce-core/Discussion/Thread/Comment/index.vue
```diff
@@ -0,0 +1,112 @@
...
+      <comment-header
+        :author="author"
+        :is-edited="isEdited"
+        :created-at="comment.createdAt" />
+      <text-editor
+        v-model="content"
+        :is-focused="isEditing"
+        :show-preview="!isEditing"
+        class="content" />
+      <span v-if="isEditing" class="float-left">
```
underscope: Buttons should be consistently aligned to the right (like post comment icon; see comment on element link if that's the problem). Because of the btn padding (seen when hoverer)
image
buttons look misaligned

image
please fix this either by lowering the padding or changing the btn style to depressed (only the second one in this case; once aligned to the right).

underscope: Consider using d-flex justify-end instead of float.

client/components/common/tce-core/Discussion/Thread/Comment/index.vue
```diff
@@ -0,0 +1,112 @@
...
+        <v-btn @click="reset" text small>Cancel</v-btn>
+        <v-btn
+          @click="save"
+          :disabled="content && !content.trim()"
```
underscope: Should we support deleting comments by editing and saving? If so, this check needs to be removed (i.e. slack allows it)

client/components/common/tce-core/Discussion/Thread/Comment/index.vue
```diff
@@ -0,0 +1,112 @@
+<template>
+  <div class="comment">
+    <v-avatar size="34" class="comment-avatar">
+      <img :src="author.imgUrl">
+    </v-avatar>
+    <div class="comment-body pl-3">
+      <comment-header
```
underscope: Comment header should be outside of the comment body 🤔  ?
This also limits the input width and seems a bit too small
image
Consider expanding it (right; empty space; can be still a bit padded on the right but not this much)

client/components/common/tce-core/Discussion/Thread/Comment/index.vue
```diff
@@ -0,0 +1,112 @@
+<template>
+  <div class="comment">
+    <v-avatar size="34" class="comment-avatar">
```
underscope: Avatar, user label, comment date, and actions should be part of a header.

client/components/common/tce-core/Discussion/Thread/Comment/index.vue
```diff
@@ -0,0 +1,112 @@
...
+};
+</script>
+
+<style lang="scss" scoped>
+.comment {
+  display: flex;
+  font-family: Roboto, Arial, sans-serif;
+
+  &-avatar {
+    width: 2.5rem;
```
underscope: Avatar size 🤔 ?

client/components/common/tce-core/Discussion/Thread/Comment/index.vue
```diff
@@ -0,0 +1,112 @@
...
+    margin-top: 0.375rem;
+    overflow: hidden;
```
underscope: 🤔 ?

client/components/common/tce-core/Discussion/Thread/Comment/Header.vue
```diff
@@ -0,0 +1,51 @@
...
+<script>
+export default {
+  name: 'comment-header',
+  props: {
+    author: { type: Object, required: true },
+    isEdited: { type: Boolean, default: false },
+    createdAt: { type: [Number, String], required: true }
+  },
+  computed: {
+    authorLabel: vm => vm.author.fullName || vm.author.email
```
underscope: Author should have label field attached (see user model)

client/components/common/tce-core/Discussion/Thread/ElementLink.vue
```diff
@@ -0,0 +1,47 @@
+<template>
```
underscope: After seeing the implications of positioning this in the footer (need to switch cancel/save on edit on left) I'd like to move this to the header after the `timeago` date and add a vertical separator or some wording between; i.e _about 2 hours ago on **HTML** element_  with HTML colored and linked using our teal accent color.

client/components/common/tce-core/Discussion/Thread/ElementLink.vue
```diff
@@ -0,0 +1,47 @@
...
+  <div class="element-link">
+    <v-tooltip left>
+      <template v-slot:activator="{ on }">
+        <v-btn
+          v-on="on"
+          @click="linkToElement"
+          :color="isShown ? 'teal' : 'blue-grey'"
+          text x-small>
+          {{ elementLabel }}
+          <v-icon x-small class="ml-1">mdi-arrow-top-right-thick</v-icon>
```
underscope: btn is already x-small?

client/components/common/tce-core/Discussion/Thread/ElementLink.vue
```diff
@@ -0,0 +1,47 @@
...
+  name: 'editor',
+  params: { activityId },
+  query: { elementId: contentElement.uid }
+});
+
+export default {
+  name: 'element-link',
+  props: {
+    activityId: { type: Number, required: true },
+    contentElement: { type: Object, required: true },
```
underscope: Only uid is used

client/components/common/tce-core/Discussion/Thread/ElementLink.vue
```diff
@@ -0,0 +1,47 @@
...
+    elementLabel: { type: String, required: true },
```
underscope: Consider renaming to label (Since component name is element link)

client/components/common/tce-core/Discussion/Thread/UnseenSeparator.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+  <div class="unseen-separator">
+    <v-divider />
+    <v-chip
+      @click="$emit('markSeen')"
```
underscope: Just seen?

client/components/common/tce-core/Discussion/Thread/UnseenSeparator.vue
```diff
@@ -0,0 +1,46 @@
...
+      @click:close="$emit('markSeen')"
+      close-icon="mdi-close"
+      color="teal"
+      small close outlined>
```
underscope: Group style props i.e. `outlined small close`

client/components/common/tce-core/Discussion/Thread/index.vue
```diff
@@ -0,0 +1,128 @@
+<template>
+  <ul v-intersect="onIntersect" class="discussion-thread">
+    <li
+      v-for="comment in visibleItems"
```
underscope: Consider aligning iter var name with collection name (i.e. rename comment into item or visibleItems to visibleComments)

client/components/common/tce-core/Discussion/Thread/index.vue
```diff
@@ -0,0 +1,128 @@
...
+        ref="unseen-separator"
+        @markSeen="markSeen"
+        :unseen-comments-count="unseenThread.length" />
+      <v-divider v-else class="thread-separator" />
+      <thread-comment
+        @update="onUpdate"
+        @remove="$emit('remove', comment)"
+        v-bind="{ comment, user }"
+        class="mb-3" />
+      <element-link
```
underscope: Should be slotted or part of the component

client/components/common/tce-core/Discussion/Thread/index.vue
```diff
@@ -0,0 +1,128 @@
...
+        v-if="isActivityThread && comment.contentElement"
+        v-bind="comment"
+        :is-editor="isEditor"
+        :element-label="elementLabel(comment)" />
```
underscope: Method name should start with a verb.

client/components/common/tce-core/Discussion/Thread/index.vue
```diff
@@ -0,0 +1,128 @@
...
+    items: { type: Array, required: true },
+    showAll: { type: Boolean, default: false },
+    minDisplayed: { type: Number, default: 5 },
+    isActivityThread: { type: Boolean, default: false },
+    unseenComments: { type: Array, required: true },
+    user: { type: Object, required: true }
+  },
+  data: () => ({ isVisible: false }),
+  computed: {
+    isEditor: vm => vm.$route.name === 'editor',
```
underscope: Should be decoupled from router (you can pass this as a prop). Might not be needed / see  ElementLink component.

client/components/common/tce-core/Discussion/Thread/index.vue
```diff
@@ -0,0 +1,128 @@
...
+    visibleItems: vm => vm.showAll ? vm.items : takeRgt(vm.items, vm.minDisplayed),
+    unseenThread: vm => orderBy(vm.unseenComments, 'createdAt', 'asc'),
+    firstUnseenComment: vm => vm.unseenThread[0]
```
underscope: Might be nicer if you extract li listing and then do a v-if for unseen block with divider (rather than doing the check for each item and having this firstUnseen logic.

client/components/common/tce-core/Discussion/Thread/index.vue
```diff
@@ -0,0 +1,128 @@
...
+  },
+  methods: {
+    onUpdate(comment, content) {
+      this.$emit('update', { ...comment, content, updatedAt: Date.now() });
+    },
+    elementLabel({ contentElement: { type } }) {
+      return find(this.$teRegistry._registry, { type })?.name;
+    },
+    showUnseenSeparator({ id, author }) {
```
underscope: Not needed if one of the prev comments is adopted (li extraction).

client/components/common/tce-core/ElementDiscussion.vue
```diff
@@ -25,7 +26,8 @@
       @save="save"
       @update="save"
       @remove="editorBus.emit(events.REMOVE, $event)"
-      v-bind="{ comments, user }"
+      @markSeen="setLastSeen"
+      v-bind="{ comments, user, unseenComments }"
```
underscope: comments, unseenComments, user (group related)

client/store/modules/repository/comments/getters.js
```diff
@@ -7,12 +8,15 @@ export const getComments = state => params => {
...
-  const lastSeen = seen.activity[activity.uid] || 0;
-  return filter(items, it =>
-    it.contentElementId === null &&
-    it.activityId === activity.id &&
-    it.authorId !== auth.user.id &&
-    new Date(it.createdAt).getTime() > lastSeen
-  );
+  const { items, seen } = state;
+  const lastActivitySeen = get(seen.activity, activity.uid, 0);
+  return filter(items, it => {
```
underscope: Filtering by activity first might save some additional computation and simplify the filter (early return; rather then doing it later on in if; also consider using getComments helper to make it a bit more readable rather than doing it all here)

client/components/common/tce-core/Discussion/Thread/Comment/Header.vue
```diff
@@ -0,0 +1,51 @@
+<template>
+  <div class="header">
+    <div class="d-block">
+      <v-tooltip right>
+        <template v-slot:activator="{ on }">
+          <span v-on="authorLabel.length > 25 ? on : null" class="author">
+            {{ authorLabel | truncate(25) }}
```
underscope: Since this will be a one-liner just use text-truncate class and the tooltip can be always shown (not a big value in doing it conditionally; filters removed in new vue version so let's avoid it when we can).

client/components/common/tce-core/Discussion/Thread/Comment/index.vue
```diff
@@ -0,0 +1,112 @@
...
+        </v-btn>
+      </span>
+    </div>
+    <div v-if="showOptions" class="actions">
+      <v-btn
+        v-for="{ name, action, icon } in options"
+        :key="name"
+        @click="action"
+        x-small icon
+        class="mr-1">
```
underscope: ml-1 seems like a way to go since we want to use space on the right as much and need to distance actions on the left.

client/store/modules/repository/comments/getters.js
```diff
@@ -7,12 +8,15 @@ export const getComments = state => params => {
...
-  const { seen, items } = state;
-  const lastSeen = seen.activity[activity.uid] || 0;
-  return filter(items, it =>
-    it.contentElementId === null &&
-    it.activityId === activity.id &&
-    it.authorId !== auth.user.id &&
-    new Date(it.createdAt).getTime() > lastSeen
-  );
+  const { items, seen } = state;
+  const lastActivitySeen = get(seen.activity, activity.uid, 0);
```
underscope: lastActivitySeen name implies the last activity that was seen rather then specific activity last seen date

client/store/modules/repository/comments/getters.js
```diff
@@ -7,12 +8,15 @@ export const getComments = state => params => {
...
+  return filter(items, it => {
+    const isCurrentActivityComment = it.activityId === activity.id;
+    const isAuthor = it.authorId === auth.user.id;
+    const createdAt = new Date(it.createdAt).getTime();
+    if (!isCurrentActivityComment || isAuthor || lastActivitySeen >= createdAt) return;
+    if (!it.contentElement) return true;
```
underscope, line 17: Add comment why this is being done

client/components/common/tce-core/Discussion/Thread/ElementLink.vue
```diff
@@ -0,0 +1,47 @@
...
+    isEditor: { type: Boolean, required: true }
```
underscope: Would like to remove this entirely, possibly by emitting an event rather than doing the navigation within this page (coupled at the moment).

client/components/common/tce-core/Discussion/Thread/ElementLink.vue
```diff
@@ -0,0 +1,47 @@
...
+  },
+  computed: {
+    isShown: vm => vm.contentElement.uid === vm.$route.query.elementId
```
underscope: This can be probably removed (See comment above).

client/components/common/tce-core/Discussion/Thread/UnseenSeparator.vue
```diff
@@ -0,0 +1,46 @@
...
+</script>
+
+<style lang="scss" scoped>
+.unseen-separator {
+  text-align: center;
+
+  ::v-deep .v-chip.v-chip--outlined.v-chip {
+    margin: -1rem 0 0.5rem 0;
+    border-radius: 1rem !important;
+    background-color: #fafafa !important;
```
underscope: setting grey lighten-5 class doesn't do the trick?

client/components/common/tce-core/Discussion/Thread/index.vue
underscope: Instead of method you can just figure out index of firstUnseen

client/components/common/tce-core/Discussion/Thread/index.vue
```diff
@@ -0,0 +1,128 @@
...
+      const isAuthor = user.id === author.id;
+      return !isAuthor && firstUnseenComment?.id === id;
+    },
+    onIntersect(_entries, _observer, isIntersected) {
+      this.isVisible = isIntersected;
+    },
+    scrollToFirstUnseen() {
+      this.$emit('showAll', true);
+      this.$nextTick(() => {
+        const element = this.$refs['unseen-separator'][0].$el;
```
underscope: use camelCase for refs

client/components/common/tce-core/Discussion/Thread/index.vue
```diff
@@ -0,0 +1,128 @@
...
+    markSeen() {
+      this.$emit('markSeen');
+      this.$emit('showAll', false);
+    }
+  },
+  watch: {
+    isVisible(val) {
+      const { unseenThread, minDisplayed } = this;
+      if (!val || !unseenThread.length) return;
+      if (unseenThread.length < minDisplayed) return;
```
underscope: since repeated it can be moved to scrollToFirstUnseen (consider updating the method name aswell since it does more than that)

client/components/common/tce-core/Discussion/Thread/index.vue
```diff
@@ -0,0 +1,128 @@
...
+
+  .thread-separator {
+    margin: 0 0.25rem 0.5rem 0.25rem;
+  }
+
+  &:first-child .thread-separator {
+    display: none;
+  }
+
+  ::v-deep .element-link .v-btn {
```
underscope: Can be removed, see prev comment.

client/components/editor/ActivityContent/index.vue
```diff
@@ -70,11 +71,11 @@ export default {
     ...mapState({ user: state => state.auth.user }),
     activityId: vm => vm.activity.id,
     processedElements() {
-      const { elements, seen, activityId } = this;
+      const { elements, seen, activity, activityId } = this;
```
underscope: no need to pick up activityId if activity is being included ?

client/components/editor/ActivityContent/index.vue
```diff
@@ -70,11 +71,11 @@ export default {
...
       return transform(elements, (acc, it) => {
         const comments = this.getComments({ activityId, contentElementId: it.id });
-        const lastSeen = seen.contentElement[it.uid] || 0;
-        acc[it.uid] = { ...it, comments, lastSeen };
+        const allSeen = [seen.contentElement[it.uid], seen.activity[activity.uid]];
```
underscope: no need to prefix with all (seen === allSeen). In this case, it is not even allSeen so I would rather do max on this line and have the old name back. Thoughts? 🤔

client/components/editor/ActivityContent/index.vue
underscope: No need to change it for now

client/components/editor/ActivityContent/index.vue
```diff
@@ -170,6 +172,12 @@ export default {
       [[removeSelection, false], [isSelected, true]].forEach(([items, isSelected]) => {
         items.forEach(({ elementId, ...user }) => this.selectElement(elementId, user, isSelected));
       });
+    },
+    $route(route) {
+      const { elementId } = route.query;
+      if (!elementId) return;
+      this.selectElement(elementId);
```
underscope: We should somehow reuse this section within isLoading watcher and document when this is being used.

client/components/common/tce-core/Discussion/index.vue
```diff
@@ -24,18 +24,21 @@
...
       @update="$emit('update', $event)"
       @remove="$emit('remove', $event)"
+      @markSeen="$emit('markSeen')"
+      @showAll="showAll = $event"
       :items="thread"
       :user="user"
-      :min-displayed="commentsShownLimit"
       :show-all="showAll"
+      :min-displayed="commentsShownLimit"
+      :unseen-comments="unseenComments"
```
underscope: align naming (items and unseen-comments). Should be grouped as well.

client/components/common/tce-core/Discussion/index.vue
```diff
@@ -54,17 +57,27 @@ export default {
...
+    user: { type: Object, required: true }
   },
   data: () => ({ showAll: false, comment: initCommentInput() }),
   computed: {
-    thread: vm => orderBy(vm.comments, ['createdAt'], ['asc']),
+    thread() {
+      const { comments, unseenComments } = this;
+      const processedThread = comments.map(comment => {
+        const unseen = unseenComments.find(it => it.id === comment.id);
+        return { ...comment, unseen: !!unseen };
```
underscope, line 78: No need to pass unseenArray if on the processed array you are passing it as attr (you can just pass items and rely on this for other usages aswell, much cleaner approach; you can also get rid of passing count and last seen dates later on)

client/components/editor/ActivityContent/index.vue
lukasaric: I agree. We can also rename `allSeen` to maybe `seenUids`, I think it's descriptive enough. Gather all seen `uids`, then assign `max` timestamp to lastSeen?

client/components/common/tce-core/Discussion/Thread/ElementLink.vue
lukasaric: I think not. Icon size inside is decoupled from button size. Default size is medium I think. e.g (icon will not inherit `x-small` size from `btn`)

client/components/common/tce-core/Discussion/Thread/ElementLink.vue
kronicker: Or you could make this component a reusable `tce-core/EditorLink` component that always navigates to editor (with optional `elementId` param)?
Elements are only shown in the editor so you don't actually need to care about the environment where the link is used and using `router-link` will take care of possible duplicated navigation errors.

client/components/common/tce-core/Discussion/Thread/UnseenSeparator.vue
lukasaric: No it does not. It will also apply color on border.

client/components/common/tce-core/Discussion/TextEditor.vue
```diff
@@ -5,7 +5,7 @@
     </div>
     <v-textarea
       v-else
-      @change="$emit('input', $event)"
+      @input="$emit('input', $event)"
```
kronicker: Think about proxying events and attributes and using `v-model.trim` in the parent

client/store/modules/repository/comments/getters.js
```diff
@@ -1,18 +1,22 @@
...
-  const lastSeen = seen.activity[activity.uid] || 0;
-  return filter(items, it =>
-    it.contentElementId === null &&
-    it.activityId === activity.id &&
-    it.authorId !== auth.user.id &&
-    new Date(it.createdAt).getTime() > lastSeen
-  );
+export const getUnseenActivityComments = ({ seen }, getters, { auth }) => activity => {
+  const activityComments = getters.getComments({ activityId: activity.id });
+  const lastActivitySeenDate = get(seen.activity, activity.uid, 0);
```
kronicker: `activitySeenAt`

client/store/modules/repository/comments/getters.js
```diff
@@ -1,18 +1,22 @@
...
+  return filter(activityComments, it => {
+    const isAuthor = it.authorId === auth.user.id;
+    const createdAt = new Date(it.createdAt).getTime();
+    if (isAuthor || lastActivitySeenDate >= createdAt) return;
+    if (!it.contentElement) return true;
+    // Return unseen activity comment if contentElement is not set
+    const lastElementSeenDate = get(seen.contentElement, it.contentElement.uid, 0);
```
kronicker: `elementSeenAt`

client/components/editor/ActivityContent/index.vue
```diff
@@ -139,23 +141,31 @@ export default {
...
       const { elementId } = this.$route.query;
-      if (val || !elementId) return;
+      if (!elementId) return;
       // Select and scroll to element if elementId is set
       setTimeout(() => {
         this.selectElement(elementId);
         this.scrollToElement(elementId);
+        if (!multipleSelections) return;
         this.collaboratorSelections
           .forEach(({ elementId, ...user }) => this.selectElement(elementId, user));
```
kronicker: This shouldn't be in this function (semantically)

client/components/editor/ActivityContent/index.vue
```diff
@@ -70,11 +71,12 @@ export default {
...
     activityId: vm => vm.activity.id,
     processedElements() {
-      const { elements, seen, activityId } = this;
+      const { elements, seen } = this;
+      const { id: activityId, uid: activityUid } = this.activity;
       return transform(elements, (acc, it) => {
         const comments = this.getComments({ activityId, contentElementId: it.id });
-        const lastSeen = seen.contentElement[it.uid] || 0;
-        acc[it.uid] = { ...it, comments, lastSeen };
+        const seenUids = [seen.contentElement[it.uid], seen.activity[activityUid]];
```
kronicker: `lastSeen = max(...`

client/components/common/tce-core/Discussion/index.vue
```diff
@@ -54,17 +56,28 @@ export default {
...
   computed: {
-    thread: vm => orderBy(vm.comments, ['createdAt'], ['asc']),
+    thread() {
+      const { comments, unseenComments } = this;
+      const unseenCount = unseenComments.length;
+      const processedThread = comments.map(comment => {
+        const isUnseen = !!unseenComments.find(it => it.id === comment.id);
+        return { ...comment, isUnseen, unseenCount };
+      });
+      return orderBy(processedThread, ['isUnseen', 'createdAt'], 'asc');
```
kronicker: Return object with `comments` array prop and `unseenCount`

kronicker (changes requested):
To be continued... 😉

client/components/common/tce-core/Discussion/TextEditor.vue
```diff
@@ -1,12 +1,12 @@
 <template>
```
kronicker: This can be functional compoment 😉

client/components/common/tce-core/Discussion/Thread/Comment/index.vue
```diff
@@ -0,0 +1,84 @@
...
+      this.isEditing = !this.isEditing;
+    },
+    save() {
+      const { comment, content } = this;
+      if (!content) return this.$emit('remove', comment);
+      this.toggleEdit();
+      this.$emit('update', comment, content);
+    },
+    remove() {
+      this.$emit('remove', this.comment);
```
kronicker, line 50: I'd say this is good behavior but I would add a confirmation modal for this since it's not obvious.

client/components/common/tce-core/Discussion/Thread/Comment/index.vue
```diff
@@ -0,0 +1,84 @@
...
+    content: vm.comment.content,
+    isEditing: false
+  }),
+  methods: {
+    toggleEdit() {
+      this.isEditing = !this.isEditing;
+    },
+    save() {
+      const { comment, content } = this;
+      if (!content) return this.$emit('remove', comment);
```
kronicker: Either use method or inline emit in template

client/components/common/tce-core/Discussion/Thread/index.vue
```diff
@@ -0,0 +1,92 @@
...
+    }
+  },
+  methods: {
+    onUpdate(comment, content) {
+      this.$emit('update', { ...comment, content, updatedAt: Date.now() });
+    },
+    onIntersect(_entries, _observer, isIntersected) {
+      this.isVisible = isIntersected;
+    },
+    revealUnseen(unseenCount) {
```
kronicker: `count`

client/components/common/tce-core/Discussion/index.vue
```diff
@@ -64,24 +66,27 @@ export default {
...
-        return { ...comment, isUnseen, unseenCount };
+        const unseen = unseenComments.find(it => it.id === comment.id);
+        return { ...comment, unseen: !!unseen };
       });
-      return orderBy(processedThread, ['isUnseen', 'createdAt'], 'asc');
+      return orderBy(processedThread, ['unseen', 'createdAt'], 'asc');
     },
     commentsCount: vm => vm.thread.length,
-    showAllToggle: vm => vm.commentsShownLimit < vm.thread.length,
+    showAllToggle: vm => vm.commentsShownLimit < vm.commentsCount,
```
kronicker: `hasHiddenComments` or something along those lines

kronicker (changes requested):

kronicker (approved):
Nice job! LGTM 👌 
@underscope As far as I'm concerned this can be merged.

PR #689, opened 2020-11-26
Move JWT to cookie

kronicker:
### This PR:
- moves JWT access token from local storage to HTTP only, secure and signed cookie
- adds `/me` route to check access and load user info on app boot
- removes user data from local storage

Subtask of #632.

#### QA Note: (cc @kjuej @markolovric)
Full regression focused on authentication would be preferable since this is impacting authentication on the whole platform.
N.B.: For this to work properly locally please make sure that your `.env` values are set to
```env
HOSTNAME=localhost
PROTOCOL=http
```
Also, please test with both scenarios - setting (new) `AUTH_JWT_COOKIE_SECRET` env value to any string or leaving it empty (this will make cookies signed or unsigned respectively).


server/shared/auth/authenticator.js
```diff
@@ -17,10 +21,46 @@ class Auth extends Authenticator {
...
+  _afterAuthenticate(setCookie) {
+    return function ({ user }, res, next) {
+      if (!setCookie) return next();
+      const token = user.createToken({
+        audience: Audience.Scope.Access,
+        expiresIn: '5 days'
+      });
+      const { name, ...options } = config.jwt.cookie;
+      res.cookie(config.jwt.cookie.name, token, {
+        ...options,
```
underscope: Options like secret are not used / not sure if we care, want to filter them out and pass exact options for readability sake.

server/shared/auth/authenticator.js
```diff
@@ -17,10 +21,46 @@ class Auth extends Authenticator {
...
+        maxAge: 5 * 1000 * 60 * 60 * 24 // 5 days
```
underscope: 🗡️ 🤡 
Consider updating order to make it more readable `5 x 24 x 60 x 60 x 1000` ?

server/shared/auth/authenticator.js
```diff
@@ -17,10 +21,46 @@ class Auth extends Authenticator {
...
+    const onUserAuthenticated = this._afterAuthenticate(options.setCookie);
+    return function (req, res, next) {
+      authenticateUser(req, res, function (err) {
+        if (arguments.length > 0) return next(err);
+        onUserAuthenticated(req, res, next);
+      });
+    };
+  }
+
+  _afterAuthenticate(setCookie) {
```
underscope: Since this is all about setting the cookie, consider renaming the callback to something more along those lines to make it more readable here and previously

server/shared/auth/authenticator.js
```diff
@@ -17,10 +21,46 @@ class Auth extends Authenticator {
...
+      options = {};
+    }
+    options = options || {};
     // NOTE: Setup passport to forward errors down the middleware chain
     // https://github.com/jaredhanson/passport/blob/ad5fe1df/lib/middleware/authenticate.js#L171
-    return super.authenticate(strategy, { ...options, failWithError });
+    if (options.failWithError !== false) options.failWithError = true;
+    const authenticateUser = super.authenticate(strategy, options, callback);
+    const onUserAuthenticated = this._afterAuthenticate(options.setCookie);
+    return function (req, res, next) {
```
underscope: See if you can write this a bit less convoluted (maybe extract callback into named function / moving it)

server/shared/auth/authenticator.js
```diff
@@ -17,10 +21,46 @@ class Auth extends Authenticator {
     return super.initialize(options);
   }
 
-  authenticate(strategy, { failWithError = true, ...options } = {}) {
+  authenticate(strategy, options, callback) {
+    if (isFunction(options)) {
+      callback = options;
```
underscope: This shifting is not that nice :bowtie:  Consider rather spreading the stuff after strategy and then determining

client/store/modules/auth/mutations.js
```diff
@@ -1,12 +1,7 @@
...
-  state.token = token;
-};
-
-export const logout = state => {
-  Object.assign(state, { user: null, token: null });
 };
 
-export const setUser = (state, user) => {
-  state.user = user;
+export const resolveLoading = state => {
```
underscope: ?

server/shared/auth/authenticator.js
kronicker: This is c/p from [passport source code](https://github.com/jaredhanson/passport/blob/master/lib/middleware/authenticate.js#L71) but I agree that it can be polished 👍

underscope:
@kjuej Please take a look

underscope (approved):

Issue #399, opened 2019-11-21
Unable to scroll through the page when dragging an element in tailor


kjuej:
## I'm submitting a...

- [x] Bug report  
- [ ] Regression (a behavior that used to work and stopped working in a new release)
- [ ] Feature request
- [ ] Documentation issue or request

## Context

Usually when i drag the element on the top/bottom of the page, the page would automatically scroll up/down.

![PageScroll](https://user-images.githubusercontent.com/34127147/69321051-a8f19180-0c42-11ea-9d68-5743f80a26e0.gif)

PR #692, opened 2020-11-30
Update editor toolbar positioning in DOM and hide element toolbar when element is being dragged

hrvojevu:
### This PR:
Updates editor's toolbar position to `static` which resolves the bug described in issue https://github.com/ExtensionEngine/tailor/issues/399. Bug occurred due to toolbar overlapping `ElementList` component which handles `sortablejs` auto scroll feature when sorting items. Since toolbar overlapped the component, `sortablejs` wouldn't register mouse position and it wouldn't scroll up. Additionally, element toolbar is now hidden when element is being dragged using the element focus event.

### DIFF
--- a/client/components/common/tce-core/ElementList.vue
+++ b/client/components/common/tce-core/ElementList.vue
@@ -1,8 +1,8 @@
 <template>
   <div class="list-group">
     <draggable
-      @start="dragElementIndex = $event.oldIndex"
-      @end="dragElementIndex = -1"
+      @start="onDragStart"
+      @end="onDragEnd"
       @update="reorder"
       :list="elements"
       v-bind="options"
@@ -76,6 +76,14 @@ export default {
   methods: {
     get,
     getElementId,
+    onDragStart({ oldIndex }) {
+      this.dragElementIndex = oldIndex;
+      this.$emit('drag', true);
+    },
+    onDragEnd() {
+      this.dragElementIndex = -1;
+      this.$emit('drag', false);
+    },
     reorder({ newIndex: newPosition }) {
       const items = this.elements;
       this.$emit('update', { newPosition, items });
--- a/client/components/common/tce-core/ElementList.vue
+++ b/client/components/common/tce-core/ElementList.vue
@@ -1,8 +1,8 @@
 <template>
   <div class="list-group">
     <draggable
-      @start="onDragStart"
-      @end="onDragEnd"
+      @start="dragElementIndex = $event.oldIndex"
+      @end="dragElementIndex = -1"
       @update="reorder"
       :list="elements"
       v-bind="options"
@@ -10,8 +10,8 @@
       <div
         v-for="(element, index) in elements"
         :key="getElementId(element)"
-        @dragstart="dragElementIndex = index"
-        @dragend="dragElementIndex = -1"
+        @dragstart="onDragStart(index)"
+        @dragend="onDragEnd(element)"
         :class="`col-xs-${get(element, 'data.width', 12)}`">
         <slot
           :element="element"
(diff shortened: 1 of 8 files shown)


client/components/common/tce-core/ElementList.vue
```diff
@@ -48,6 +48,9 @@ import Draggable from 'vuedraggable';
 import get from 'lodash/get';
 import { getElementId } from 'tce-core/utils';
 import last from 'lodash/last';
+import { mapChannels } from '@/plugins/radio';
+
+const CE_FOCUS_EVENT = 'element:focus';
```
underscope, line 53: Not an action item for now since it might result in some bigger conflicts, but at some point, we should start extracting these to have a single listing/overview of used events and the ability to easily change.

client/components/editor/ActivityContent/index.vue
```diff
@@ -161,6 +162,6 @@ export default {
 }
 
 .loader {
-  margin-top: 4.375rem;
+  margin-top: 1.25rem;
```
underscope: You have a vertical movement now when the skeleton loader switches to containers (should look like the same sheet)

client/components/editor/ActivityContent/index.vue
```diff
@@ -148,8 +148,9 @@ export default {
 
 <style lang="scss" scoped>
 .activity-content {
+  flex-grow: 1;
   min-height: 100%;
-  padding: 4.375rem 1.5625rem 0 26.25rem;
```
underscope, line 152: We lost spacing on lower resolution =>
image

client/components/editor/Toolbar/index.vue
```diff
@@ -63,9 +63,11 @@ export default {
 
 <style lang="scss" scoped>
 .toolbar-wrapper {
-  position: fixed;
   width: 100%;
-  z-index: 99;
+
+  ::v-deep .element-toolbar-wrapper {
```
underscope: Since the element-toolbar-wrapper is just a toolbar resolver and this one is an element for placing it within the editor, consider keeping the positioning within this one.

underscope (changes requested):

client/components/editor/ActivityContent/index.vue
```diff
@@ -148,8 +148,10 @@ export default {
...
+  padding: 4.375rem 2.5rem 0 1.5625rem;
+  padding-top: 1.25rem;
```
underscope: ? Leftover

client/components/editor/Toolbar/index.vue
```diff
@@ -63,9 +63,12 @@ export default {
...
+    position: absolute;
```
underscope: Not sure if by mistake but prev comment hasn't been resolved.

underscope:
In addition to the previous comments, there are some new issues like horizontal movement upon focusing on different elements. Consider keeping the old layout and just implementing the focus out upon drag on this one 😉 
[Screen Recording 2020-12-04 at 16.56.14.mov.zip](https://github.com/ExtensionEngine/tailor/files/5643826/Screen.Recording.2020-12-04.at.16.56.14.mov.zip)

hrvojevu:
@underscope horizontal movement issue was produced by sidebar changing width due to flex-basis not being set on activity content and sidebar. That has been fixed now and there is no more issues when focusing elements.

client/components/editor/index.vue
```diff
@@ -73,8 +79,19 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-.editor-container {
+$sidebar-width: 400px;
```
underscope: => rems

client/components/editor/index.vue
```diff
@@ -73,8 +79,19 @@ export default {
...
+
+.editor-content-container {
   display: flex;
-  flex-direction: column;
+  height: calc(100% - 50px);
```
underscope: ☝️

PR #695, opened 2020-12-03
Serve storage assets through CloudFront

MiroDojkic:
### This PR
- enables storage access using CloudFront
- fallbacks to own proxy endpoint: `GET /proxy/:key`
- access is verified with signed cookies
- cookies allow access to defined resources, i.e. `repository/access`
- cookies are obtained on repository endpoints: `ALL /repository/:id`


server/app.js
```diff
@@ -1,11 +1,13 @@
 'use strict';
 
 const bodyParser = require('body-parser');
+const cookieParser = require('cookie-parser');
 const cors = require('cors');
 const express = require('express');
 const helmet = require('helmet');
 const origin = require('./shared/origin');
 const path = require('path');
+const proxy = require('./shared/storage/proxy');
```
underscope: Please rename to `storageProxy` to be more exact (there could be other types of proxies in this type of file, includes readability in the following lines ;) )

client/api/asset.js
```diff
@@ -1,9 +1,14 @@
 import request from './request';
 
 const urls = {
-  base: '/assets'
+  base: '/assets',
+  setCookies: '/assets/set-cookies'
```
underscope: Does it make sense to be even more specific (i.e. set-access-cookies)

client/api/asset.js
```diff
@@ -1,9 +1,14 @@
...
 };
 
+function setCookies() {
```
underscope: Same as prev; feels a bit too generic when called from outside (what kind of cookies 🤔 )

config/server/index.js
```diff
@@ -15,6 +16,7 @@ const previewUrl = process.env.PREVIEW_URL;
 module.exports = {
   protocol,
   hostname,
+  localhost,
```
underscope: If flag, please keep the is prefix. Move on the first place(protocol, hostname, port looks like expected order/grouping)

config/server/storage.js
```diff
@@ -10,5 +11,16 @@ module.exports = {
   filesystem: {
     path: process.env.STORAGE_PATH
   },
-  provider: process.env.STORAGE_PROVIDER
+  proxy: {
+    provider: process.env.STORAGE_PROXY,
```
underscope: Default here to proxy for backwards comp?

server/shared/storage/helpers.js
```diff
@@ -100,7 +101,7 @@ async function resolveAssetsMap(element) {
     if (!url) return set(element.data, key, url);
     const isStorageResource = url.startsWith(STORAGE_PROTOCOL);
     const resolvedUrl = isStorageResource
-      ? (await storage.getFileUrl(url.substr(STORAGE_PROTOCOL.length, url.length)))
+      ? proxy.getFileUrl(url.substr(STORAGE_PROTOCOL.length, url.length))
```
underscope, line 104: Omit async from f def.

server/shared/storage/helpers.js
```diff
@@ -5,16 +5,17 @@ const { elementRegistry } = require('../content-plugins');
 const get = require('lodash/get');
 const isString = require('lodash/isString');
 const isUrl = require('is-url');
+const config = require('../../../config/server').storage;
 const mime = require('mime-types');
 const Promise = require('bluebird');
+const proxy = require('./proxy');
```
underscope, line 11: ✅ Here (within storage helpers)

server/shared/storage/storage.controller.js
```diff
@@ -16,21 +15,10 @@ async function upload({ file }, res) {
   const hash = sha256(file.originalname, buffer);
   const extension = path.extname(file.originalname);
   const name = path.basename(file.originalname, extension).substring(0, 180).trim();
-  const key = path.join(ASSET_ROOT, `${hash}___${name}${extension}`);
+  const key = path.join(config.path, `${hash}___${name}${extension}`);
```
underscope, line 18: Lets consider to start scoping these per repo.

server/shared/storage/proxy/mw.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+const { FORBIDDEN } = require('http-status-codes');
+const miss = require('mississippi');
+const path = require('path');
+const proxy = require('./');
+const storage = require('../');
+const config = require('../../../../config/server').storage;
```
underscope: Order

poi.config.js
```diff
@@ -37,6 +38,7 @@ const devServer = {
   },
   proxy: {
     '/api': { target: serverUrl },
+    ...(STORAGE_PROXY === 'local' ? { '/proxy': serverUrl } : {}),
```
underscope: Should the base URL for the proxy (local or cf) always be exposed in order to be able to construct asset url from client based on key?

underscope:
🏆 🎖️

client/api/asset.js
MiroDojkic: Cookies are now set on the response for the `/repository/:id` request, so I removed this as it is no longer used.

config/server/index.js
MiroDojkic: Removed this too since it's not used anymore.

underscope (approved):

PR #698, opened 2020-12-07
Move element delete button from element toolbar to the element

hrvojevu:
This PR:
- Removes element delete button from element toolbar and places it next to element itself

Due to low spacing near elements, I advise we use smallest button possible which doesn't take much of the horizontal spacing. This is especially problematic when button is displayed in compound elements, x axis overflow occurs if the button is positioned further away from the element or if it takes too much horizontal space.

Screen Shot 2020-12-08 at 2 54 23 PM


hrvojevu:
Tagging @underscope to discuss button examples.

hrvojevu:
Leaving both `work in progress` and `needs QA` labels since the functionality change is done but actual button design is up for discussion, tagging @kjuej and @markolovric .

bubafinder:
I suggest something like this. Background transparency on hover can be lower, then the button will be more noticeable

image

bubafinder:
@hrvojevu please double-check the delete feature inside the table element, after deleting content inside the cell, the whole table is missing. I don't know if this is related to this feature but you can check.

image

bubafinder:
Looks even better as you made it 👍 

image

PR #701, opened 2020-12-10
Enable silent auth flow and logout from IdP

### LINKED ISSUE #632: SSO

kronicker:
This PR:
- [x] Adds silent OIDC token refresh/auth on app boot
- [x] Adds logout from IdP if OIDC session is active
- [x] Enables non-standard post logout URL key
- [x] Keeps support for both local and OIDC login

Tested with:
- [x] Auth0
- [x] Okta
- [x] Google

Closes #632


server/oidc/index.js
```diff
@@ -14,17 +16,52 @@ const OIDCErrors = [
...
+const getSilentAuthParams = req => {
+  const strategy = req.passport.strategy('oidc');
+  const callbackUri = new URL(strategy.options.callbackURL);
+  callbackUri.searchParams.set('silent', 'true');
+  return { redirect_uri: callbackUri.href };
+};
+
 const isOIDCError = err => OIDCErrors.some(Ctor => err instanceof Ctor);
 
 router
```
lukasaric, line 39: It would nice to extract all callbacks, at least as separate functions and then call it for each route.

server/shared/auth/oidc.js
```diff
@@ -28,7 +28,15 @@ module.exports = class OIDCStrategy extends BaseOIDCStrategy {
 
   logoutUrl(params = {}) {
     const { client } = this;
-    return client.endSessionUrl({ ...params, client_id: client.client_id });
+    const url = new URL(client.endSessionUrl({
+      ...params,
+      client_id: client.client_id
+    }));
+    const customRedirectUriKey = this.options.postLogoutUriKey;
+    if (!this.options.postLogoutUriKey) return url.href;
```
lukasaric:
```suggestion
    if (!customRedirectUriKey) return url.href;
```

client/OidcClient.js
```diff
@@ -0,0 +1,75 @@
+import path from 'path';
+
+const isProduction = process.env.NODE_ENV === 'production';
+const SILENT_REFRESH_TIMEOUT = 5000;
+
+class OidcClient {
+  constructor() {
+    this.enabled = process.env.OIDC_ENABLED;
+    this.baseUrl = path.join(process.env.API_PATH, '/oidc');
```
lukasaric: `'oidc'` is enough. You don't need backslash prefix since`path.join` automatically joins all string with `/`.

client/OidcClient.js
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  logout() {
+    window.location.replace(this._getLogoutUrl());
+  }
+
+  slientlyRefresh() {
+    return new Promise((resolve, reject) => {
+      const iframe = window.document.createElement('iframe');
+      iframe.style.visibility = 'hidden';
```
lukasaric: It would be nice to extract these `iframe` style props/options.

client/OidcClient.js
```diff
@@ -0,0 +1,75 @@
...
+const SILENT_REFRESH_TIMEOUT = 5000;
+
+class OidcClient {
+  constructor() {
+    this.enabled = process.env.OIDC_ENABLED;
+    this.baseUrl = path.join(process.env.API_PATH, '/oidc');
+    this._silentRefreshTimeout = null;
+  }
+
+  _getSilentUrl() {
```
lukasaric: Consider rewriting url `getter` methods (silent, resign & logout url) as  class getters

client/api/request.js
```diff
@@ -29,7 +29,7 @@ client.auth = new EventEmitter();
 
 client.interceptors.response.use(res => res, err => {
   if (err.response && [FORBIDDEN, UNAUTHORIZED].includes(err.response.status)) {
```
lukasaric: Consider extracting second check as variable with a proper naming.

client/router.js
```diff
@@ -131,7 +130,7 @@ export default function getRouter() {
     if (to.matched.some(it => it.meta.auth) && !auth.user) {
       return next({ path: '/login', query: { redirect: to.fullPath } });
     }
-    if (!isAllowed(to)) {
+    if (!isAllowed(to, auth)) {
```
lukasaric:
```suggestion
     return !isAllowed(to, auth) ? next({ path: from.fullPath }) : next();
```
Also you could write it in multiline ternary for for better readability.

client/router.js
```diff
@@ -131,7 +130,7 @@ export default function getRouter() {
     if (to.matched.some(it => it.meta.auth) && !auth.user) {
```
lukasaric: You could extract first condition as small on liner function with `route` arg. Also with proper naming, something like `requiresAuth`.

lukasaric (changes requested):

client/OidcClient.js
```diff
@@ -0,0 +1,99 @@
+import { EventEmitter } from 'events';
+import path from 'path';
+
+const isProduction = process.env.NODE_ENV === 'production';
+const SILENT_REFRESH_TIMEOUT = 5000;
+
+class OidcClient {
+  constructor() {
+    this.enabled = process.env.OIDC_ENABLED;
+    this.baseUrl = path.join(process.env.API_PATH, 'oidc');
```
abasic: Can this be done without importing the node.js 'path' module on the client side?

client/store/modules/auth/actions.js
```diff
@@ -22,8 +22,8 @@ export const resetPassword = (_, { token, password }) => {
 
 export const fetchUserInfo = ({ commit }) => {
   return api.getUserInfo()
-    .then(({ data: { user } }) => commit('setUser', user))
-    .catch(() => commit('setUser', null));
+    .then(({ data: { user, authStrategy } }) => commit('setAuth', { user, authStrategy }))
+    .catch(() => commit('resetAuth', null));
```
abasic: No need for the `null` here.

client/api/auth.js
```diff
@@ -30,7 +30,7 @@ function changePassword(currentPassword, newPassword) {
 }
 
 function getUserInfo() {
-  return request.get(urls.profile);
+  return request.base.get(urls.profile);
```
abasic, line 33: What is the motivation behind this change? I assume it is to avoid the `auth` interceptor because of the fetch when the store is initiated. But I am wondering if there was a different approach since it seems to me to be fixing the symptom instead of the cause.

server/oidc/index.js
```diff
@@ -14,17 +15,52 @@ const OIDCErrors = [
...
   .get('/', (req, res, next) => {
-    const isResign = req.query.resign === 'true';
-    const params = { scope, ...isResign && { prompt: 'login' } };
+    const strategy = req.passport.strategy('oidc');
+    if (isLogoutRequest(req)) return strategy.logout()(req, res, next);
+    const params = {
+      scope,
+      ...getPromptParams(req),
+      ...isSilentAuth(req) && getSilentAuthParams(req)
```
abasic: Can this be changed so that `getSilentAuthParams` call `isSilentAuth` and then returning an empty object if it is not? If it can not then there should at least be brackets around this. It is not easy to follow what is happening and what has priority.

abasic (changes requested):

client/OidcClient.js
kronicker: We already have `path` used for `extname` in `tce-video` so it's not anything new in the bundle and it's safer to build paths with `path#join`. Do you have an idea of how to do it without it on the client side? 🤔

client/api/auth.js
kronicker: The motivation behind introducing `client#base` (by @vladimyr I think) was to have a _pure_ Axios client with no interceptors attached to be used for auth calls like `login`. 
`getUserInfo` is primarily auth checking endpoint now that the only way to check auth and get user profile is using server so I see it as a good candidate for `client#base`.
I don't see that as a symptom per se, but intended behavior. I'm interested in how do you think this should be taken care of? Are you concerned that this is used for both profile fetch and auth check or that this is a slightly different API from other methods in this file (except `login`) that might not be so obvious?

client/OidcClient.js
abasic: I think we used `url-join` in these cases. I can not for the life of me remember the entire discussion and with whom I had it but it was about avoiding using Node.js modules on the client-side. But, if it is already used on the client-side then keep it.

client/api/auth.js
abasic: This then might be more of a philosophical question. If this is the only time we are calling this API endpoint and technically the user does not need to be authenticated to access it, should it be returning `FORBIDDEN` or `UNAUTHORIZED`? Should it return something else like `BAD_REQUEST` or even `OK` since we do not require them to be authenticated at that point. We just use the authentication flow to get the user if they are authenticated.

client/OidcClient.js
kronicker: You're right, I haven't check for `url-join`, but now I see that it's more widely used. It's better not to reinforce the incorrect convention by introducing another precedent so I'll change it to `url-join`.

client/api/auth.js
kronicker: Hmm, actually this is not the same case as `login` now that you mention it. Login is the one where authentication is not needed or to be more precise where `jwt` strategy is not used but authentication is done via basic strategy instead.
`/me` endpoint uses our _regular_ `jwt` strategy and does require authentication and I'm leaning on that fact in implementation. The only difference between `/login` and `/me` and other endpoints is that we have the handling in place for unauthorized errors and don't need an interceptor to play global authorization error handler. The issue is that interceptor comes before the error handling and not after so the only way to use it as a global error handler is to keep it as is and bypass it via `client#base` when not needed.
I wonder if it would be better to have a whitelist of routes where interceptor should not be triggered but I feel that would be a maintenance overhead and also I prefer this as it's more explicit and is usage driven (if `/me` is used on another place I don't want it to bypass interceptor).
@vladimyr @abasic Any thoughts, since I don't have a good and/or constructive idea here?

abasic: I wanted to raise the question just to see if it was considered. Maybe a better question should have been, should using the JWT strategy always mean that the user is being authenticated. If it does then it should stay as is. If not then we could look into a different approach for such routes but that might be a general discussion and outside of this PR. I do not want to hold this up too much. Maybe we can talk tomorrow during the stand up about this topic.

kronicker: After an offline discussion, we decided to leave it as is. 
In an ideal world, checking authentication would be decoupled from fetching the profile (since those are semantically different) and the former would avoid interceptor and handle `UNAUTHORIZED` error as one of the expected responses while the latter would be treated as any other request.
Since in the current state of the codebase, there is no other use-case for profile fetch and the new strategy of storing the JWT requires at least one API call on the app boot to check for user auth this currently makes the most sense. (Since it at the same time checks for auth and hydrates client with user profile which is also needed on the app boot if the user is authorized, but was previously pulled from local storage).

bubafinder:
1. Logout on Okta side
2. After logout click on the Tailor side, error presented

After manual back to the localhost URL, the user is still logged

image

markolovric:
1. Set OIDC_ALLOW_SIGNUP=0
2. Tried to log in through SSO provider
3. User is redirected back to Tailor login screen with an error message (screenshot) 
4. On the second try, I am asked to verify my Okta password (second screenshot)

image
image

kronicker:
@markolovric Should work properly now. Tested with okta, auth0, and google.

underscope:
@markolovric Works for me and @kronicker (dev + prod mode). There was a configuration issue on OKTA that I updated/fixed today (in case that was the provider you were testing with). If you are still having issues it is most likely local config issue; let's sync regarding it on Mon).

@kronicker App hangs a bit upon start due to auth init behind the scenes. Please revisit and see how we can improve it.

PR #702, opened 2020-12-11
Implement publish notification system

hrvojevu:
### This PR:
- implements publish notification system using webhooks
- authorization is done using oauth [Client Credentials Flow](https://auth0.com/docs/flows#client-credentials-flow)

### Note:
Node version needs to be updated to `12.13.0` (LTS) or higher since the `simple-oauth2` [v4 requires Node 12.x or higher](https://github.com/lelylan/simple-oauth2#requirements)


.env.example
```diff
@@ -103,3 +103,10 @@ FORCE_COLOR=1
 # Do not group published resources by outline item id.
 # repository/id/resourceId* instead of repository/id/outlineId/resourceId*
 FLAT_REPO_STRUCTURE=1
+
+# Consumer publish notification settings
+CONSUMER_WEBHOOK_URL=
+CONSUMER_CLIENT_ID=
+CONSUMER_CLIENT_SECRET=
+CONSUMER_CLIENT_TOKEN_HOST=
+CONSUMER_CLIENT_TOKEN_PATH=
```
MiroDojkic, line 112: Are these used to construct URL to consumer API's token endpoint?
If so, we might be able to simplify configuration a bit by replacing these with a single `CONSUMER_CLIENT_TOKEN_URL` variable.

server/shared/publishing/oauthClient.js
```diff
@@ -0,0 +1,37 @@
...
+const { ClientCredentials } = require('simple-oauth2');
+const config = require('../../../config/server/consumer');
+const Deferred = require('../../shared/util/Deferred');
+const request = require('axios');
+
+const clientId = config.clientID;
+const clientSecret = config.clientSecret;
+const tokenHost = config.tokenHost;
+const tokenPath = config.tokenPath;
+const webhookUrl = config.webhookUrl;
```
MiroDojkic: Please consider destructuring these properties to reduce noise.

server/shared/publishing/oauthClient.js
```diff
@@ -0,0 +1,37 @@
...
+const client = new ClientCredentials({
+  client: { id: clientId, secret: clientSecret },
+  auth: { tokenHost, tokenPath }
+});
+const deferred = new Deferred();
+
+deferred.promise = getAccessToken();
+
+let accessToken = deferred.resolve();
```
MiroDojkic: What's the idea behind the deferred?
Maybe I'm missing something, but seems to me that `deferred.resolve()` will resolve promise that's created in Deferred constructor, resulting in undefined value being assigned to `accessToken`.
If that's the case, result of promise from `deferred.promise = getAccessToken()` is never used.

server/shared/publishing/oauthClient.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
```
MiroDojkic: Would a name `webhook.js` or `webhookProvider.js` describe more accurately what this module does?

server/shared/publishing/publishing.service.js
```diff
@@ -15,7 +16,8 @@ class PublishingService {
   }
 
   publishActivity(activity) {
-    return this.queue.add(() => publishActivity(activity));
+    return this.queue.add(() => publishActivity(activity)
+      .then(data => oauthClient.send(data) && data));
```
MiroDojkic: Sorry for nitpicking, but would this be a bit easier to read if we'd move callback with the promise chain to a variable?
For example
```suggestion
    const publish = () => publishActivity(activity)
      .then(data => oauthClient.send(data) && data);
    return this.queue.add(publish);
```

server/shared/publishing/oauthClient.js
MiroDojkic: Also, this whole module seems generic, so it might be worth considering to move it out of `publishing` module.

MiroDojkic (changes requested):
@hrvojevu left a couple of thoughts. Please check if any of them make sense. 🙃

.env.example
hrvojevu: These variables are used for `simple-oauth2` `ClientCredentials` configuration and this configuration requires `tokenHost` and `tokenPath` to be two properties.

config/server/consumer.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+const { env } = process;
+
+module.exports = {
+  webhookUrl: env.CONSUMER_WEBHOOK_URL,
+  clientID: env.CONSUMER_CLIENT_ID,
```
MiroDojkic: I believe the convention here would imply camelcase:
```suggestion
  clientId: env.CONSUMER_CLIENT_ID,
```
You wouldn't need to alias it when importing [here](https://github.com/ExtensionEngine/tailor/pull/702/files#diff-7efff118cb81130382f12499a92b5d65874901ddd644818c97a1cfc3596cee5aR4)

MiroDojkic (approved):
Left one small suggestion, otherwise seems good to me.

PR #703, opened 2020-12-14
Custom TCE hooks

MiroDojkic:
### This PR:
- deprecates content element's `handleStatics` function
- function is replaced with beforeSave, afterSave, afterRetrieve and afterLoaded hooks for content elements
- `tce-image` processor and resolver are moved to `tce-image/util` and replaced with hooks

Hooks are called when following conditions are met, in specific order:
- before create or update: beforeSave
- after create or update: afterSave, afterLoaded
- after fetch: afterRetrieve, afterLoaded

### DIFF
--- a/server/shared/content-plugins/elementRegistry.js
+++ b/server/shared/content-plugins/elementRegistry.js
@@ -2,6 +2,9 @@
 
 const BaseRegistry = require('./BaseRegistry');
 const elementsList = require('../../../config/shared/core-elements');
+const hooks = require('./elementHooks');
+const pick = require('lodash/pick');
+const services = require('../../shared/services.js');
 
 const EXTENSIONS_LIST = '../../../extensions/content-elements/index';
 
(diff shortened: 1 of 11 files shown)


server/shared/storage/helpers.js
```diff
@@ -47,40 +39,14 @@ function processQuestion(element) {
 
 function processPrimitive(primitive) {
   if (!isPrimitive(primitive)) throw new Error('Invalid primitive');
-  if (!processor[primitive.type]) return Promise.resolve(primitive);
-  return processor[primitive.type](primitive);
+  return Promise.resolve(primitive);
```
underscope: You can now entirely eliminate the 'default' processor.

MiroDojkic: ✅

server/shared/content-plugins/elementRegistry.js
```diff
@@ -31,3 +58,11 @@ class ElementsRegistry extends BaseRegistry {
 }
 
 module.exports = new ElementsRegistry();
+
+function deprecateHandleStatics(element) {
+  const name = `tce-${element.type.toLowerCase()}`;
```
underscope, line 63: toLowerCase won't be a way to go for multi-word types 😉

PR #704, opened 2020-12-15
Improve element drag and drop feature

### LINKED ISSUE #699: Improve element drag and drop feature
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request

## Context
It's currently quite hard to drag and drop and element between elements that are not in your current field of view:
DragAndDrop

hrvojevu:
This PR:
- adds light gray `#f4f5f5` background to dragged element drop area and hides the content inside
- sets `max-height` to `9.375rem` (150px) on dragged element drop area

### DIFF
--- a/client/components/common/tce-core/ElementList.vue
+++ b/client/components/common/tce-core/ElementList.vue
@@ -119,6 +119,7 @@ export default {
 
 ::v-deep .sortable-drag .content-element {
   max-height: auto;
+  background: #fff;
 }
 
 </style>
--- a/client/components/common/tce-core/ElementList.vue
+++ b/client/components/common/tce-core/ElementList.vue
@@ -69,9 +69,7 @@ export default {
     ...mapChannels({ editorChannel: 'editor' }),
     options: vm => ({
       ...vm.dragOptions,
-      handle: '.drag-handle',
-      scrollSpeed: 15,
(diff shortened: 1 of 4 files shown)


underscope (changes requested):
Question container element styling is impacted (white background)

image

client/components/common/tce-core/ElementList.vue
```diff
@@ -69,9 +69,7 @@ export default {
     ...mapChannels({ editorChannel: 'editor' }),
     options: vm => ({
       ...vm.dragOptions,
-      handle: '.drag-handle',
```
underscope, line 72: According to this => https://www.npmjs.com/package/vuedraggable#all-sortable-options
This is the current list of sortable options =>  https://github.com/SortableJS/sortablejs/blob/master/docs/sortable-options.md and these props are removed from here. They are part of AutoScroll plugin => https://github.com/SortableJS/sortablejs/tree/master/packages/plugins/auto-scroll so I guess we would need to fiddle a bit with it in order to set it.

underscope: It is not included by default => https://github.com/SortableJS/Vue.Draggable/blob/master/src/vuedraggable.js

PR #705, opened 2020-12-16
Fix handleStatics deprecation message

MiroDojkic:
### This PR:
- fixes type casing for handleStatics deprecation message


server/shared/content-plugins/elementRegistry.js
```diff
@@ -60,7 +61,7 @@ class ElementsRegistry extends BaseRegistry {
 module.exports = new ElementsRegistry();
 
 function deprecateHandleStatics(element) {
-  const name = `tce-${element.type.toLowerCase()}`;
+  const name = `tce-${toCase.slug(element.type)}`;
```
MiroDojkic, line 64: @underscope are we ok with this or should we use the same logic with question-specific type resolution as in `getComponentName` from `client/components/common/tce-core/utils`?

PR #710, opened 2020-12-16
Avoid compiling server files inside the custom content element

ikovac:
**This PR:**
- exclude server files inside the custom content element from webpack build
- edit eslint configuration
- import server hooks from the `extensions/custom-elements/element-name/server/index.js`

❗ **Breaking change**
- `tce/util.js` should be moved to `tce/server/index.js`
[Tce repo](https://github.com/ExtensionEngine/tailor-element-template) also needs to be updated.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,6 +14,7 @@ module.exports = {
   },
   overrides: [{
     files: ['extensions/**'],
+    excludedFiles: ['extensions/**/util.js'],
     parserOptions: {
       parser: '@babel/eslint-parser',
       sourceType: 'module'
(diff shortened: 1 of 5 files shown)


.eslintrc.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
   },
   overrides: [{
     files: ['extensions/**'],
+    excludedFiles: ['extensions/**/server/**/*'],
```
MiroDojkic: Is `extensions/**/server/**` sufficient?

MiroDojkic (changes requested):
This change should break/ignore hooks in `tce-image`. Please apply the new structure there and test it by checking if the image is fetched from proxy endpoint instead of base64.
Same applies to `tce-scorm`.

MiroDojkic (approved):

PR #714, opened 2021-01-04
Fixed issue with deleted containers and publish:

### LINKED ISSUE #400: Custom containers are not marked as deleted
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request

## Context
If you create a custom container and publish it, it will remain in the bucket even after it's been deleted, and has no deletedAt property.

Screen Shot 2019-11-25 at 16 23 23

abasic:
- After publishing, published files of deleted containers are removed
- After unpublishing, files are bulk deleted instead of one at a time
- Added deleteFiles method to storage providers
- Updated listFiles methods for storage providers to return the same data structure

Closes #400

### DIFF
--- a/server/shared/publishing/helpers.js
+++ b/server/shared/publishing/helpers.js
@@ -12,6 +12,7 @@ const {
   getSupportedContainers
 } = require('../../../config/shared/activities');
 const { containerRegistry } = require('../content-plugins');
+const difference = require('lodash/difference');
 const filter = require('lodash/filter');
 const find = require('lodash/find');
 const findIndex = require('lodash/findIndex');
@@ -43,8 +44,7 @@ function publishActivity(activity) {
     activity.publishedAt = new Date();
     addToSpine(spine, activity);
 
-    return cleanActivityFolder(repository, activity)
-      .then(() => publishContent(activity))
+    return publishContent(activity)
       .then(content => {
         const publishedData = find(spine.structure, { id: activity.id });
         return attachContentSummary(publishedData, content);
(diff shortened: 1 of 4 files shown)


server/shared/publishing/helpers.js
```diff
@@ -254,8 +278,8 @@ function defaultSummaryBuilder({ id, uid, type, publishedAs, elements = [] }) {
   return { id, uid, type, publishedAs, elementCount: elements.length };
 }
 
-function getContentContainerFilenames({ contentContainers = [] }) {
-  return map(contentContainers, it => `${it.id}.${it.publishedAs}`);
+function getContainerFilePaths(baseUrl, containers = []) {
```
kronicker: `getContainersFilePaths` since its array of containers?

server/shared/publishing/helpers.js
```diff
@@ -89,19 +95,21 @@ function unpublishActivity(repository, activity) {
...
-      spine.structure = filter(spine.structure, ({ id }) => !find(deleted, { id }));
-      return saveSpine(spine)
-        .then(savedSpine => updateRepositoryCatalog(repository, savedSpine.publishedAt))
-        .then(() => activity.save());
-    });
+    return Promise
+      .map(deleted, it => {
+        const baseUrl = getBaseUrl(repository.id, it.id);
+        const filePaths = getContainerFilePaths(baseUrl, it.contentContainers);
+
```
kronicker: I would say that this space is not necessary since the flow between the lines is consistent and obvious

server/shared/publishing/helpers.js
```diff
@@ -183,6 +195,18 @@ async function fetchCustomContainers(parent, config) {
...
+async function unpublishDeletedContainers(parent, containers) {
+  const baseUrl = getBaseUrl(parent.repositoryId, parent.id);
+
+  const filePaths = getContainerFilePaths(baseUrl, containers);
+  const publishedFilePaths = await storage.listFiles(baseUrl);
+
+  const redundantFilePaths = difference(publishedFilePaths, filePaths);
+  if (redundantFilePaths.length) await storage.deleteFiles(redundantFilePaths);
+
+  return containers;
```
kronicker: I think you would benefit in readability from the promise chain here rather than async/await with blank spaces.

kronicker: Also, the signature of this method is a bit misleading for me since it's named `unpublishDeletedContainers` while it receives published containers that are to be preserved and is returning that original set of containers as a result 🤔

abasic: > I think you would benefit in readability from the promise chain here rather than async/await with blank spaces.

Did you mean something like this:
```js
function unpublishDeletedContainers(parent, containers) {
  const baseUrl = getBaseUrl(parent.repositoryId, parent.id);
  const filePaths = getContainersFilePaths(baseUrl, containers);
  return storage
    .listFiles(baseUrl)
    .then(publishedFilePaths => {
      const redundantFilePaths = difference(publishedFilePaths, filePaths);
      if (redundantFilePaths.length) return storage.deleteFiles(redundantFilePaths);
    });
}
```

underscope: @kronicker ☝️ 👀

kronicker: @abasic 👍

server/shared/publishing/helpers.js
```diff
@@ -89,19 +95,20 @@ function unpublishActivity(repository, activity) {
...
-      spine.structure = filter(spine.structure, ({ id }) => !find(deleted, { id }));
-      return saveSpine(spine)
-        .then(savedSpine => updateRepositoryCatalog(repository, savedSpine.publishedAt))
-        .then(() => activity.save());
-    });
+    return Promise
+      .map(deleted, it => {
+        const baseUrl = getBaseUrl(repository.id, it.id);
+        const filePaths = getContainersFilePaths(baseUrl, it.contentContainers);
+        return storage.deleteFiles(filePaths);
```
underscope, line 104: @abasic Since this is not transactional should we save spine first in case some of the containers fail or something breaks repo is not corrupt from consumer end)? 🤔

abasic: If this is just a clean up and it is a nice to have that these are deleted then yes, we should move it after the entire activity is saved. We could even do a cron job or something of the sorts that makes sure everything is cleaned up. But, outside of this, at some point we should find a way to make the process of publishing rollback-able.

PR #716, opened 2021-01-14
Feature: Enable adding environment variables for TCE server hooks

ikovac:
- Catch onElementSave error and emit it back to the element
- Add tce namespace inside config object
- Fix async reducer

❗ This PR introduces a way of declaring `TCE` env variables. 
The `TCE` env variables should be prefixed with `TCE` in .env file.

**Example:**
1. Declare TCE env variable inside .env file
```
// .env file

TCE_BRIGHTCOVE_VIDEO_API_KEY=myapikey
```

2. Declared env variable can be found inside config object in `tce` namespace. Env variable is converted to [camelCase](https://lodash.com/docs/4.17.15#camelCase) and the `TCE` prefix is dropped.
```js
const config = require('config/server');

console.log(config.tce.brightcoveVideoApiKey); // Prints "myapikey"
```

3. TCE HOOKS example
```js
// tce-brightcove-video/server/index.js

function beforeSave(asset, { config }) {
	console.log(config.tce.brightcoveVideoApiKey); // Prints "myapikey"
	return asset;
}
```


client/components/content-elements/tce-api-video/server/index.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+
+const { afterLoaded, afterSave, beforeSave } = require('tce-apivideo/dist/server/index.js');
```
MiroDojkic: Can we import the whole module (assuming it only exports hooks) and then destructure it inside of exported object?
I.e.
```suggestion
const server = require('tce-apivideo/dist/server');
```

client/components/editor/ActivityContent/ContainerList.vue
```diff
@@ -94,10 +94,15 @@ export default {
...
-        this.showNotification();
-      });
+      return this.saveElement(element)
+        .then(() => {
+          this.$radio.channel(`element:${getElementId(element)}`).emit('saved');
+          this.showNotification();
+        })
+        .catch(err => {
+          this.$radio.channel(`element:${getElementId(element)}`)
+            .emit('error', err.response.data);
```
MiroDojkic: Does the element lose access to error's status and headers if we process it here?
If that's the case, it might be a good idea to emit the whole error object and process it within the element, otherwise we're setting ourselves for a breaking change in the future.

package.json
```diff
@@ -83,7 +83,7 @@
     "hashids": "^2.2.1",
     "helmet": "^3.21.1",
     "html-to-text": "^6.0.0",
-    "http-errors": "^1.7.3",
+    "http-errors": "^1.8.0",
```
MiroDojkic: Why was this bump needed?

.env.example
```diff
@@ -103,3 +103,7 @@ FORCE_COLOR=1
 # Do not group published resources by outline item id.
 # repository/id/resourceId* instead of repository/id/outlineId/resourceId*
 FLAT_REPO_STRUCTURE=1
+
+# api.video API key
+# https://api.video/
+API_VIDEO_API_KEY=
```
MiroDojkic: Might be sufficient to document this in the element's `README` to prevent bloating this file. 🤔

MiroDojkic (changes requested):

client/components/editor/ActivityContent/ContainerList.vue
ikovac: `err.response.data` contains status code, but headers will be lost. I will fix it 👍

.env.example
ikovac: It's already documented there. Here is just a mini note to inform users if they never heard of that content element. Should I delete notes from here?

config/server/tce.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+const camelCase = require('lodash/camelCase');
+const coreElements = require('../shared/core-elements');
+const extensionElements = require('../../extensions/content-elements');
```
MiroDojkic: This will throw if there's no `index.js` in `extensions/content-elements`.

client/components/editor/ActivityContent/ContainerList.vue
```diff
@@ -94,10 +94,15 @@ export default {
...
-      });
+      return this.saveElement(element)
+        .then(() => {
+          this.$radio.channel(`element:${getElementId(element)}`).emit('saved');
+          this.showNotification();
+        })
+        .catch(err => {
+          this.$radio.channel(`element:${getElementId(element)}`)
+            .emit('error', err.response);
+        });
```
MiroDojkic: We could reduce repetition here by factoring out ```this.$radio.channel(`element:${getElementId(element)}`)```

server/content-element/hooks.js
```diff
@@ -53,7 +54,7 @@ function add(ContentElement, Hooks, Models) {
   function customElementHook(hookType, element) {
     const elementHookTypes = elementHookMappings[hookType];
     if (!elementHookTypes) return;
-    return elementHookTypes
+    return Promise.resolve(elementHookTypes)
```
MiroDojkic, line 57: 🙇‍♂️

config/server/tce.js
```diff
@@ -0,0 +1,17 @@
...
+const regex = new RegExp('^TCE_(.*)');
+
+const tceConfig = Object.keys(envs)
+  .map(it => it.match(regex))
+  .filter(Boolean)
+  .map(([env, secret]) => ({ env, secret: camelCase(secret) }))
+  .reduce((config, { env, secret }) => ({
+    ...config,
+    [secret]: process.env[env]
+  }), {});
```
MiroDojkic: Please consider if any of the following changes make it a bit easier to read:
```suggestion
const tceConfig = Object.keys(process.env)
  .map(key => key.match(/^TCE_(.*)/))
  .filter(Boolean)
  .reduce((config, [prefixedKey, key]) => ({
    ...config,
    [camelCase(key)]: process.env[prefixedKey]
  }), {});
```

.env.example
```diff
@@ -103,3 +103,6 @@ FORCE_COLOR=1
 # Do not group published resources by outline item id.
 # repository/id/resourceId* instead of repository/id/outlineId/resourceId*
 FLAT_REPO_STRUCTURE=1
+
+# tce-api-video API key
+TCE_API_VIDEO_API_KEY=
```
MiroDojkic: As discussed, please remove tce specific secrets from `.env.example` and keep them in the tce's readme.

MiroDojkic (changes requested):

server/content-element/hooks.js
ikovac: 😂

MiroDojkic (approved):

config/server/tce.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
```
underscope, line 1: server root might not be the most appropriate place (don't want to pollute it over time). Why not part of common config (create folder and rename config.js => index.js)

underscope (changes requested):

underscope:
🎖️

PR #718, opened 2021-01-22
Role based guards resolution fix

mcapeta:
This PR fixes role based auth which was broken due to store resolution changes in #689.

QA note: navigating to admin section should now be enabled.


client/router.js
```diff
@@ -146,5 +146,6 @@ export default function getRouter() {
 
 function isAllowed(route) {
```
mcapeta, line 146: `hasAccess` instead of `isAllowed`? 😄

mcapeta:
Covered in #701.

Closing...

PR #719, opened 2021-01-25
Replace cloneDeep with immer

ikovac:
Improve performance by replacing `lodash/cloneDeep` with [immer](https://immerjs.github.io/immer/docs/introduction)

### DIFF
--- a/client/components/content-elements/tce-audio/edit/Toolbar.vue
+++ b/client/components/content-elements/tce-audio/edit/Toolbar.vue
@@ -19,8 +19,7 @@
 <script>
 import get from 'lodash/get';
 import InputAsset from '@/components/common/InputAsset';
-import produce from 'immer';
-import set from 'lodash/set';
+import { set } from '@/utils/reducers';
 
 export default {
   name: 'tce-audio-toolbar',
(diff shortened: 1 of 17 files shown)


zcuric:
Just a thought. Do we need a new utility library in the codebase? 🤔

ikovac:
Thanks for sharing your thoughts @zcuric. 
This library should improve render performance by replacing the heavy-computing clondeDeep function. 
Also, the code should be less error-prone because of immutability.
Example:
Current code:
```js
console.log(arg); // Outputs: { a: 1 }
function fun(arg) {
  arg.b = 5;
  return arg;
}
fun(arg);
console.log(arg); // Outputs: { a: 1, b: 5 }
```
New code:
```js
console.log(arg); // Outputs: { a: 1 }
function fun(arg) {
  return immer(arg, draft => { draft.b = 5; })
}
fun(arg);
console.log(arg); // Outputs: { a: 1 }
```

I hope I got it right, @MiroDojkic please correct me if I'm wrong.

MiroDojkic:
@zcuric I share your concern, even though it has potential benefits of reducing scripting time and re-renders.
We tried to simplify the introduced API, but it still seems too obtrusive for an unnoticeable gain, so I'm closing this one.

PR #720, opened 2021-01-26
Fix: Unable to access the admin page

ikovac:
The link to the admin page is broken.
Kapture 2021-01-26 at 09 37 29

This PR fixes mentioned issue.


mcapeta:
#718 😄 
Let's consult on our stand up meeting should we introduce new export or use async one 😉

PR #721, opened 2021-01-26
Truncate long schema names

mcapeta:
Yet another hacktoberfest PR 😂 

This PR truncates long schema name so they stack nicely with the cog and publishing info indicator.
Before:
before
After:
after

### DIFF
--- a/client/components/catalog/Card/index.vue
+++ b/client/components/catalog/Card/index.vue
@@ -122,6 +122,11 @@ export default {
       this.isTruncated = clientWidth < scrollWidth;
     }
   },
+  watch: {
+    '$vuetify.breakpoint.width'() {
+      this.detectTruncation();
+    }
+  },
   mounted() {
     this.$nextTick(() => this.detectTruncation());
   },
--- a/client/components/catalog/Card/index.vue
+++ b/client/components/catalog/Card/index.vue
@@ -8,9 +8,9 @@
       <div @click="navigateTo()" class="card-body">
         <div class="header ml-4">
           <v-chip :color="repository.data.color" x-small class="readonly px-1" />
-          <v-tooltip open-delay="300" top>
+          <v-tooltip :disabled="!isTruncated" open-delay="300" top>
             <template v-slot:activator="{ on }">
-              <span v-on="on" class="schema-name mx-2">{{ schema }}</span>
+              <span ref="schemaName" v-on="on" class="schema-name mx-2">{{ schema }}</span>
             </template>
             {{ schema }}
           </v-tooltip>
@@ -93,6 +93,7 @@ export default {
   props: {
     repository: { type: Object, required: true }
   },
+  data: () => ({ isTruncated: false }),
   computed: {
     name: ({ repository }) => repository.name,
     description: ({ repository }) => repository.description,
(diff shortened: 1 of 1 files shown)


zcuric:
@mcapeta v-tooltip to display full name on hover? 🤔

mcapeta:
> @mcapeta v-tooltip to display full name on hover? 🤔

@zcuric
Thanks for pointing me to display entire schema name if it's being truncated 😉 
Nevertheless, I don't like how if looks with the tooltip version for a few reasons:
 - tooltip is redundant for short schema names (we should somehow detect there is no text overflow to disable that 🤹 )
 - not looking good in my opinion (restyling the tooltip is an option, of course)
long tooltip

What I suggest here is the following:
expand-schema-name (1)

Cons of the suggested solution:
 - limited to max three lines (feel free to argue if you don't think that's enough, cc @markolovric)
 - transition implementation doesn't feel smooth with the CSS transition only so I leveraged `<v-hover/>` component, too

bubafinder:
I am totally for the first sample, tooltip on hover for text overflow

mcapeta:
> I am totally for the first sample, tooltip on hover for text overflow

Done! Please note that tooltip exists for all schema names, not for truncated ones only.

bubafinder:
Don't like it. 
You can look at a sample on tags, we are only showing a tooltip for truncated text.

image

client/components/catalog/Card/index.vue
```diff
@@ -133,22 +146,25 @@ export default {
...
-    padding: 0 0 0 0.25rem;
+    flex-grow: 1;
     color: #fafafa;
     font-size: 0.75rem;
     font-weight: 500;
     letter-spacing: 1px;
     text-transform: uppercase;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
```
abasic: Vuetify has a class for this called [text-truncate](https://vuetifyjs.com/en/styles/text-and-typography/#wrapping-and-overflow).

client/components/catalog/Card/index.vue
```diff
@@ -133,22 +146,25 @@ export default {
 .card-body {
   padding: 0.625rem 0 0;
 
-  .v-card__title {
-    line-height: 1.75rem;
+  .header {
+    display: flex;
+    align-items: center;
```
abasic: There are also [classes for flex](https://vuetifyjs.com/en/styles/flex/).

client/components/catalog/Card/index.vue
```diff
@@ -133,22 +146,25 @@ export default {
...
   }
 
   .schema-name {
-    padding: 0 0 0 0.25rem;
+    flex-grow: 1;
```
abasic: Also a class for [flex-grow](https://vuetifyjs.com/en/styles/flex/#flex-grow-and-shrink).

abasic (changes requested):

abasic (approved):

client/components/catalog/Card/index.vue
```diff
@@ -105,19 +116,22 @@ export default {
         name,
         params: { repositoryId: this.repository.id }
       });
+    },
+    detectTruncation() {
+      const { clientWidth, scrollWidth } = this.$refs.schemaName;
```
underscope, line 121: @mcapeta 
Would it make sense to hook into https://vuetifyjs.com/en/features/breakpoints/#breakpoint-service 🤡

underscope:
@mcapeta When you get a chance, extract it into a separate component so it doesn't pollute Card as much 😉 
ty ❤️

PR #730, opened 2021-02-08
Move tree view to structure view

MiroDojkic:
### This PR:
- moves tree view under structure view
- button is added to outline toolbar to toggle between structure view and tree view
- fixes delete confirmation dialog (swapped ID for short ID) when activity is deleted while graph view is on


MiroDojkic:
@underscope I've fixed cut/padding issue, but I left grey button as was before the discussion.
Also, I didn't change the behavior when Outline nav link is clicked, so if the graph view was shown before leaving view, it'll be displayed again when the user returns.

Let me know if you'd like me to change any of that. 🙂

bubafinder:
Don't see any problems in terms of functionality.
Maybe to consider using the `v-switch` component for view toggling instead of using a button?
https://vuetifyjs.com/en/components/switches/#switches

It would be a benefit in the case when a user from tree view goes into the activity page, and after returning back to the structure page, the view switch would still be active on the tree view, as the last selected user state.

PR #745, opened 2021-02-12
Improve Resource helper

### LINKED ISSUE #740: Issue with editing repository settings
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request

## Context
If super admin is not added to the repository he is not able to edit repository meta (i.e. name).

kronicker:
This PR aims to resolve issues when the store collection `update` helper is invoked w/o fetching items first by:
- Adding `id`-`uid` mapping after `getById`
- Favoring `id` from the model (if present) in `update`

@underscope Even though either of these two changes fixes the issue I believe it would be good to do both to be on the safe side for any future use-cases. Thoughts?

Closes #740

### DIFF
--- a/client/store/helpers/resource.js
+++ b/client/store/helpers/resource.js
@@ -88,7 +88,7 @@ export default class Resource {
    * @param {object} changes - Key-value collection of properties to update.
    */
   update(uid, changes) {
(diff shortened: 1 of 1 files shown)


underscope:
ty 🏆

PR #747, opened 2021-02-15
Fix activity repository_id foreign key error on repository import

kronicker:
This PR:
- ~changes `Activity#bulkCreate` to use transaction from options by default~
- fixes the order of arguments for `sequelize#transaction`

This fixes the issue with import breaking with the `repository_id` foreign key constraint error.
`Activity#bulkCreate` was running in different isolation scope from the import job because it was spawning new transaction.

### DIFF
--- a/server/activity/activity.model.js
+++ b/server/activity/activity.model.js
@@ -75,28 +75,28 @@ class Activity extends Model {
   }
 
   static async create(data, opts) {
-    return this.sequelize.transaction(async t => {
-      const transaction = opts.transaction || t;
-      const activity = await super.create(data, { ...opts, transaction });
-      if (activity.isTrackedInWorkflow) {
-        const defaultStatus = getDefaultActivityStatus(activity.type);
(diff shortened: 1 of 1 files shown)


server/activity/activity.model.js
```diff
@@ -75,26 +75,28 @@ class Activity extends Model {
...
   static async create(data, opts) {
-    return this.sequelize.transaction(async transaction => {
-      const activity = await super.create(data, { ...opts, transaction });
-      if (activity.isTrackedInWorkflow) {
-        const defaultStatus = getDefaultActivityStatus(activity.type);
-        await activity.createStatus(defaultStatus, { transaction });
-      }
-      return activity;
-    }, { transaction: opts.transaction });
+    return this.sequelize.transaction({ transaction: opts.transaction },
```
kronicker, line 78: This is possible because of the [undocumented `transaction` option param](https://github.com/sequelize/sequelize/blob/v4/lib/transaction.js#L38).

MiroDojkic (approved):

PR #749, opened 2021-02-16
Pass context to the element hooks

ikovac:
**This PR **
Fixes diff preview for tce-sprout-video by passing context to the element hooks.

### DIFF
--- a/server/content-element/hooks.js
+++ b/server/content-element/hooks.js
@@ -55,13 +55,13 @@ function add(ContentElement, Hooks, Models) {
     return Comment.update({ activityId: null }, options);
   }
 
-  function customElementHook(hookType, element) {
+  function customElementHook(hookType, element, options) {
     const elementHookTypes = elementHookMappings[hookType];
     if (!elementHookTypes) return;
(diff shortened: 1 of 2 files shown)


server/content-element/hooks.js
```diff
@@ -55,13 +55,13 @@ function add(ContentElement, Hooks, Models) {
...
 
-  function customElementHook(hookType, element) {
+  function customElementHook(hookType, element, options) {
     const elementHookTypes = elementHookMappings[hookType];
     if (!elementHookTypes) return;
     return Promise.resolve(elementHookTypes)
       .map(hook => elementRegistry.getHook(element.type, hook))
       .filter(Boolean)
-      .reduce((result, hook) => hook(result), element);
+      .reduce((result, hook) => hook(result, options.context), element);
```
MiroDojkic: I'd suggest that we pass the whole options object since we're already using it. In case we need to pass something else, we wouldn't need to change the signature of the hooks and therefore update all elements.

MiroDojkic (changes requested):

MiroDojkic (approved):

PR #752, opened 2021-02-17
Fix: Toolbar is not updated on element update

ikovac:
When an element is updated, the toolbar still has an outdated version of the selected element.
This PR fixes mentioned issue.

### DIFF
--- a/client/components/editor/index.vue
+++ b/client/components/editor/index.vue
@@ -27,6 +27,7 @@
 <script>
 import { mapActions, mapGetters, mapMutations, mapState } from 'vuex';
 import ActivityContent from './ActivityContent';
+import find from 'lodash/find';
 import get from 'lodash/get';
 import { getElementId } from 'tce-core/utils';
 import Sidebar from './Sidebar';
@@ -40,16 +41,18 @@ export default {
     repositoryId: { type: Number, required: true },
     activityId: { type: Number, required: true }
   },
-  data: () => ({
-    isLoading: true,
-    selectedElement: null
-  }),
+  data: () => ({ isLoading: true }),
   computed: {
     ...mapGetters('editor', ['activity', 'contentContainers', 'rootContainerGroups']),
     ...mapGetters('repository', ['repository', 'outlineActivities']),
     ...mapGetters('repository/userTracking', ['getActiveUsers']),
+    ...mapGetters('repository/contentElements', ['elements']),
     ...mapState('editor', ['showPublishDiff']),
-    activeUsers: vm => vm.getActiveUsers('activity', vm.activityId)
+    activeUsers: vm => vm.getActiveUsers('activity', vm.activityId),
+    selectedElement() {
+      const { elementId } = this.$route.query;
+      return find(this.elements, { uid: elementId });
+    }
   },
   methods: {
     ...mapMutations('editor', ['togglePublishDiff']),
@@ -58,7 +61,6 @@ export default {
       this.togglePublishDiff(false);
     },
     selectElement(element) {
-      this.selectedElement = element;
       const selectedElementId = getElementId(element);
       const { elementId: queryElementId, ...query } = this.$route.query;
       if (selectedElementId === queryElementId) return;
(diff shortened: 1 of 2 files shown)


client/components/editor/index.vue
```diff
@@ -40,16 +41,18 @@ export default {
...
     ...mapGetters('repository', ['repository', 'outlineActivities']),
     ...mapGetters('repository/userTracking', ['getActiveUsers']),
+    ...mapGetters('repository/contentElements', ['elements']),
     ...mapState('editor', ['showPublishDiff']),
-    activeUsers: vm => vm.getActiveUsers('activity', vm.activityId)
+    activeUsers: vm => vm.getActiveUsers('activity', vm.activityId),
+    selectedElement() {
+      const { elementId } = this.$route.query;
+      return find(this.elements, { uid: elementId });
+    }
```
MiroDojkic: Does it make sense to move this to the `repository/contentElements` getter so it can be easily reused?

client/store/modules/repository/content-elements/getters.js
```diff
@@ -1 +1,5 @@
+import find from 'lodash/find';
+
 export const elements = state => state.items;
+
+export const getElementByUid = state => uid => find(state.items, { uid });
```
underscope, line 5: Since items / elements is object keyed by id, no need to use find 😉

client/components/editor/index.vue
```diff
@@ -40,16 +40,18 @@ export default {
...
   computed: {
     ...mapGetters('editor', ['activity', 'contentContainers', 'rootContainerGroups']),
     ...mapGetters('repository', ['repository', 'outlineActivities']),
     ...mapGetters('repository/userTracking', ['getActiveUsers']),
+    ...mapGetters('repository/contentElements', ['getElementByUid']),
     ...mapState('editor', ['showPublishDiff']),
-    activeUsers: vm => vm.getActiveUsers('activity', vm.activityId)
+    activeUsers: vm => vm.getActiveUsers('activity', vm.activityId),
+    selectedElement() {
+      const { elementId } = this.$route.query;
```
underscope, line 52: Consider extracting it as selectedElementId computed (If you are not going to move it in the store; see prev comment) :bowtie:

client/components/editor/index.vue
```diff
@@ -58,7 +60,6 @@ export default {
       this.togglePublishDiff(false);
     },
     selectElement(element) {
-      this.selectedElement = element;
```
underscope, line 61: If all elements are not going to create clone of element within Edit component, SSE might override their local changes for edited element. Do we need some kind of Edit cache; with last change wins strategy to avoid weird behaviour? Please check all potential side effects and decide about desired behaviour. 

Side note: I noticed that saving element meta (element configurable inputs; reach out to @abasic  / @kronicker  if you need help configuring it) are not properly being updated upon SSE change triggered by another user until user refocuses. That relates to previous topic / all sections should have similar behaviour.

client/components/editor/index.vue
```diff
@@ -40,16 +40,18 @@ export default {
...
+  data: () => ({ isLoading: true }),
   computed: {
     ...mapGetters('editor', ['activity', 'contentContainers', 'rootContainerGroups']),
     ...mapGetters('repository', ['repository', 'outlineActivities']),
     ...mapGetters('repository/userTracking', ['getActiveUsers']),
+    ...mapGetters('repository/contentElements', ['getElementByUid']),
     ...mapState('editor', ['showPublishDiff']),
-    activeUsers: vm => vm.getActiveUsers('activity', vm.activityId)
+    activeUsers: vm => vm.getActiveUsers('activity', vm.activityId),
+    selectedElement() {
```
underscope, line 51: This might also be a getter within `editor` module, same as activity.

client/components/editor/index.vue
MiroDojkic: @underscope it seems to me that the behavior might be weird for one of the users in either case, as long as we're not showing that the content is being edited by the other user.
Do you think this solution would behave ok/predictably if we'd show that? Avoiding caching logic without a UX trade-off looks like a nice win. 😄

ikovac:
I am closing this PR because it introduces some new issues with the tce-carousel element.
The tce-carousel element uses embed elements which are stored as stringify json inside tce-carousel's data field instead as rows inside the `content-element` table. 
This approach is causing that `getElementByUid(elementId)` getter returns undefined as it can't find embed elements inside the element list. The final result is that toolbar is not shown when an element inside the-carousel is clicked.

PR #757, opened 2021-03-05
Feature: Add assets permissions

ikovac:
This PR extends an existing storage proxy. It introduces a new storage folder structure where assets are stored under `repository/repositoryId`. This structure is suitable for checking asset permissions.

**Todo:**
- [x] Change storage folder structure
- [x] Mount assets to the repository router
- [x] Pass `repositoryId` to the storage proxy
- [x] Clear storage cookies on logout
- [x] Update tce-scorm to match new changes
- [x] Add migration script

Note: In order to work with the tce-scorm element this PR need to be merged: https://github.com/ExtensionEngine/tce-scorm/pull/2

**Important** 💥 
This PR introduces breaking changes related to assets. It is necessary to run a migration script to make it work properly.
Run `npm run migrateAssetsLocation`


client/components/common/mixins/upload.js
```diff
@@ -1,11 +1,15 @@
 import downloadMixin from 'utils/downloadMixin';
 import loader from '@/components/common/loader';
+import { mapGetters } from 'vuex';
 import { mapRequests } from '@/plugins/radio';
 
 export default {
   inject: ['$storageService'],
   mixins: [downloadMixin],
   data: () => ({ uploading: false }),
+  computed: {
```
lukasaric: You don't need to spread repository getter. Just inline it 😉

package.json
```diff
@@ -31,6 +31,7 @@
     "integration:token": "node -r ./server/script/preflight ./server/script/generateIntegrationToken.js",
     "detachDeletedRepos": "node -r ./server/script/preflight ./server/script/detachDeletedRepos.js",
     "migrateAssessmentPools": "node -r ./server/script/preflight ./server/script/migrateAssessmentPools.js",
+    "migrateAssetsLocation": "node -r ./server/script/preflight ./server/script/migrateAssetsLocation.js",
```
lukasaric, line 34: Since you are adding another one script, `preflight` command logic is same in all scripts above. It would be nice thing to separate it as `preflight` script and then reuse it where it's needed. :bowtie:

server/script/migrateAssetsLocation.js
```diff
@@ -0,0 +1,194 @@
...
+const fromPairs = require('lodash/fromPairs');
+const get = require('lodash/get');
+const { Op } = require('sequelize');
+const Promise = require('bluebird');
+const { protocol } = require('../../config/server/storage');
+const storage = require('../shared/storage');
+const toPairs = require('lodash/toPairs');
+
+const regex = /repository\/assets\/(.*)/;
+const ceTypes = ['IMAGE', 'VIDEO', 'AUDIO', 'PDF', 'CAROUSEL', 'MODAL'];
```
lukasaric: It should be uppercased. And name should be more precise...e.g `ELEMENT_TYPES`.

server/script/migrateAssetsLocation.js
```diff
@@ -0,0 +1,194 @@
...
+const revisionsTypes = ['CONTENT_ELEMENT', 'ACTIVITY'];
```
lukasaric: Same as above ☝️

server/script/migrateAssetsLocation.js
```diff
@@ -0,0 +1,194 @@
...
+    const payload = await migrateActivity(it);
+    return it.update(payload, { transaction });
+  });
+}
+
+async function migrateActivity(activity) {
+  const { repositoryId, data: meta } = activity;
+  const fileMetas = getActivityMetadata(activity)
+    .filter(it => it.type === 'FILE')
+    .map(it => it.key);
```
lukasaric: This could be solved as one `reduce`. But for me this is still fine, since the number of iterations is negligible. 
This also applies on other places in this file where you use similar chain.

server/shared/storage/index.js
```diff
@@ -54,6 +54,10 @@ class Storage {
     return this.provider.copyFile(key, newKey, options);
   }
 
+  getAssetsPath(repositoryId) {
+    return `repository/${repositoryId}/${config.path}`;
```
lukasaric: Use `path.join` instead. It's already imported and it's semantically better. Also be sure that all arguments are strings 😉

server/shared/storage/proxy/index.js
```diff
@@ -27,21 +32,34 @@ class Proxy {
     return this.isSelfHosted && this.provider.path;
   }
 
-  getSignedCookies(resource, maxAge) {
-    return this.provider.getSignedCookies(resource, maxAge);
+  getSignedCookies(repositoryId, maxAge) {
+    const resource = `repository/${repositoryId}`;
```
lukasaric: Same as above (`path.join`) ☝️

server/shared/storage/storage.controller.js
```diff
@@ -17,36 +17,37 @@ function getUrl(req, res) {
...
 }
 
-async function uploadArchiveContent(archive, name) {
+async function uploadArchiveContent(archive, name, repository) {
   const buffer = await readFile(archive);
   const content = await JSZip.loadAsync(buffer);
   const files = pickBy(content.files, it => !it.dir);
   const keys = await Promise.all(Object.keys(files).map(async src => {
-    const key = path.join(config.path, `${name}/${src}`);
+    const key = path.join(getAssetsPath(repository.id), `${name}/${src}`);
```
lukasaric:
```javascript
path.join(getAssetsPath(repository.id), name, src);
```

server/script/migrateAssetsLocation.js
```diff
@@ -0,0 +1,194 @@
...
+  if (!url) return data;
+  const { key, newKey } = getKeysFromUrl(url, repositoryId) || {};
+  if (!key || !newKey) return data;
+  await storage.copyFile(key, newKey);
+  return { ...element.data, url: newKey };
+}
+
+async function embedsMigrationHandler(element) {
+  const { repositoryId, data } = element;
+  const embeds = await Promise.map(
```
lukasaric: Hm..maybe extract this map logic as `getEmbeds` separate function. It's lot going one, so it will be more readable

lukasaric (changes requested):
Nice work! :rocket:
@MiroDojkic Still needs a more detailed sweep on `migrateAssetsLocation.js`.

server/script/migrateAssetsLocation.js
ikovac:
```js
async ([id, embed]) => {
  if (embed.type !== 'VIDEO') return [id, embed];
  const data = await defaultMigrationHandler({ repositoryId, ...embed });
  return [id, { ...embed, data }];
}
```
The map iteration function is not that heavy (3 lines of code). What do you think of extracting full get embeds logic(map + reduce) instead?

```js
...
  const embeds = getEmbeds(...);
  return { ...data, embeds };
}

function getEmbeds(...) {
  return Promise.map(
    Object.entries(embeds),
    async ([id, embed]) => {
      if (embed.type !== 'VIDEO') return [id, embed];
      const data = await defaultMigrationHandler({ repositoryId, ...embed });
      return [id, { ...embed, data }];
    })
    .reduce((acc, [id, embed]) => ({ ...acc, [id]: embed }), {});
}

File: server/script/migrateAssetsLocation.js (hunk 5 above)
Comment (ikovac) [reply server/script/migrateAssetsLocation.js]: Yeah, I agree, but I think filter + map combination is more readable than using a reducer so it better fits in my opinion as long as the number of iterations is insignificant.

File: server/script/migrateAssetsLocation.js (hunk 9 above)
Comment (lukasaric) [reply server/script/migrateAssetsLocation.js]: That was actually what I meant you to do 😉

File: package.json (hunk 2 above)
Comment (kronicker) [reply package.json:34]: That's not possible since `-r ./server/script/preflight` is a runtime executable argument - [check the docs](https://nodejs.org/api/cli.html#cli_r_require_module). 
If you split it to `node -r ./server/script/preflight` and `node [script-name.js]` you would get one node REPL with `preflight` prerequired and script run without `preflight`.

**Comment by ikovac** [comment]
After conversation with @underscope, we decided to change migration logic a little bit so that we don't have to specify the exact CONTENT_ELEMENT which to migrate and rather migrate all content elements which contain assets/embeds data property.

File: config/server/storage.js [hunk 10]
```diff
@@ -2,7 +2,7 @@
 
 module.exports = {
   provider: process.env.STORAGE_PROVIDER,
-  path: 'repository/assets',
+  path: 'assets', // repository/${repositoryId}/${path}
```
MiroDojkic: Should we add more context / describe this behavior in a sentence?

server/shared/publishing/helpers.js
```diff
@@ -198,9 +198,11 @@ async function fetchCustomContainers(parent, config) {
 function unpublishDeletedContainers(parent, containers) {
   const baseUrl = getBaseUrl(parent.repositoryId, parent.id);
   const filePaths = getContainersFilePaths(baseUrl, containers);
+  const assetsPath = storage.getStoragePath(parent.repositoryId);
   return storage
     .listFiles(baseUrl)
     .then(publishedFilePaths => {
+      publishedFilePaths = publishedFilePaths.filter(it => !it.startsWith(assetsPath));
```
MiroDojkic: Do you think moving this to a function (i.e. `removeAssetFilePaths`) would make this function more straightforward to read?

server/shared/storage/index.js
```diff
@@ -54,6 +54,10 @@ class Storage {
     return this.provider.copyFile(key, newKey, options);
   }
 
+  getStoragePath(repositoryId) {
```
MiroDojkic: Consider removing the `Storage` from the name, since it's contained in the context: `storage.getPath()`

server/shared/storage/proxy/index.js
```diff
@@ -27,21 +32,34 @@ class Proxy {
...
+    const { REPOSITORY } = storageCookies;
+    const isRepositoryId = cookies[REPOSITORY] === repositoryId.toString();
+    return isRepositoryId && this.provider.hasCookies(cookies);
   }
 
   getFileUrl(key) {
     return this.provider.getFileUrl(key);
   }
+
+  getStorageCookieNames() {
```
MiroDojkic: Should we rename this method to `getCookieNames` to make it consistent with the rest (`hasCookies`, `verifyCookies`)?

server/shared/storage/proxy/index.js
```diff
@@ -27,21 +32,34 @@ class Proxy {
     return this.isSelfHosted && this.provider.path;
   }
 
-  getSignedCookies(resource, maxAge) {
-    return this.provider.getSignedCookies(resource, maxAge);
+  getSignedCookies(repositoryId, maxAge) {
```
MiroDojkic: Should we consider making a new module within the `repository` module which would extend `Storage` and `Proxy` classes with repository-based access?
The motivation for such approach would be to keep the generic storage implementation which we can then move out of the Tailor to be reused by other Node projects.

server/shared/storage/proxy/providers/cloudfront.js
```diff
@@ -48,6 +50,10 @@ class CloudFront {
   getFileUrl(key) {
     return urlJoin(this.host, key);
   }
+
+  getStorageCookieNames() {
+    return values(storageCookies);
```
MiroDojkic: What's the reason to use `lodash/value` instead of `Object.values`?

server/shared/storage/storage.controller.js
```diff
@@ -17,36 +17,37 @@ function getUrl(req, res) {
...
 
 module.exports = { getUrl, upload };
 
-async function uploadFile(file, name) {
+async function uploadFile(file, name, repository) {
   const buffer = await readFile(file);
   const hash = sha256(file.originalname, buffer);
   const extension = path.extname(file.originalname);
-  const key = path.join(config.path, `${hash}___${name}${extension}`);
+  const sufix = `${hash}___${name}${extension}`;
```
MiroDojkic: `sufix` -> `suffix` ✏️ 
Would `fileName` or something like that describe this value in a more helpful way?

MiroDojkic (changes requested):

server/shared/storage/storage.controller.js
```diff
@@ -17,36 +17,37 @@ function getUrl(req, res) {
...
   }
-  const asset = await uploadFile(file, name);
+  const asset = await uploadFile(file, name, repository);
   return res.json(asset);
 }
 
 module.exports = { getUrl, upload };
 
-async function uploadFile(file, name) {
+async function uploadFile(file, name, repository) {
```
underscope: Consider having repository as the first argument (uploading file for repository) and passing only repository id (minimal info required) :bowtie: 😉

server/shared/storage/storage.controller.js
```diff
@@ -17,36 +17,37 @@ function getUrl(req, res) {
...
-  const key = path.join(config.path, `${hash}___${name}${extension}`);
+  const fileName = `${hash}___${name}${extension}`;
+  const key = path.join(getPath(repository.id), fileName);
   await saveFile(key, buffer, { ContentType: file.mimetype });
   const publicUrl = await getFileUrl(key);
   return { key, publicUrl, url: getStorageUrl(key) };
 }
 
-async function uploadArchiveContent(archive, name) {
+async function uploadArchiveContent(archive, name, repository) {
```
underscope: See prev comment

server/script/migrateAssetsLocation.js
```diff
@@ -0,0 +1,252 @@
...
+  const { type, data } = element;
+  if (type === 'IMAGE') return imageMigrationHandler(element);
+  if (data.embeds) return embedsMigrationHandler(element);
+  if (data.assets) return defaultMigrationHandler(element);
+  return data;
+}
+
+async function migrateContentElementMeta(element) {
+  const { repositoryId } = element;
+  const metaInputs = schemasIds
```
underscope: Resolve it once, upon running the script.

server/script/migrateAssetsLocation.js
```diff
@@ -0,0 +1,252 @@
...
+    .map(id => getElementMetadata(id, element))
+    .filter(meta => !meta.isEmpty)
+    .map(meta => meta.inputs);
+  const fileMetas = flatten(metaInputs)
+    .filter(it => it.type === 'FILE')
+    .map(it => it.key);
+  return getNewMeta(fileMetas, element.meta, repositoryId);
+}
+
+async function getNewMeta(fileMetas, meta, repositoryId) {
```
underscope: Consider tweaking naming / adding docs and updating parameter order to match the content hierarchy, e.g.:

```js
 async function migrateFileMeta(repositoryId, metaInputs, metaConfigs) {
    ...
 }
```

server/script/migrateAssetsLocation.js
```diff
@@ -0,0 +1,252 @@
...
+  const newMeta = await Promise.map(toPairs(meta), async it => {
+    const [id, value] = it;
+    if (!fileMetas.includes(id)) return it;
```
underscope: Why iterating per data set, wouldn't it be more intuitive for each defined (config) file meta to check if it exists/is set ? 🤔

server/script/migrateAssetsLocation.js
```diff
@@ -0,0 +1,252 @@
...
+  };
+}
+
+async function migrateRevision(revision) {
+  const { entity, state } = revision;
+  const payload = await (mapEntityToAction[entity] && mapEntityToAction[entity](state));
+  return { state: { ...state, ...payload } };
+}
+
+function getKeysFromUrl(url, repositoryId) {
```
underscope: accent that this is repository ('map') where file is being copied. Consider updating the function name to something like `resolveNewURL(assetURL, targetDir)` (or NewLocation/NewKey)

server/script/migrateAssetsLocation.js
```diff
@@ -0,0 +1,252 @@
...
+    async ([id, embed]) => {
+      const payload = await migrateContentElement({ repositoryId, ...embed });
+      return [id, { ...embed, ...payload }];
+    }
+  ).reduce((acc, [id, embed]) => ({ ...acc, [id]: embed }), {});
+}
+
+async function defaultMigrationHandler(element) {
+  const { repositoryId, data } = element;
+  const url = get(element, 'data.assets.url');
```
underscope: Why picking up .url 🤔 Any key within assets dir that has storage prefix should be resolved ⚠️

underscope (changes requested):

server/shared/publishing/helpers.js
```diff
@@ -195,15 +194,12 @@ async function fetchCustomContainers(parent, config) {
...
   const baseUrl = getBaseUrl(parent.repositoryId, parent.id);
+  const prevFilePaths = getContainersFilePaths(baseUrl, prevContainers);
   const filePaths = getContainersFilePaths(baseUrl, containers);
-  return storage
-    .listFiles(baseUrl)
-    .then(publishedFilePaths => {
-      const redundantFilePaths = difference(publishedFilePaths, filePaths);
-      if (redundantFilePaths.length) return storage.deleteFiles(redundantFilePaths);
-    });
+  const deletedContainersFiles = differenceWith(prevFilePaths, filePaths);
```
underscope: => deletedContainerFiles 🤔 ?

PR #767, opened 2021-03-08
Fix publish diff for deleted container

### LINKED ISSUE #765: Compare with published doesn't work as expected if content container is deleted
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request

## Context
If you publish an element, then add a couple more content elements inside the content container, once you delete the whole content container, the compare with published feature will still show new (not published) elements inside the non-existing content container.

CompareIssue

MiroDojkic:
### This PR:
Resolves #765 - fixes publish diff shown for deleted container with elements.

To make it consistent with the non-deleted container elements, new behavior should be as follows:
- elements that were added after the publish are not shown (those are deleted at the moment and were never published so there's no diff)
- rest of the published elements (unchanged, changed or removed) are shown as removed


underscope:
@kjuej @markolovric Please revisit on test server

PR #789, opened 2021-03-17
Replace active users map with in-memory key-value store

ikovac:
This PR:
Active users map is replaced using key-value in-memory storage.
Supported providers:
- [lru-cache](https://github.com/isaacs/node-lru-cache#readme)
- [redis](https://github.com/luin/ioredis)

QA note:
- this PR affects the concurrent element editing and the repository tracker of active users.
- in order to test it using Redis store, [Docker](https://docs.docker.com/docker-for-mac/install/) needs to be installed on your system. 
- start docker 
- run: `docker run --name redis -p 6379:6379 redis`
- edit redis store option in .env if needed
- feel free to ping me if you have any question :)


.env.example
```diff
@@ -113,3 +113,18 @@ CONSUMER_CLIENT_ID=
...
+# Default value is 0 (no expiration time)
+STORE_TTL=0
+
+# redis options
+# redis port - default is 6379
+REDIS_PORT=
+# redis host - default is 'localhost'
+REDIS_HOST=
+# redis password - empty by default
+REDIS_PASSWORD=
```
MiroDojkic: Besides default values, these comments are one-to-one match with the variable names. Should we set defaults as values and get rid of the comments?

server/repository/feed/store.js
```diff
@@ -1,28 +1,40 @@
...
+  return store.set(key, { ...record, contexts });
+}
 
-  _findOrCreate(user) {
-    if (!this.has(user.id)) {
-      const connectedAt = new Date();
-      this.set(user.id, { ...user, connectedAt, contexts: [] });
-    }
-    return this.get(user.id);
+async function removeContext(user, filterFn) {
```
MiroDojkic: `filterFn` -> `predicate`?

server/repository/feed/store.js
```diff
@@ -1,28 +1,40 @@
...
+  const key = getKey(user.id);
+  const record = await store.get(key);
+  if (!record) return;
+  const contexts = record.contexts.filter(it => !filterFn(it));
+  if (contexts.length <= 0) return store.del(key);
```
MiroDojkic: Don't think that the `length` property of an array can ever have a negative value. 🤔 
Therefore we can replace this with:
```suggestion
  if (contexts.length) return store.del(key);
```

config/server/store.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+module.exports = {
+  provider: process.env.STORE_PROVIDER || 'local',
+  local: {
+    maxAge: parseInt(process.env.STORE_TTL, 10) || 0
```
MiroDojkic: Since TTL is common config, should we move it below `provider` and assign it to the `maxAge` in the `local` strategy?

server/repository/feed/store.js
```diff
@@ -1,28 +1,40 @@
...
+  return store.set(key, { ...record, contexts });
+}
+
+async function getActiveUsers() {
+  const activeUserKeys = await store.keys('active-user-*');
```
MiroDojkic: We could extract key prefix to a constant, reuse `getKey('*')` or have separate `Store.getAllKeys(prefix: string): Key` method to keep this logic in one place.

server/repository/feed/store.js
```diff
@@ -1,28 +1,40 @@
...
+  return Promise
+    .map(activeUserKeys, key => store.get(key))
+    .reduce((acc, user) => ({ ...acc, [user.id]: user }), {});
+}
+
+async function findOrCreate(user) {
+  const key = getKey(user.id);
+  if (!await store.has(key)) {
```
MiroDojkic: Please consider if awaiting outside of the `if` condition would make this simpler to read.

server/shared/store/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+  get(key) {
+    return this.provider.get(key);
+  }
+
+  has(key) {
+    return this.provider.has(key);
+  }
+
+  keys(pattern = '*') {
```
MiroDojkic: Consider renaming this method to `getKeys` to indicate that it's a function, not a class field.

server/shared/store/index.js
```diff
@@ -0,0 +1,55 @@
...
+    return this.provider.keys(pattern);
+  }
+
+  del(key) {
```
MiroDojkic: `del` -> `delete`?

server/shared/store/providers/local.js
```diff
@@ -0,0 +1,47 @@
...
+
+class Local {
+  constructor(config) {
+    config = schema.validateSync(config, { stripUnknown: true });
+    this.name = 'local';
+    this.maxAge = config.maxAge;
+    this.client = new LRU({ maxAge: config.maxAge });
+  }
+
+  set(key, value, ttl) {
```
MiroDojkic: Would this work
```suggestion
  set(key, value, ttl = this.maxAge) {
```
?

server/shared/store/providers/local.js
```diff
@@ -0,0 +1,47 @@
...
+const yup = require('yup');
+
+const schema = yup.object().shape({
+  maxAge: yup.number()
+});
+
+class Local {
+  constructor(config) {
+    config = schema.validateSync(config, { stripUnknown: true });
+    this.name = 'local';
```
MiroDojkic: What's the purpose of this field, I can't find it being used anywhere?

MiroDojkic (changes requested):

server/repository/feed/store.js
ikovac: Yeah, I agree, that was a little bit strange to me too, but I keep it because I didn't want potentially break something so I just replaced Map store with the key-value store and left all code as it is.

server/shared/store/providers/local.js
ikovac: The purpose is that consumer can get information about the key-value store provider which is being used. In addition, a consumer can make some actions depend on the store type. 
It is not used anywhere in this PR but I find it potentially useful for future store consummation. What do you think?

MiroDojkic: @ikovac it's easy to add this if necessary, so I suggest we remove it to avoid piling up dead code.
Also, we could do this check with the `instanceof` operator and avoid having to change multiple things to preserve such behavior in the future. 😄

server/repository/feed/store.js
```diff
@@ -1,28 +1,41 @@
...
 
-  toJSON() {
-    return Array.from(this.entries())
-      .reduce((acc, [key, value]) => ({ ...acc, [key]: value }), {});
-  }
+async function addContext(user, context) {
+  const key = getKey(user.id);
+  const record = await findOrCreate(user);
+  const contexts = [...record.contexts, context];
+  return store.set(key, { ...record, contexts }, 0);
```
MiroDojkic: Would aliasing `0` with something like `NO_EXPIRATION_TTL` / `INFINITE_TTL` make it easy to infer what's going on?

MiroDojkic (approved):

kjuej:
@ikovac @MiroDojkic Seems to be working as expected!

PR #791, opened 2021-03-18
Handle missing schema meta validators

lukasaric:
This PR imports all `vee-validate` rules to avoid using meta input validators which are not currently exposed.


config/shared/activities.js
```diff
@@ -179,3 +180,12 @@ function getRepositoryRelationships(schemaId) {
   return flatMap(structure, it => it.relationships)
     .reduce((acc, { type }) => union(acc, [type]), []);
 }
+
+function getMetaValidators() {
+  const validators = reduce(SCHEMAS, (acc, { structure }) => {
+    const metas = flatMap(structure, 'meta');
```
underscope: Meta can be attached to activity, repository and content element ⚠️

client/utils/validation.js
```diff
@@ -66,7 +69,23 @@ const rules = {
   url
 };
 
-forEach(rules, (rule, name) => extend(snakeCase(name), {
-  message: messages[name],
-  ...rule
-}));
+processRules(rules).then(registerRules);
```
underscope: This will still include everything upon build so I am not sure if it makes sense to do it.
@MiroDojkic @abasic?

MiroDojkic: @underscope agreed. The module is [light](https://bundlephobia.com/result?p=@vee-validate/rules@4.1.20) so I think we should be fine with the static import of all the rules.

lukasaric: Since module is light, I see no further point in extracting meta validations and further comparisons. I think it would fine to expose/register all of them instead registering only the ones we want to use.

underscope:
@kjuej Please take a look if validations work on input forms (creating a repo, adding user, creating activities, etc...)

PR #793, opened 2021-03-23
Trigger webhook on repository details publish

lukasaric:
This PR enables `webhook` on `repository` info publish.
In order to properly test this you need to:
- Set in motion [Client Credentials Flow](https://oauth.tools/collection/1599045253169-GHF)
- According to above said, set `.env`:
e.g
```
CONSUMER_CLIENT_ID=demo-backend-client
CONSUMER_CLIENT_SECRET= ---> given secret from link above <---
CONSUMER_CLIENT_TOKEN_HOST=https://login-demo.curity.io
CONSUMER_CLIENT_TOKEN_PATH=/oauth/v2/oauth-token
```
- Visit [Webhook](https://webhook.site/)
- Copy given `unique URL` to `CONSUMER_WEBHOOK_URL` in `.env`


server/shared/publishing/publishing.service.js
```diff
@@ -15,17 +15,18 @@ class PublishingService {
     this.queue = new PromiseQueue(1, Infinity);
   }
 
+  async publish(action, payload) {
```
MiroDojkic: Do you think we could find a name that better describes this method since it doesn't really do publishing specifically (it just ensures that the webhook is called after action)?

Or it could return a function (i.e. `toPublishJob`) so we can get rid of the arrows in the `queue.add` call. :bowtie:

MiroDojkic: Also, should we make it private?

server/shared/publishing/publishing.service.js
```diff
@@ -42,3 +37,11 @@ class PublishingService {
 }
 
 module.exports = new PublishingService();
+
+function toPublishJob(action, payload) {
```
underscope: consider using create rather than to ;)

PR #797, opened 2021-03-26
Enable proxy access management extension

MiroDojkic:
### This PR:
is an attempt to achieve the following goals:
- decouple storage and proxy modules from the rest of the Tailor services, so we can package them and make them reusable in any Node app
- enable developers to extend proxy's access management implementation; for example, to add new allowed resources, we need to add the resource along with the cookie to identify it (so we don't need to decrypt the signature cookie each time we want to verify if the resource is allowed)
- enable developers to extend storage service; for example, to specialize the service instance to save files in a specific folder
- support proxying multiple storages

#### 🔖 Todos:
- [x] decouple access management from the proxy
- [x] add proxy support for multiple storages
- [x] enable extension of the proxy access manager
- [x] decouple storage from storage middleware to support mounting on different paths
- [x] parametrize upload mixin to enable passing `repositoryId`option
- [x] enable removing all proxy cookies
- [ ] enable adding allowed `resources` without deleting previous `resources` in the storage cookies
- [ ] create repository-specific `UploadProvider` component that always saves asset in the appropriate repository folder
- [ ] update assets migration script


rkusan00:
Closing, storage-proxy layer this PR extends was removed from the codebase in the https://github.com/tailor-cms/author

PR #798, opened 2021-03-27
Improve assets migration script

ikovac:


server/script/migrateAssetsLocation.js
```diff
@@ -24,224 +22,255 @@ const REVISION_TYPES = ['REPOSITORY', 'ACTIVITY', 'CONTENT_ELEMENT'];
...
-    },
-    {
-      title: 'Migrate revisions',
-      task: () => migrateRepositoryRevisions(repository.id, schemaFileMetas, transaction)
-    }
-  ]);
+  async run() {
+    const tasks = await this.getTasks();
+    return tasks.run().then(() => this.transaction.commit());
+  }
```
ikovac, line 53: I'm thinking to move this Migration class logic outside the class and put it inside `migrate` function and just keep `RepositoryMigration` class. Doing it that way I will avoid this: 
```js
migration.initialize().then(() => {
  migration.run()
    .then(() => {
      console.info('Migration script was executed successfully.');
      process.exit(0);
    })
    .catch(error => {
      console.error(error.message);
      process.exit(1);
    });
});
```
where I have nested `.then` methods.
@underscope what do you think?

Issue #799, opened 2021-03-29
Unable to delete an activity with a teaching element inside of it


kjuej:
## I'm submitting a...

- [x] Bug report  
- [ ] Regression (a behavior that used to work and stopped working in a new release)
- [ ] Feature request
- [ ] Documentation issue or request

## Context
(Please see the attached screenshots)
Steps to recreate:
1. Add a new Sample Course
2. Enter the sample course and add a competency activity
3. Enter the competency and add a html element
4. Exit to repository structure and try to delete the competency - Notice the issue occurs here

<img width="1514" alt="Screen Shot 2021-03-29 at 15 24 54" src="https://user-images.githubusercontent.com/34127147/112843847-53818c00-90a3-11eb-9e38-76f885e50451.png">
<img width="1680" alt="Screen Shot 2021-03-29 at 15 24 27" src="https://user-images.githubusercontent.com/34127147/112843860-58464000-90a3-11eb-85b1-79babe6b84a5.png">

PR #800, opened 2021-03-30
Fix broken activity delete action, hide discussion on empty progress page

MiroDojkic:
### This PR:
- fixes broken activity delete action (#799)
- state is hydrated with a single SSE after bulk activity update; previously SSE was sent per item
- discussion is hidden in the progress page's sidebar when there's no selected activity

⚠️ Issue was caused by Sequelize's `individualHooks` flag in the model methods. When used, it loads instances with the appropriate scopes before running the `update` query. In this case, it caused the bug which would assign included model's alias (from the default scope) to `update` query's column.

#### :mag: QA note:
This change affects functionalities that save data to the server (creating and updating elements, repositories, etc.)

### DIFF
--- a/client/store/helpers/mutations.js
+++ b/client/store/helpers/mutations.js
@@ -30,6 +30,10 @@ export const save = (state, model) => {
   Vue.set(state.items, model.uid, model);
 };
 
+export const saveAll = (state, models) => {
+  each(models, it => save(state, it));
+};
+
 export const remove = (state, models) => {
   models.forEach(it => Vue.delete(state.items, it.uid));
 };
--- a/client/store/helpers/mutations.js
+++ b/client/store/helpers/mutations.js
@@ -1,3 +1,4 @@
+import castArray from 'lodash/castArray';
 import each from 'lodash/each';
 import omit from 'lodash/omit';
 import uuid from '@/utils/uuid';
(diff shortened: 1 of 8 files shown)


client/store/helpers/mutations.js
```diff
@@ -30,6 +30,10 @@ export const save = (state, model) => {
   Vue.set(state.items, model.uid, model);
 };
 
+export const saveAll = (state, models) => {
```
underscope: Consider casting model on save to array and have save do what is now part of save & saveAll

server/activity/hooks.js
```diff
@@ -27,6 +28,12 @@ function add(Activity, Hooks, Models) {
     sse.channel(activity.repositoryId).send(Events.Update, activity);
   }
 
+  async function sseBulkUpdate(_, { where }) {
+    const activities = await Models.Activity.findAll({ where });
+    const [activity] = activities;
+    sse.channel(activity.repositoryId).send(Events.BulkUpdate, activities);
```
underscope: Should we make sure that all have same repositoryId (groupBy and then do multiple bulk updates; for future sake)

server/activity/hooks.js
```diff
@@ -51,3 +58,8 @@ function add(Activity, Hooks, Models) {
     return outlineActivity && outlineActivity.touch(transaction);
   }
 }
+
+const afterTransaction = method => (type, opts) => {
```
underscope, line 65: 🏆

server/activity/hooks.js
MiroDojkic: Good idea! 👍

kjuej:
@MiroDojkic @underscope Seems to be functioning as expected!

PR #803, opened 2021-03-31
Implement user name format validation

lukasaric:
In order to improve name validation, this PR adds:
- `name_format` custom validation
- Validation uses [Unicode regex](https://www.regular-expressions.info/unicode.html)
- Integrates it where it's needed (user first & last names)

🔍  **QA note**:

Check all name input validations (e.g first and last name)


client/utils/validation.js
```diff
@@ -6,6 +6,14 @@ import { messages } from 'vee-validate/dist/locale/en.json';
 import some from 'lodash/some';
 import userApi from '@/api/user';
 
+const NAME_FORMAT_REGEX = /^[\p{Letter}\s'-.]+$/u;
+
+const nameFormat = {
+  validate: value => NAME_FORMAT_REGEX.test(value),
+  message: `The {_field_} field may only contain alphabetic characters, spaces,
+    dots (.), apostrophes (') and hyphens (-)`
```
underscope, line 14: @malizeus Why dots 🤔

lukasaric: Input names can be written in abbreviated forms (e.g Cuba Gooding Jr., J. K. Rowling ...) 😅

kjuej:
@malizeus Some improvement suggestions:
1. Dots apostrophes and hyphens should not appear in the beginning or end of the name
2. There should not be more than 1 hyphen/apostrophe/dot in a row.

Issue #808, opened 2021-03-31
Add status colors on the `Status` dropdown list 


bubafinder:
## I'm submitting a...

- [ ] Bug report  
- [ ] Regression (a behavior that used to work and stopped working in a new release)
- [x] Feature request
- [ ] Documentation issue or request

## Context
Add status colors on the `Status` dropdown list as on the `Filter by status` filter dropdown list

`Status` dropdown list
<img src="https://user-images.githubusercontent.com/41724080/113136703-f5c87d80-9223-11eb-90c0-661a8a39d447.png" width="400px">

`Filter by status` filter dropdown list
<img src="https://user-images.githubusercontent.com/41724080/113136914-31fbde00-9224-11eb-8843-91becf4bfad3.png" width="400px">

PR #811, opened 2021-04-01
Add common workflow status selection

lukasaric:
- This PR addresses #808
- It adds common workflow status selection which is used as `status fiter` & `status selection` in sidebar.

### DIFF
--- a/client/components/repository/Workflow/Filters/index.vue
+++ b/client/components/repository/Workflow/Filters/index.vue
@@ -12,7 +12,9 @@
       @change="updateFilter('status', $event)"
       :items="statusOptions"
       :value="status"
+      item-text="label"
       placeholder="Filter by status"
+      clearable
       class="status-field ml-7 mr-3" />
     <assignee-filter
       v-if="assigneeOptions"
@@ -35,7 +37,7 @@
 
 <script>
 import AssigneeFilter from './Assignee';
-import StatusFilter from './Status';
+import StatusFilter from '../Status';
 
 export default {
   name: 'workflow-filters',
--- a/client/components/repository/Workflow/Filters/index.vue
+++ b/client/components/repository/Workflow/Filters/index.vue
@@ -8,7 +8,7 @@
       placeholder="Search by ID or name"
       clearable
       class="search-field" />
-    <status-filter
+    <select-status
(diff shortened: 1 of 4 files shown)


client/components/repository/Workflow/Filters/index.vue
```diff
@@ -12,7 +12,9 @@
       @change="updateFilter('status', $event)"
       :items="statusOptions"
       :value="status"
+      item-text="label"
```
MiroDojkic: Should we make this a default value for `StatusFilter`'s `ItemText` prop (along with `ItemValue`)?
If we change the `Status` model in the future, we'd need to change default value _only_ in the `StatusFilter` component.

client/components/repository/Workflow/Status.vue
```diff
@@ -1,20 +1,20 @@
 <template>
-  <v-select v-on="$listeners" v-bind="$attrs" clearable>
+  <v-select v-on="$listeners" v-bind="$attrs">
```
MiroDojkic: Since there's no more `Filters` context coming from a parent directory, should we change the name of this component to `SelectStatus`?

MiroDojkic (changes requested):

client/components/repository/Workflow/SelectStatus.vue
```diff
@@ -1,20 +1,26 @@
 <template>
-  <v-select v-on="$listeners" v-bind="$attrs" clearable>
+  <v-select v-on="$listeners" v-bind="{ ...$attrs, itemText, itemValue }">
     <template #selection="{ item }">
       <div class="selection d-flex flex-nowrap">
         <v-icon :color="item.color" small class="mr-2">mdi-circle</v-icon>
-        <span class="text--default text-truncate">{{ item.text }}</span>
+        <span class="text--default text-truncate">{{ item.label }}</span>
```
MiroDojkic: If the `itemValue` is provided and its value is not `label`, will this break?

If that's the case, we could fix it with:
```suggestion
        <span class="text--default text-truncate">{{ item[itemText] }}</span>
```

MiroDojkic (changes requested):

MiroDojkic (approved):

Issue #804, opened 2021-03-31
User is not redirected from login page if already logged in


bubafinder:
## I'm submitting a...

- [x] Bug report  
- [ ] Regression (a behavior that used to work and stopped working in a new release)
- [ ] Feature request
- [ ] Documentation issue or request

## Context
1. Login to the platform
2. Go to the login page
[server-url]/#/login

Issue: The user is not redirected from the Login page to the home page (Catalog page).

![image](https://user-images.githubusercontent.com/41724080/113129503-425b8b00-921b-11eb-9b28-ba345f4dae06.png)

PR #812, opened 2021-04-01
Fixed login route redirect for already logged users

lukasaric:
- This PR addresses #804
- It properly redirects already logged user to root route (catalog) who's trying to access login route

### DIFF
--- a/client/router.js
+++ b/client/router.js
@@ -121,18 +121,18 @@ export default function getRouter() {
   const router = new Router(options);
   router.beforeEach((to, from, next) => {
     const { user } = router.app.$store.state.auth;
-    if (requiresAuth(to) && !user) {
+    if (isAuthRequired(to) && !user) {
       const query = { redirect: to.fullPath };
       return next({ path: '/login', query });
     }
-    if (!requiresAuth(to) && user) next({ path: '/' });
+    if (!isAuthRequired(to) && user) next({ path: '/' });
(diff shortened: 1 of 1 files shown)


underscope:
@malizeus This assumes that we can't have a public page down the road that logged-in users can access. This also introduces a bug upon clicking on forgot password btn on the profile page.

client/router.js
```diff
@@ -120,20 +120,24 @@ const options = {
 export default function getRouter() {
   const router = new Router(options);
   router.beforeEach((to, from, next) => {
-    const auth = router.app.$store.state.auth;
-    if (isAuthRequired(to, auth)) {
-      return next({ path: '/login', query: { redirect: to.fullPath } });
-    }
-    return isAllowed(to, auth) ? next() : next({ path: from.fullPath });
+    const { user } = router.app.$store.state.auth;
+    if (isAuthRequired(to, user)) return redirectToLogin(to, next);
```
underscope: This might be a bit clearer if isAuthRequired would just check is auth required and not check if a user is logged in 😉  💅

client/router.js
```diff
@@ -120,20 +120,24 @@ const options = {
...
+    if (to.name === 'login' && user) return next({ path: '/' });
+    return !isAllowed(to, user) ? next({ path: from.fullPath }) : next();
```
underscope: Why adding the `!` and moving the other condition first 🤔 ?

PR #815, opened 2021-04-02
Integrate reset token validation on reset password landing

lukasaric:
This PR:
- Checks if `reset token` is valid when user lands on `reset password` page.
- Both, reset password & token validation routes, now are on `base` axios instance to avoid location replacement


server/user/index.js
```diff
@@ -1,18 +1,22 @@
...
 const { User } = require('../shared/database');
 
+// Public routes:
 router
-  // Public routes:
   .post('/login', authenticate('local', { setCookie: true }), ctrl.getProfile)
   .post('/forgot-password', ctrl.forgotPassword)
   .post('/reset-password', authenticate('token'), ctrl.resetPassword)
-  // Protected routes:
+  .post('/reset-token-validation', authenticate('token'), (_, res) => res.sendStatus(ACCEPTED));
```
underscope: would it make sense for this to be nested under the reset-password route (i.e. `/reset-password/token-status` or similar). Add rate limiting.

client/store/modules/auth/actions.js
```diff
@@ -20,6 +20,8 @@ export const resetPassword = (_, { token, password }) => {
   return api.resetPassword(token, password);
 };
 
+export const validateResetToken = (_, { token }) => api.validateResetToken(token);
```
underscope: consider moving it to `api/auth`

client/components/auth/ResetPassword.vue
```diff
@@ -1,77 +1,93 @@
...
+        </validation-provider>
+        <v-btn
+          type="submit"
+          color="primary darken-4"
+          rounded block dark
+          class="my-1">
+          Change password
+        </v-btn>
+      </template>
+      <router-link v-else :to="{ name: 'forgot-password' }" class="forgot-password">
```
underscope: Consider rewriting this in a way you are not toggling part of the form (this should be outside).

client/components/auth/ResetPassword.vue
```diff
@@ -56,22 +61,32 @@
...
   methods: {
-    ...mapActions(['resetPassword']),
     submit() {
-      const { token } = this.$route.params;
-      return this.resetPassword({ password: this.password, token })
+      const { token, password } = this;
+      return api.resetPassword(token, password)
         .then(() => this.$router.push('/'))
-        .catch(() => (this.error = 'An error has occurred!'));
+        .catch(() => (this.error = INVALID_TOKEN_ERROR));
```
underscope: Why an invalid token error at this point 🤔 ?

client/components/auth/ResetPassword.vue
```diff
@@ -81,4 +96,8 @@ export default {
   padding-right: 0.25rem;
   background: #ececec;
 }
+
+.forgot-password .v-icon {
```
underscope: consider using icon size property

Issue #805, opened 2021-03-31
Disable adding repository with an already used name


bubafinder:
## I'm submitting a...

- [ ] Bug report  
- [ ] Regression (a behavior that used to work and stopped working in a new release)
- [ ] Feature request
- [ ] Documentation issue or request
- [x] Enhancement 

## Context
1. Go to the Catalog page
2. Try to add the repository with an already used name

The user can add the repository with an already used name. Introduce a warning message to the User that a Course with that name already exists.

![image](https://user-images.githubusercontent.com/41724080/113130042-e5140980-921b-11eb-8bee-49f273f5125e.png)

underscope:
Changing the label since it is intended behavior rather than an issue. Let's chat about it.

bubafinder:
Adding @Amer21214 to the discussion.

Amer21214:
If we will allow new Courses to be created with the name of a pre-existing Course, I really do suggest we introduce a warning message/validation to the User that a Course with that name already exists. The benefit of this action is worth the minor amount of work needed to implement specified behavior.

lukasaric:
@markolovric @Amer21214  Should we display warning message on repository (course) settings details while user is editing repo name?

Amer21214:
@malizeus - place it right underneath the repository name field, please.

underscope:
@markolovric @Amer21214 Should this happen when the user does not see an existing repository (with the same name; due to not having access to it). Will it be confusing? 🤔

bubafinder:
I think it should not be confusing. The user needs to know that there is a repository with a duplicate name. Also, I requested Luka to skip soft-deleted repositories.

Amer21214:
@underscope I wasn't aware that there could be courses visible to one User and not to another User. The underlying goal here is to avoid having any User see 2 Courses with the same name. 
This, for instance, includes the case where any "sub-User" creates a Course with a name that exists in the Course List of another sub-User, but both Courses would be visible in the list to an Admin/Instructor.

PR #817, opened 2021-04-03
Add existing repository name warning message

lukasaric:
- This PR addresses #805
- It gives user warning message when user enters already existing repository name.

- [x] Add warning message within repository creation name field
- [x] Add warning message within repository settings name field


client/api/repository.js
```diff
@@ -3,6 +3,7 @@ import request from './request';
 
 const urls = {
   root: '/repositories',
+  names: () => `${urls.root}/names`,
```
underscope: No need for an extra route this is repository listing with query by name call (just find exact match)

client/components/catalog/Add.vue
```diff
@@ -118,13 +125,17 @@
...
+import debounce from 'lodash/debounce';
 import loader from '@/components/common/loader';
 import { mapGetters } from 'vuex';
 import { SCHEMAS } from 'shared/activities';
+import some from 'lodash/some';
 import TailorDialog from '@/components/common/TailorDialog';
 
 const NEW_TAB = 0;
 
+const EXISTING_REPO_MESSAGE = '⚠️ Warning: a Repository with that name already exists.';
```
underscope: Why doing icon like this 🤔 ?

client/components/repository/Settings/General.vue
```diff
@@ -33,27 +33,37 @@ import find from 'lodash/find';
 import { getRepositoryMetadata } from 'shared/activities';
 import Meta from 'tce-core/MetaInput';
 import set from 'lodash/set';
+import some from 'lodash/some';
+
+const EXISTING_REPO_MESSAGE = '⚠️ Warning: a Repository with that name already exists.';
```
underscope: Repeating this here and convoluting meta feels wrong, we should find a better way of handling this (perhaps a custom component for the name that will be used here and within the creation modal). Ideally, we would extend vv to show warnings.

client/components/catalog/Add.vue
```diff
@@ -170,13 +183,19 @@ export default {
...
       this.archive = null;
       this.serverError = '';
+      this.existingRepoWarning = '';
     }
   },
   watch: {
-    isVisible(val) {
+    async isVisible(val) {
       if (!val) return setTimeout(() => this.$refs.form.reset(), 60);
+      this.repositoryNames = await api.getRepositoryNames();
```
underscope: 🤔 Why fetching all before?

underscope (changes requested):

Issue #818, opened 2021-04-06
Half width element turns to full width after input


kjuej:
## I'm submitting a...

- [x] Bug report  
- [ ] Regression (a behavior that used to work and stopped working in a new release)
- [ ] Feature request
- [ ] Documentation issue or request

## Context
![HalfWidthIssue](https://user-images.githubusercontent.com/34127147/113690306-8b07be00-96cb-11eb-9f53-947e754b6d89.gif)

underscope:
@malizeus  Please take a look as soon as you get a chance 😉

kjuej:
@malizeus @underscope Issue seems to still occur on half width:
1. Image element
2. Deprecated html element
3. Embed element

PR #820, opened 2021-04-08
Half width element issues fix & name format validation message enhancement

lukasaric:
This Pr addresses #818
It fixes half width issue on remaining content elements:

- Image
- Embed
- Deprecated HTML (text)

Bonus:
- Simplified `name_format` validation message


kjuej:
@malizeus @underscope Seems to be working as expected!

Issue #821, opened 2021-04-08
Element toolbar stays open after switching to new activity


kjuej:
## I'm submitting a...

- [x] Bug report  
- [ ] Regression (a behavior that used to work and stopped working in a new release)
- [ ] Feature request
- [ ] Documentation issue or request

## Context
Notice when clicking an image element then switching the activity, the image element toolbar stays open.
![ToolbarIssue](https://user-images.githubusercontent.com/34127147/113999454-5a09c380-985a-11eb-83e4-66e9709659c8.gif)

markolovric:
Looks good 👍

PR #822, opened 2021-04-08
Hide element toolbar on activity/editor change

lukasaric:
- This PR addresses #821
- It clears `selected element` after activity/editor changes which automatically removes editor toolbar.

🔍  QA Note:
- Test every action thoroughly which can be affected after `editor` route (id) is changed


bubafinder:
Looks good 👍

PR #825, opened 2021-04-12
Improve tracker for active users

ikovac:
- [x] Add heartbeat on the client-side. The client periodically pings the server.
- [x] Set expiration time on the server
- [x] Implement tapster
- [x] Fix endless active user connections bug


underscope:
@kjuej @markolovric

PR #834, opened 2021-04-27
Automation - Initial configuration

bubafinder:
Introducing Cypress to Tailor:
- initial configuration
- two basic tests added (running on port 3000)


cypress/fixtures/credentials.json
```diff
@@ -0,0 +1,5 @@
+{
```
underscope, line 1: We can pick it up from seed to make it less prone to future changes.

cypress/framework/selectors/login.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+const alertContent = '//div[@class="v-alert__content"]';
```
underscope, line 3: https://docs.cypress.io/guides/references/best-practices
@MiroDojkic @kronicker @vladimyr  Should we conform to => https://docs.cypress.io/guides/references/best-practices#Selecting-Elements

vladimyr: Mr React gave us the answers we need, no need for second thoughts 😄 
https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change#whats-with-the-data-testid-query

underscope: 'Mr React' is using testing-library https://testing-library.com/docs/queries/bytestid.

`
Using data-testid attributes do not resemble how your software is used and should be avoided if possible. That said, they are way better than querying based on DOM structure or styling css class names.
`

Everything makes purrfect sense now 🤡

underscope:
@markolovric Closing this one in favor of avoiding Page Objects and xpath 😉

PR #836, opened 2021-05-01
Implement max login attempts

ikovac:
- [x] Use [Tapster](https://github.com/ExtensionEngine/tapster) cache manager for request limiter default store
- [x] Implement login request limiter
- [x] Handle 429 (Too many requests) status code on the client side

Initially, I set max retries to 10, maybe we can open discussion to find out the best max retries value for our case. Max retries count is reset on successful login.

### DIFF
--- a/client/components/auth/Login.vue
+++ b/client/components/auth/Login.vue
@@ -76,6 +76,8 @@
 import { mapActions } from 'vuex';
 
 const LOGIN_ERR_MESSAGE = 'The email or password you entered is incorrect.';
+const TOO_MANY_REQ_CODE = 429;
+const TOO_MANY_REQ_ERR_MESSAGE = 'Too many requests. Please try again later.';
 const getOidcErrorMessage = (email, buttonLabel) =>
   `Account with email ${email} does not exist.
   Click "${buttonLabel}" to try with a different account.`;
@@ -104,7 +106,12 @@ export default {
       this.message = '';
       this.login({ email: this.email, password: this.password })
         .then(() => this.$router.push('/'))
-        .catch(() => (this.localError = LOGIN_ERR_MESSAGE));
+        .catch(err => {
+          const code = err?.response?.status;
+          this.localError = code === TOO_MANY_REQ_CODE
+            ? TOO_MANY_REQ_ERR_MESSAGE
+            : LOGIN_ERR_MESSAGE;
+        });
     }
   }
 };
--- a/client/components/auth/Login.vue
+++ b/client/components/auth/Login.vue
@@ -77,7 +77,7 @@ import { mapActions } from 'vuex';
 
 const LOGIN_ERR_MESSAGE = 'The email or password you entered is incorrect.';
 const TOO_MANY_REQ_CODE = 429;
-const TOO_MANY_REQ_ERR_MESSAGE = 'Too many requests. Please try again later.';
+const TOO_MANY_REQ_ERR_MESSAGE = 'Too many login attempts. Please try again later.';
 const getOidcErrorMessage = (email, buttonLabel) =>
   `Account with email ${email} does not exist.
   Click "${buttonLabel}" to try with a different account.`;
--- a/server/shared/request/mw.js
+++ b/server/shared/request/mw.js
@@ -32,7 +32,7 @@ class Store {
   }
 
   resetKey(key) {
-    this.cache.delete(key);
+    return this.cache.delete(key);
   }
 }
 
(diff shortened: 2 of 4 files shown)


client/components/auth/Login.vue
```diff
@@ -76,6 +76,8 @@
 import { mapActions } from 'vuex';
 
 const LOGIN_ERR_MESSAGE = 'The email or password you entered is incorrect.';
+const TOO_MANY_REQ_CODE = 429;
+const TOO_MANY_REQ_ERR_MESSAGE = 'Too many requests. Please try again later.';
```
MiroDojkic: Is the "login attempts" more user-friendly term than "requests"?

server/shared/request/mw.js
```diff
@@ -1,11 +1,48 @@
...
+  constructor() {
+    this.cache = new Tapster({
+      ...options[provider],
+      store: provider,
+      namespace: 'request-limiter'
+    });
+  }
+
+  async incr(key, cb) {
+    const record = await this.cache.has(key) ? await this.cache.get(key) : { hits: 0 };
```
MiroDojkic: This looks lengthy, please consider breaking it into multiple lines.

MiroDojkic: Would aliasing `const initialState = { hits: 0 };` make it a slightly bit easier to read? :bowtie:

server/shared/request/mw.js
```diff
@@ -1,11 +1,48 @@
...
+    const hits = record.hits + 1;
+    await this.cache.set(key, { ...record, hits });
+    cb(null, hits);
+  }
+
+  async decrement(key) {
+    const record = await this.cache.get(key);
+    if (!record || !record.hits) return;
+    const hits = record.hits - 1;
+    await this.cache.set(key, { ...record, hits });
```
MiroDojkic: Why not just return this promise instead of awaiting it?

server/shared/request/mw.js
```diff
@@ -1,11 +1,48 @@
...
+
+  resetKey(key) {
+    return this.cache.delete(key);
+  }
+}
+
+function requestLimiter({
+  max = 10,
+  windowMs = DEFAULT_WINDOW_MS,
+  store = new Store(),
```
MiroDojkic: Why create a new connection each time this middleware factory is invoked vs provide a singleton as default?

server/user/index.js
```diff
@@ -33,3 +43,14 @@ module.exports = {
...
+function getKeyFromRequest(req, res, next) {
+  const key = [req.ip, req.body.email].join(':');
+  req.key = crypto.createHash('sha256').update(key).digest('base64');
+  return next();
+}
+
+function resetLoginAttempts(req, res, next) {
+  return loginRequestLimiter.resetKey(req.key)
+    .then(() => next());
+}
```
MiroDojkic: Consider moving these to `server/user/mw.js`.

server/user/mw.js
```diff
@@ -0,0 +1,18 @@
...
+
+const crypto = require('crypto');
+
+function getKeyFromRequest(req, res, next) {
+  const key = [req.ip, req.body.email].join(':');
+  req.key = crypto.createHash('sha256').update(key).digest('base64');
+  return next();
+}
+
+function resetLoginAttempts(requestLimiter) {
```
MiroDojkic: Consider moving `loginRequestLimiter` to this file - it's a user-specific middleware so this looks like the place for it and we this factory can become a simple middleware.

server/user/mw.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const crypto = require('crypto');
+
+function getKeyFromRequest(req, res, next) {
+  const key = [req.ip, req.body.email].join(':');
+  req.key = crypto.createHash('sha256').update(key).digest('base64');
```
MiroDojkic: Should we prefix `req.key` with something to reduce the chance of collision and make it easier to read (i.e. `req.userKey`)?

server/shared/request/mw.js
```diff
@@ -1,11 +1,53 @@
...
+
+  async incr(key, cb) {
+    const initialState = { hits: 0 };
+    const record = await this.cache.has(key)
+      ? await this.cache.get(key)
+      : initialState;
+    const hits = record.hits + 1;
+    await this.cache.set(key, { ...record, hits });
+    cb(null, hits);
+  }
```
MiroDojkic: Consider if it's worth it to make this and `decrement` a bit more concise:
```suggestion
  async incr(key, cb) {
    const initialState = { hits: 0 };
    const { hits, ...record } = await this.cache.has(key)
      ? await this.cache.get(key)
      : initialState;
    await this.cache.set(key, { ...record, hits: hits + 1 });
    cb(null, hits);
  }
```

MiroDojkic (changes requested):

PR #851, opened 2021-05-06
Change avatar border color on active state

### LINKED ISSUE #840: Upon filtering by user, additional border to user avatar is added
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request

## Context
On the Progress (workflow) page, upon filtering by the user (clicking on user avatar), an additional border to the user avatar is added. Remove the inner border.

image

ikovac:
Closes #840.

This PR also fixes a minor bug introduced [here](https://github.com/ExtensionEngine/tailor/pull/846/files#diff-582fe5b3c853a6d2a182275286382bcec959a4856611c74d303b07198e88583aR27) where the color prop is boolean instead of string.

### DIFF
--- a/client/App.vue
+++ b/client/App.vue
@@ -58,7 +58,7 @@ html {
 }
 
 .v-main {
-  background-color: #f5f5f5;
+  background-color: $main-background-color;
 }
 
 .v-main .view {
--- a/client/App.vue
+++ b/client/App.vue
@@ -58,7 +58,7 @@ html {
 }
 
 .v-main {
-  background-color: $main-background-color;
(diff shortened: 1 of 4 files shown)


MiroDojkic:
@ikovac @underscope @markolovric this seems to show the wrong (old) border color when the avatar filter is not selected/active - border color should always be the same as the background color, right?

ikovac:
@MiroDojkic we decided to keep the old border color when the avatar is not active and only remove the old color when the avatar is active. In the second case, we are showing border color which is the same as the background color.

PR #855, opened 2021-05-10
Cypress init

underscope:
https://user-images.githubusercontent.com/6253820/119985916-8cff4680-bfc3-11eb-80e9-745df00293df.mov


cypress/support/commands.js
```diff
@@ -23,3 +23,13 @@
 //
 // -- This will overwrite an existing command --
 // Cypress.Commands.overwrite('visit', (originalFn, url, options) => { ... })
+
+Cypress.Commands.add('loginWithUI', () => {
```
Amer21214: We're trying to avoid using Cypress commands as our go-to for page interaction, as this structure is Cypress's way of opposing Selenium's Page Object Model (which is built in the /framework). From there, a set of commands under 1 function like loginWithUI can be made.

underscope: Not sure I completely follow, lets touch base after I finish with the POC 😉

Amer21214: Ok, no problemo :)

PR #856, opened 2021-05-10
Run remove-user-token migration in a transaction

vkaracic:
In the case where there are empty passwords the `changeColumn` command
for the `password` column would fail, but the `token` column would be
deleted causing `column "token" of relation "user" does not exist`
errors on new migration runs.


kronicker (approved):
👌

underscope:
Please base it on release/5.0 and reopen 😉

vkaracic:
Done: https://github.com/ExtensionEngine/tailor/pull/857

PR #859, opened 2021-05-21
Force the POSIX path separators in API routes

2Pacalypse-:
`path.join()` on its own will use a path separator of whatever the OS
the code is currently running on (e.g. "/" on Unix, "\\" on Windows). API
routes should always use a forward-slash as its path separator, and this
is what is done in this PR.

There are probably a couple of more instances in the codebase which
should always force the POSIX path separator, but I'll fix those once I
encounter them as I'm not sure where they are off the top of my head.


vladimyr (approved):
Good catch 👍
Also, I love the commit message ❤️

PR #872, opened 2021-06-02
OIDC: Return href instead URL object

zcuric:
### This PR
- [x] Fixes issues with OIDC which return URL object instead of string and it would fail with silent auth and other cases


kronicker (approved):

kronicker:
@zcuric Thanks for raising PR 
@underscope This should be merged ASAP since it causes issues with silent auth in production.

underscope:
@zcuric Base should be release/5.0

zcuric:
@underscope done! 🤝

PR #892, opened 2021-06-15
Implement email notifications unsubscribe

hrvojevu:
## This PR:
- updates user model with new JSONB `notifications` column which contains user notification preferences
- updates user settings form to support comment and assignment notification subscription
Screen Shot 2021-06-15 at 3 07 05 PM
- updates comment after create hook to check which users have comment notifications enabled
- updates activity status hooks to check if user has assignee notifications disabled before sending notification
- updates comment and assignee email notification templates to include profile page link
Screen Shot 2021-06-16 at 2 25 04 PM

Screen Shot 2021-06-16 at 2 24 13 PM

### DIFF
--- a/server/comment/hooks.js
+++ b/server/comment/hooks.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const mail = require('../shared/mail');
-const map = require('lodash/map');
 const pick = require('lodash/pick');
 const { schema } = require('@tailor-cms/config');
 const sse = require('../shared/sse');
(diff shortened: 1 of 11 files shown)


server/comment/hooks.js
```diff
@@ -80,7 +79,10 @@ exports.add = (Comment, Hooks, db) => {
       action: isCreate ? 'left' : 'updated',
       ...pick(comment, ['id', 'content', 'createdAt'])
     };
-    const collaborators = map(repository.repositoryUsers, 'user.email');
+    const collaborators = repository.repositoryUsers.reduce((acc, { user }) => {
```
2Pacalypse-, line 82: Excellent usage of the `reduce` method. Bravo!

PR #893, opened 2021-06-16
Allow storage proxy to work with IPv4

droguljic:
Storage proxy signed cookies would get assigned to wrong domain if
configured host name is IPv4 address, thus blocking display of assets.
Problem is in `psl` package which parses IPv4 addresses, thus producing
wrong domains.
Fix detects IPv4 addresses using regular expression and just returns
`null` for domain.

### DIFF
--- a/server/shared/storage/proxy/mw.js
+++ b/server/shared/storage/proxy/mw.js
@@ -2,23 +2,18 @@
 
 const config = require('../../../../config/server');
 const { FORBIDDEN } = require('http-status-codes');
+const isIp = require('is-ip');
 const miss = require('mississippi');
 const path = require('path');
 const router = require('express').Router();
 const psl = require('psl');
 
-const IPV4_REGEX = /^((25[0-5]|(2[0-4]|1[0-9]|[1-9]|)[0-9])(\.(?!$)|$)){4}$/;
-
 function getDomain() {
-  if (IPV4_REGEX.test(config.hostname)) return null;
+  if (isIp.v4(config.hostname)) return null;
   return psl.parse(config.hostname).domain;
 }
 
 module.exports = (storage, proxy) => {
-  if (IPV4_REGEX.test(config.hostname) && config.storage.proxy.provider === 'cloudfront') {
-    throw new Error('CloudFront storage proxy cannot be used alongside IPv4 host name');
-  }
-
   function getFile(req, res, next) {
     const key = req.params[0];
     const hasValidCookies = proxy.verifyCookies(req.cookies, key);
--- a/server/shared/storage/proxy/mw.js
+++ b/server/shared/storage/proxy/mw.js
@@ -15,6 +15,10 @@ function getDomain() {
 }
 
 module.exports = (storage, proxy) => {
+  if (IPV4_REGEX.test(config.hostname) && config.storage.proxy.provider === 'cloudfront') {
+    throw new Error('CloudFront storage proxy cannot be used alongside IPv4 host name');
+  }
+
   function getFile(req, res, next) {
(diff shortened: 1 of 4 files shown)


underscope:
@droguljic Please base this on release/5.0 😉 Thx!

server/shared/storage/proxy/mw.js
```diff
@@ -7,6 +7,13 @@ const path = require('path');
 const router = require('express').Router();
 const psl = require('psl');
 
+const IPV4_REGEX = /^((25[0-5]|(2[0-4]|1[0-9]|[1-9]|)[0-9])(\.(?!$)|$)){4}$/;
+
+function getDomain() {
+  if (IPV4_REGEX.test(config.hostname)) return null;
+  return psl.parse(config.hostname).domain;
+}
```
MiroDojkic: Does this work only for the local proxy provider?

droguljic: Mostly, as domain is not specified and thus determined by viewer URL, which in turn won't expose signed cookies to CloudFront. 
There is possibility for this to work with CloudFront in following circumstances, host name is set to IPv4, viewer URL is different from host name and CloudFront CNAME option is set to match viewer URL, thus signed cookies are set on correct domain, but I doubt this as a real use case.
This fix mostly targets Local proxy, as without it local proxy cannot work when host name is IPv4.

MiroDojkic: @droguljic got it. Can we throw an error in case that the proxy is set to CloudFront and the hostname is IPV4 to warn developer that their configuration is invalid?

droguljic: @MiroDojkic done. Please see the latest commit.

MiroDojkic: @droguljic sorry for nitpicking, but should we move config validation to storage providers where the rest of the validation is?

droguljic: @MiroDojkic That was my initial plan, but host name is not available inside storage provider. Importing `config` inside storage provider didn't seem right, so I opted for proxy middleware as config is already available there.
Maybe I missed something, as I'm not that familiar with the codebase, so feel free to provide suggestion.

MiroDojkic: @droguljic since this is basically config validation, it seems to me that we could perform it inside the root server config file.

droguljic: @MiroDojkic inside this file https://github.com/ExtensionEngine/tailor/blob/release/5.0/config/server/index.js?
In order to put it here, logic for testing is something IPv4 needs to be extracted into separate function, probably into `server/shared/util/isIpv4`. Importing something from `server` into `config` seems a bit weird and copy pasting regular expression is not DRY, so I'm in dilemma again.

MiroDojkic: @droguljic yep. If it's not being used anywhere else, we could put it at the bottom of `config/server/index.js`. If the file becomes too bloated with noise, would it make sense to extract it to something like `config/server/validation.js`?

droguljic: @MiroDojkic it needs to be used in two places, `server/shared/storage/proxy/mw.js` and `config/server/index.js`, already is used in former and will be added to latter. Maybe adding new dependency [is-ip](https://www.npmjs.com/package/is-ip) is a better option and it'll definitely resolve this dilemma, so going with new dependency.

MiroDojkic: @droguljic good idea!

droguljic: @MiroDojkic done.

PR #897, opened 2021-06-17
Add ability to copy activity links in the editor sidebar

2Pacalypse-:
This PR adds a button in the editor sidebar that allows content authors
to easily copy links to other activities in the same repository. This
is especially useful when content authors want to reference a different
activity inside content elements.

The button is shown only on hover, and is shown for both editable and
non-editable activities. This might be a bit of a controversial
decision, but my thinking was that linking to non-editable activities
might be useful outside Tailor, i.e. when content is published. Also,
I feel like eventually, Tailor's editor itself might want to show
*something* inside a non-editable activity, e.g. contents of that
activity and/or options to add sub-activities to it.

### DIFF
--- a/client/components/editor/Sidebar/Navigation.vue
+++ b/client/components/editor/Sidebar/Navigation.vue
@@ -20,7 +20,7 @@
       :search="search"
       open-all
       class="pt-4">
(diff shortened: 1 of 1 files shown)


underscope:
Closing as this is being resolved on a fork

PR #901, opened 2021-06-29
Integrate lazy loading using `v-lazy` with data table in workflow overview

hrvojevu:
Integration virtual scroll with data table seems not possible at this moment, issue on vuetify github here https://github.com/vuetifyjs/vuetify/issues/3435. 

However, using `v-lazy` seems possible, with some issues arising, such as items not showing when sort (by whatever) is used on the data table. The problem seems to be with the `v-intersect` which doesn't detect that the element is in the view when sort is applied.


hrvojevu:
@markolovric please regression test extensively since the component has been change quite a bit.

PR #903, opened 2021-07-01
Set default scope in user model to exclude password field in results

### LINKED ISSUE #900: Omit user's password from assignee object and remove `v-bind` from `<overview-assignee`
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request

## Context
Using `v-bind` on `<overview-assignee />` in `client/components/repository/Workflow/Overview/index.vue` component causes user's properties to be visible as element attributes, including hashed user password. Additionally, user's id is visible as an attribute as well. This is unwanted behavior since we don't want incremental ids to be seen in the browser. Suggested fix: remove `v-bind` from `<overview-assignee />` in `client/components/repository/Workflow/Overview/index.vue` and omit user's password property from the server response

Screen Shot 2021-06-28 at 1 52 36 PM

## Expected Behavior
User's password and id shouldn't be seen in the element's attributes and shouldn't be returned from the server

## Current Behavior
User's password and id can be seen in the dev tools as an element attribute

hrvojevu:
### QA note:
- users password should not be visible in any result set that includes user information

### DIFF
--- a/client/components/auth/ForgotPassword.vue
+++ b/client/components/auth/ForgotPassword.vue
@@ -29,7 +29,7 @@
       </validation-provider>
       <div>
         <v-btn
-          :disabled="showMessage"
+          v-if="!showMessage"
           type="submit"
           color="primary darken-4"
           block depressed rounded dark>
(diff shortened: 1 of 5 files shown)


MiroDojkic (approved):

underscope:
@hrvojevu @markolovric Login/auth is broken (need to set .unscoped() or .scope(null) where pwd is required).

hrvojevu:
@markolovric please do regression testing on all matters regarding user flow: login, forgot password, change password etc.

underscope:
@markolovric Please revisit on release/5.0

PR #923, opened 2021-08-23
Fix: broken vue-radio package import

ikovac:
Fix broken vue-radio package import.


vladimyr (approved):
Thank you for turning our CI green again 🎉

PR #943, opened 2021-11-25
Issue with rate limiting and ttl window

underscope:
TTL was missing on the request limiter store (due to which state was not expiring over time).


bubafinder:
@underscope I noticed some issues when sending emails with SSL configured `EMAIL_SSL=1`. Emails are not received. Works as expected when `EMAIL_SSL=0`.
@kronicker gave his opinion maybe it is problem with the lib we are using. 
Should I raise an issue?

image

PR #968, opened 2022-05-24
Fix "reinvite" user action

### LINKED ISSUE #964: The reset password URL does not work correctly when the user is reinvited
## I'm submitting a...

- [x] Bug report  
- [ ] Feature request

## Expected Behavior
When a user is reinvited to Tailor, they will receive an email with a link that will lead them to the reset password page.

## Current Behavior
When a user is reinvited to Tailor, they receive an email with a link and the link leads them to the reset password page but there is an error with the URL.

## Failure Information (for bugs)
Screenshot 2022-05-13 at 11 44 33

mcapeta:
Fixes #964 

The user token from the URL is generated with the secret composed of server secret, user password, and user's "createdAt" timestamp ([source](https://github.com/ExtensionEngine/tailor/blob/develop/server/user/user.model.js#L217)). When the token is issued, the secret does not contain all of the above elements since the user instance was loaded without the password as per [the default model scope](https://github.com/ExtensionEngine/tailor/blob/develop/server/user/user.model.js#L116-L118). On the other hand, when the token is being checked, not the same secret is used because of fetching the "unscoped" user, i.e. user instance containing the password hash ([source](https://github.com/ExtensionEngine/tailor/blob/develop/server/shared/auth/index.js#L75)).

To ensure the same secret is used for both issuing and verifying the token, it's needed to load the "unscoped" user instance for creating the invitation email.

### DIFF
--- a/server/user/user.controller.js
+++ b/server/user/user.controller.js
@@ -64,7 +64,7 @@ function changePassword({ user, body }, res) {
 }
 
 function reinvite({ params }, res) {
-  return User.findByPk(params.id)
+  return User.unscoped().findByPk(params.id)
     .then(user => user || createError(NOT_FOUND, 'User does not exist!'))
     .then(user => User.sendInvitation(user))
     .then(() => res.status(ACCEPTED).end());


abasic:
@mcapeta I believe this issue is also appearing on the release/5.0 branch. So please make that branch the base of this PR.

underscope:
Thanks ❤️

PR #988, opened 2022-07-13
Fix footer's vertical alignment

marjev:
This PR fixes the vertical alignment within the footer found on `Repository settings` page (among other places).

The reason why this is broken lies in the fact that the `row` is `flexbox` container, but `column` is not, thus, flexbox alignment properties (attributes) aren't working as expected.

This is how it currently looks on develop:
Screenshot 2022-07-13 at 09 41 06

This fix achieves the following effect:
Screenshot 2022-07-13 at 09 42 14

which is closer to the screenshot which can be found in documentation:
https://docs.google.com/document/d/1TuMm4qektujv6kweFcgffFVjI4ogWXTbJXY70Sh9G34/edit#heading=h.j1odoclusuqy


mskorsur (approved):

client/components/common/Footer.vue
```diff
@@ -1,19 +1,15 @@
...
-        <v-chip
-          color="grey lighten-4"
-          label small
-          class="readonly mr-3 grey--text text--darken-4">
-          v{{ version }} {{ codename }}
-        </v-chip>
-        Built with <v-icon color="pink">mdi-heart</v-icon>
-        Extension Engine
-      </v-col>
+    <v-row align="center" justify="center" class="body-2 grey--text text--lighten-4">
```
underscope: Consider replacing with div and align/justify classes since we removed the [col](https://vuetifyjs.com/en/components/grids/#v-row) 😉

marjev: Addressed in be7cd4d598f1cabe3c8e6146101c333a3265b856

marjev: Accidental c/p fixed in 36c04c4c63fad725d05b3631c325b18972069733.

underscope:
@marjev The only thing I noticed is that ❤️ is now a bit closer/too close to letters. Consider adding `mx-1` to the icon 😉 Thanks!

marjev:
@underscope comment addressed in [492536b](https://github.com/ExtensionEngine/tailor/pull/988/commits/492536b8ff5cf021214c09120fc9ac38632a2665)

underscope:
Thanks!

PR #990, opened 2022-07-13
Fix icon code & Bump the version of @tailor-cms/core-components

marjev:
This PR is twofold:

1. Icon is fixed to match the code provided by the Material Design documentation.
2. Since this code lies in the `core-components` section, I've bumped the version of the `@tailor-cms/core-components` package and the main dependency.

@underscope please take a look to verify that you are able to reproduce the issue and that my fix will be correctly propagated through the `npm install`. Thanks!
cc @mskorsur 

Current state of things:
Screenshot 2022-07-13 at 13 20 49

Once the fix lands, it should look like this:
Screenshot 2022-07-13 at 13 07 08

Also fixes the missing icon here:
Screenshot 2022-07-13 at 14 17 14

Overall:
Screenshot 2022-07-13 at 14 18 20

### DIFF
--- a/packages/core-components/package.json
+++ b/packages/core-components/package.json
@@ -3,7 +3,7 @@
   "description": "Reusable Tailor UI components",
   "author": "ExtensionEngine <info@extensionengine.com>",
   "license": "MIT",
-  "version": "1.0.0-beta.6",
+  "version": "1.0.0-beta.7",
   "files": [
     "dist"
   ],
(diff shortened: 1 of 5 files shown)


packages/core-components/package.json
```diff
@@ -3,7 +3,7 @@
   "description": "Reusable Tailor UI components",
   "author": "ExtensionEngine <info@extensionengine.com>",
   "license": "MIT",
-  "version": "1.0.0-beta.6",
+  "version": "1.0.0-beta.7",
```
underscope: There are [versions](https://www.npmjs.com/package/@tailor-cms/core-components) published after this one, please bump it to .11 😉

marjev: Addressed in 6a8070e

underscope:
Thanks!

PR #1010, opened 2022-07-27
Test ability to delete repository and activity

bubafinder:


cypress/e2e/repository/settings/delete-repository.cy.js
```diff
@@ -0,0 +1,29 @@
...
+      .as('repository')
+      .then(({ id: repositoryId }) => {
+        cy.navigateTo({
+          name: 'repository-info',
+          params: { repositoryId }
+        });
+      });
+  });
+
+  it('should delete the repository from settings page', function () {
```
underscope: from `the` settings page 😉 Also I am not sure if `from` is the right preposition (sounds like we are removing it from that page rather than that being place where the action is triggered). Since there is only one place where this action is being triggered (and the test is already scoped to repository/settings), consider updating the desc to `should be able to delete the repository` or improving the desc in some other way.

cypress/e2e/repository/settings/delete-repository.cy.js
```diff
@@ -0,0 +1,29 @@
...
+    cy.findByRole('button', { name: 'delete' }).click();
```
underscope, line 21: You updated the same action in the create-repository.cy.js in terms of `click()` being moved to the next row. Let's pick one for consistency's sake and stick to it 😉

cypress/e2e/repository/settings/delete-repository.cy.js
```diff
@@ -0,0 +1,29 @@
...
+    cy.confirmAction('Delete repository?');
+    cy.assertRoute('catalog');
```
underscope, line 23: Can be omitted (if we are doing visit / and not waiting for app to do the routing after deletion) 😉

cypress/e2e/repository/settings/delete-repository.cy.js
```diff
@@ -0,0 +1,29 @@
...
+    cy.visit('/');
+    cy.get('@repository').then(({ name: repositoryName }) => {
```
underscope: `repository` is available on this. You can omit `cy.get` and have  something like `findRepositoryCard(this.repository.name)` 💅

cypress/e2e/repository/structure/utils.js
```diff
@@ -7,16 +7,29 @@ const sel = {
...
     cy.vSelect('Type', type);
-    cy.findByLabelText('Name').type(`${name}{enter}`);
+    cy.findByLabelText('Name').type(`${name}`);
+    cy.findByRole('button', { name: /create/i }).click();
+  });
+}
+
+export function createChildActivity(parentName, name, type) {
+  findActivityItem(parentName).as('parent');
+  cy.get('@parent').click()
```
underscope: Seems like you can remove the alias entirely ?

cypress/e2e/repository/structure/utils.js
```diff
@@ -7,16 +7,29 @@ const sel = {
...
+    .findByRole('button', { name: /add item into/i }).click();
+  return getActivityDialog().within(() => {
```
underscope: This piece is now duplicated, consider extracting it to another function 😉

cypress/e2e/repository/structure/activity-deletion.cy.js
```diff
@@ -0,0 +1,60 @@
...
+  createRootActivity,
+  findActivityItem,
+  generateActivityName
+} from './utils';
+
+const ROOT_ACTIVIY_TYPES = ['Module', 'Page'];
+const CHILD_ACTIVITY_TYPES = ['Module', 'Lesson', 'Knowledge check', 'Page'];
+
+const createdParent = [];
+const createdChild = [];
```
underscope: We need to replace this. Since it has some complexity, I'll do a separate PR or you can replace it with the seed that we created (needs to be done via alias)

cypress/e2e/repository/structure/activity-deletion.cy.js
```diff
@@ -0,0 +1,60 @@
...
+      clickActivityRemove(it);
+      cy.confirmAction('Delete item?');
+      findActivityItem(it).should('not.exist');
+    });
+  });
+});
+
+function clickActivityRemove(name) {
+  findActivityItem(name).click().within(() => {
+    cy.get('div[class="options-menu"]').click();
```
underscope: Use label

Issue #1009, opened 2022-07-26
Incorrect dialog & dropdown selector behavior


bubafinder:
## I'm submitting a...

- [x] Bug report  
- [ ] Regression (a behavior that used to work and stopped working in a new release)
- [ ] Feature request
- [ ] Documentation issue or request

## Context
The dialog used for adding activity into the repository (Add, Add into, and Add below) activates the scroll after the dropdown selector is opened. The whole dialog including the dropdown selector is scrollable inside the same default height with a grey background.

## Expected Behavior
After the dropdown selector is opened, options are extended outside the dialog, without activating the scroll.

## Screencast
![Kapture 2022-07-26 at 11 41 22](https://user-images.githubusercontent.com/41724080/180977300-9fe72548-bcf9-4426-8cd9-2db535a1b065.gif)

PR #1011, opened 2022-08-04
Fix issues with activity creation dialog

underscope:
- Set dialog min height
- Set type selection max height


bubafinder:
Solving issue #1009

PR #1063, opened 2023-03-31
Bump minor versions for all dependencies

jtoric:


package.json
```diff
@@ -142,15 +142,15 @@
     "sequelize": "^6.21.3",
```
droguljic: Why `sequelize` is not bumped to `6.30.0`?

server/app.js
```diff
@@ -24,13 +26,10 @@ const { STORAGE_PATH } = process.env;
...
 
 config.auth.oidc.enabled && (() => {
-  const consolidate = require('consolidate');
-  const session = require('express-session');
   app.engine('mustache', consolidate.mustache);
   app.set('view engine', 'mustache');
-  app.use(session(config.auth.session));
 })();
-
+app.use(session(config.auth.session));
```
droguljic: This change should be reverted. Making the whole app stateful, by default, is not how we should handle `passport`'s more explicit approach toward the session. Instead, we should opt to also be explicit when using stateless strategies, e.g. 'jwt', and state so, i.e. `{ session: false }`.

package.json
```diff
@@ -55,9 +55,9 @@
     "@tailor-cms/utils": "^1.0.0-beta.18",
     "@ungap/global-this": "^0.4.4",
     "auto-bind": "^3.0.0",
-    "aws-sdk": "^2.1192.0",
+    "aws-sdk": "^2.1347.0",
```
droguljic: We should test this change more thoroughly once AWS credentials are acquired.

droguljic (changes requested):

droguljic (changes requested):
This change broke e2e tests, we should find a cause and fix it.

package.json
jtoric: Bumped to 6.30.0

server/app.js
jtoric: Reverted this and added {session: false} to every function call in server/shared/auth/authenticator.js

package.json
jtoric: Tested the change, everything seems ok

server/shared/auth/authenticator.js
```diff
@@ -23,6 +23,7 @@ class Auth extends Authenticator {
   }
 
   authenticate(strategy, ...args) {
+    args[0] = { session: false, ...args[0] };
```
droguljic: This is not the correct place to set the `session` option, as there is no guarantee that `args[0]` represents options at this point. Once `parseAuthenticateOptions` is executed and options are assigned, we can set the `session` option.
Also, I would like to add a comment on why we are setting the `session` option, e.g. "The passport started to be explicit about the session from v0.6.0, and since most of our authentication strategies are stateless, we are opting to avoid session by default".

droguljic (changes requested):

server/shared/auth/authenticator.js
jtoric: Moved setting session to false from args to options

droguljic (approved):

PR #1067, opened 2023-04-25
Bump major versions for most of dependencies

jtoric:
Update all dependencies except those that can't be updated without making significant changes.

Dependencies that became pure ESM:
1. auto-bind
2. boxen
3. dargs
4. emailjs
5. exports-loader
6. hash-obj
7. humanize-string
8. path-exists
9. read-pkg-up
11. strip-json-comments
12. url-join
13. yn

Dependencies that have conflicting webpack version:
1. copy-webpack-plugin
2. imports-loader
3. sass-loader
4. val-loader

We haven't updated "stylelint" and "eslint-plugin-n" because we are dependent on "@extensionengine/stylelint-config" which doesn't yet support new versions of those packages.

We also haven't updated any dependencies regarding tailor-cms or Vue because they will be updated in different PR.


droguljic (approved):

bubafinder:
Steps to reproduce:
1. Delete `node_modules` folder
2. Install dependencies
3. Run the server

image

bubafinder:
When I try to Save any question element, validation is not visible and nothing happens

image

bubafinder:
I can't upload a file to the Video & Audio & PDF TE

image
image

droguljic:
Suppressed in favor of #1076

PR #1079, opened 2023-07-07
Update dependencies in packages

jtoric:
Updated all packages in dependencies.

Also fixed lint issues.


droguljic:
Suppressed in favor of #1084

PR #1112, opened 2023-12-12
Added Dockerfile for release/cepp

izderic:
* Added Dockerfile for release/cepp
* Moved `dargs`, `sequelize-cli` and `umzug` to regular dependencies to fix Module not Found errors
* Installed missing `mississippi` package


.dockerignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+.git
+.env
```
ikovac, line 3: We can also ignore the dist folder?

Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM node:14.21.3-slim@sha256:21b6dc452bd2c066724bd5be0f7256ef81a068c47de1adead10a2f810557de02 AS base
```
ikovac, line 1: Why are we not using bookworm-slim for the runner image?

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN apt update && apt install -y --no-install-recommends dumb-init
+ENTRYPOINT ["dumb-init", "--"]
+
+FROM node:14.21.3@sha256:2cafa3fbb0b6529ee4726b4f599ec27ee557ea3dea7019182323b3779959927f AS install
```
ikovac, line 5: We can use bookworm image as install image

package.json
```diff
@@ -122,6 +123,7 @@
     "pg-query-stream": "^3.2.0",
     "pluralize": "^8.0.0",
     "plyrue": "^2.1.4",
+    "poi": "^12.10.3",
```
ikovac: Please revert this change and install missing `mississipi` package (required in amazon storage file)

ikovac (changes requested):

Dockerfile
izderic: I could not find bookworm-slim for this node version.

Dockerfile
izderic: Looks like bookworm and bookworm-slim are available only in newer versions of node.

ikovac (approved):

abasic (approved):

PR #1127, opened 2024-02-13
Fix: Repository cloning

ikovac:
This PR fixes repository cloning when elements have refs to other elements from the repository.

Previous version:
References weren't changed after cloning

Current version:
References are correctly updated after cloning and point to newly created element


server/content-element/content-element.model.js
```diff
@@ -137,9 +140,22 @@ class ContentElement extends Model {
    */
   mapClonedReferences(mappings, transaction) {
     const { refs } = this;
-    if (!refs.objectiveId) return Promise.resolve();
-    refs.objectiveId = mappings[refs.objectiveId];
-    return this.update({ refs }, { transaction });
+    // TODO: Refactor this and extract common logic so it can be reused.
```
underscope, line 143: Is there a reason for not exporting it from that module?

server/repository/repository.model.js
```diff
@@ -115,12 +115,16 @@ class Repository extends Model {
       ContentElement.scope('withReferences').findAll(opts)
     ]);
     return Promise.join(
-      Promise.map(activities, it => it.mapClonedReferences(mappings, relationships, transaction)),
+      Promise.map(activities, it => it.mapClonedReferences(mappings.activity, relationships, transaction)),
```
underscope: Try to keep it within 80

server/repository/repository.model.js
```diff
@@ -115,12 +115,16 @@ class Repository extends Model {
...
       Promise.map(elements, it => it.mapClonedReferences(mappings, transaction)),
       (activities, elements) => ({ activities, elements })
     );
   }
 
+  // TODO: Remove this method from the model as there is no value in having it
+  // here versus having it inside service method.
+  // Clone logic should be the same as import logic and should be extracted
```
underscope: Not exactly the same, but it is a valid comment. Consider reframing to encapsulate same logic into service method.

underscope (approved):

server/repository/repository.model.js
ikovac: What behavior is different from import logic?

server/content-element/content-element.model.js
ikovac: The signatures were different and I tried to change as little as possible to don't break something I was not aware of. Also this logic wasn't shared in the first place, so with the upcoming refactor in mind I didn't want to refactor much within this bug fixing.
If you feel that it would be good to do that here, I can.

server/repository/repository.model.js
underscope: Discussed offline

.circleci/config.yml
```diff
@@ -70,7 +73,7 @@ jobs:
 workflows:
   lint-and-test:
     jobs:
-      - lint
+      - lint
```
underscope: 💅

underscope (approved):

rkusan00 (approved):

PR #1137, opened 2024-05-24
Update repo info on publish

rkusan00:
This PR updates course details upon publish

### DIFF
--- a/server/shared/publishing/helpers.js
+++ b/server/shared/publishing/helpers.js
@@ -47,6 +47,7 @@ function publishActivity(activity) {
 
     activity.publishedAt = new Date();
     addToSpine(spine, activity);
+    Object.assign(spine, getRepositoryAttrs(repository));
 
     return publishContainers(activity)
       .then(async containers => {
--- a/server/shared/publishing/helpers.js
+++ b/server/shared/publishing/helpers.js
@@ -72,12 +72,12 @@ function getRepositoryCatalog() {
   });
 }
 
-function updateRepositoryCatalog(repository, publishedAt) {
+function updateRepositoryCatalog(repository, publishedAt, updateInfo = true) {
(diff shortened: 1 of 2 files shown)


server/shared/publishing/helpers.js
```diff
@@ -47,6 +47,7 @@ function publishActivity(activity) {
 
     activity.publishedAt = new Date();
     addToSpine(spine, activity);
+    Object.assign(spine, getRepositoryAttrs(repository));
```
rkokeza: Would be good idea to create new method to handle this e.g. `applyAttributesToSpine` and apply same method to `publishRepositoryDetails` since, any changes to publishing repository details should be applied when publishing activity.

underscope:
@RoccoCocco Per offline discussion with @rkusan00, repository info should not be published upon acty publish (should be triggered by separate publish action). @rkusan00  will update this accordingly....

underscope (approved):

server/shared/publishing/helpers.js
```diff
@@ -71,12 +71,12 @@ function getRepositoryCatalog() {
   });
 }
 
-function updateRepositoryCatalog(repository, publishedAt) {
+function updateRepositoryCatalog(repository, publishedAt, updateInfo = true) {
```
rkokeza, line 74: Nitpick: Would beneficial if this was an `options` object as it already has two arguments supplied. But can leave it as is for now.

rkokeza (approved):

rkokeza (approved):

server/shared/publishing/helpers.js
rkusan00: I wanted to do that too, but since function is used in multiple places, didn't want to mess up existing code as part of this fix.

mbarisic0 (approved):

